PR #1, opened 2015-08-13
metaflac: add --scan-replay-gain option

mathstuf:
---
I think I got all the bits filled out for this (docs, etc.). Not sure if it should enforce that `--add-replay-gain` and `--scan-replay-gain` are used exclusively of each other. If not, combining the operations would be best.

Sorry if this isn't the place to send patches; I can send to the list as well if that is better. Sourceforge is giving me 403 errors on password recovery.


erikd:
Looks ok at first glance. Let me review this more thoroughly but it should be fine.

If you don't see this committed by the end of this coming weekend please feel free to ping me.

mathstuf:
I think the main thing I'm worried about is whether the output should be more structured or not. I don't know how likely someone finding a new replaygain field is though :) .

erikd:
Applied. Thanks!

PR #2, opened 2015-10-11
Let Makefile.lite build succeed on FreeBSD amd64

riggs-:
- build/config.mk: some OS call x86_64 amd64
- build/config.mk: FreeBSD needs -DHAVE_SYS_PARAM_H in CFLAGS
- build/exe.mk and lib.mk: default compilers on FreeBSD are cc/c++
- src/libFLAC++/Makefile.lite: $(OS) is not defined
- src/libFLAC++/Makefile.lite: Link -lstdc++ on FreeBSD

### DIFF
--- a/build/config.mk
+++ b/build/config.mk
@@ -53,7 +53,11 @@ ifndef PROC
             F_PIC :=
         endif
     else
-        PROC := $(shell uname -p)
+        ifeq ($(shell uname -p),amd64)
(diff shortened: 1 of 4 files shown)


erikd:
Patch manually applied and pushed to Xiph.org repo and github.

Thanks!

riggs-:
Great, thanks Erik!

PR #3, opened 2015-10-25
Fix manpage: synonyms for encoding options -0 .. -8 now match the code

riggs-:
The short options "-0" to "-8" in the manpage currently do not match the encoder settings. This patch gets the manpage in sync with the code.


erikd:
Applied. Thanks.

PR #4, opened 2015-10-29
Fix some minor build issues on Darwin/x86_64

riggs-:
Address several points on Darwin / Mac OS:
- Fix open question in cpu.c: Provide proper headers for sysctlbyname() on Darwin
- Include config.mk in Makefile.lite files for binaries and libs to allow accessing variables OS, PROC, etc.
- Allow Makefile.lite to build 64 bit binaries on Darwin/x86_64


erikd:
Applied. Thanks!

PR #5, opened 2015-12-24
Explicitly use base class virtual methods; silence unnecessary warning

riggs-:
When compiling libFLAC++ with clang++, it emits warnings to inform about potential hiding of overloaded virtual functions. There is no problem in the current code and nothing is hidden by accident. Including the base class's method declarations explicitly by "using" directives, we make the intentions clear to the compiler, and it does no longer emit the warning.


erikd:
Applied. Thanks.

PR #6, opened 2016-01-03
Cleanup FLAC__bitmath_silog2()

riggs-:
- Retire 32bit variant of silog2(), since only the _wide variant is used
- Rename FLAC__bitmath_silog2_wide() to FLAC__bitmath_silog2()
- Replace existing implementation by shorter, clearer implementation
  using optimised routines from bitmath.h
- Update Copyright string to 2016 in changed files


erikd:
Applied. Thanks.

PR #7, opened 2016-01-09
Fix description of --rice-partition-order in flac -H

riggs-:
Closing http://sourceforge.net/p/flac/bugs/352/:
This is more of a "human language versus programmer parlor" issue.
src/flac/main.c will return usage error for an argument greater
than FLAC__MAX_RICE_PARTITION_ORDER (15u).
While in programming "0..16" usually means "from zero to 15", in
natural human-to-human talk, it would rather mean "from zero to 16".
This changes the wording a bit to avoid this misunderstanding.


erikd:
Applied. Thanks.

riggs-:
What is the policy on bug closing? Can you close the issue on SF? Otherwise one might think it's not fixed and start looking into it.

erikd:
Done. Thanks for reminding me.

PR #8, opened 2016-01-22
cppcheck: fix some reports

serval2412:
[src/libFLAC/metadata_iterators.c:3247]: (error) va_list 'va' was opened but not closed by va_end().
[src/share/grabbag/snprintf.c:66]: (error) va_list 'va' was opened but not closed by va_end().
[src/utils/flactimer/main.cpp:84] -> [src/utils/flactimer/main.cpp:83]: (warning) Either the condition '!fout' is redundant or there is possible null pointer dereference


erikd:
Applied with some minor improvements to the fix and the commit message.

Thanks.

PR #9, opened 2016-02-07
Unify floating point types in libFLAC

riggs-:
By convention given in src/libFLAC/include/private/float.h, within libFLAC FLAC__{float|double} instead of just {float|double} shall be used. This applies the changes to the few remaining instances that do not respect this convention so far.

### DIFF
--- a/src/libFLAC/lpc.c
+++ b/src/libFLAC/lpc.c
@@ -53,11 +53,11 @@
 
 #if defined(_MSC_VER) && (_MSC_VER < 1800)
 #include <float.h>
-static inline long int lround(double x) {
+static inline long int lround(FLAC__double x) {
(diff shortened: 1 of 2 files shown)


erikd:
Not really very keen on this patch,  mainly because I think internal types like `FLAC__double` / `FLAC__float` are simply not worth it. They provide exactly zero benefits. 

Fortunately, these two types don't leak into the public headers, so I'd be much keener to accept a patch that does the reverse of this one, remove the `FLAC__*` versions of `double` and `float` in favour of the standard C ones.

riggs-:
Thanks for the quick response. I submitted this patch because the policy in src/libFLAC/include/private/float.h mandates it. But your argument is totally understandable. I will prepare the reverse patch.

PR #11, opened 2016-05-11
libFLAC: infer the cpu type from common predefined macros

timblechmann:
adapted from https://sourceforge.net/p/predef/wiki/Architectures/

--

this simplifies building universal binaries on osx


erikd:
Applied. Thanks.

PR #12, opened 2016-07-06
stream_decoder: check state==ABORTED after process_single() for seek

MaxKellermann:
FLAC__stream_decoder_process_single() ignores frame_sync_() errors,
which means the caller cannot rely solely on the boolean return value,
it is also required to check the new "state".

After FLAC__stream_decoder_process_until_end_of_metadata(),
state==SEARCH_FOR_FRAME_SYNC and
last_frame.header.number_type==FRAME_NUMBER.  When an application
seeks at this time, but an I/O error occurs, then
FLAC__stream_decoder_process_single() returns true, but no frame has
been read yet, i.e. last_frame.header.number_type is still
FRAME_NUMBER.  This triggers the assertion in
seek_to_absolute_sample_():

 FLAC__ASSERT(decoder->private_->last_frame.header.number_type == FLAC__FRAME_NUMBER_TYPE_SAMPLE_NUMBER);

So what needs to be done is check for state==ABORTED after the
FLAC__stream_decoder_process_single() call.

This bug can be triggered remotely with the Music Player Daemon
(https://www.musicpd.org/), and crashes the process.

### DIFF
--- a/src/libFLAC/stream_decoder.c
+++ b/src/libFLAC/stream_decoder.c
@@ -3134,7 +3134,8 @@ FLAC__bool seek_to_absolute_sample_(FLAC__StreamDecoder *decoder, FLAC__uint64 s
 		 * FLAC__stream_decoder_process_single() to return false.
 		 */
 		decoder->private_->unparseable_frame_count = 0;
-		if(!FLAC__stream_decoder_process_single(decoder)) {
+		if(!FLAC__stream_decoder_process_single(decoder) ||
+		   decoder->protected_->state == FLAC__STREAM_DECODER_ABORTED) {
 			decoder->protected_->state = FLAC__STREAM_DECODER_SEEK_ERROR;
 			return false;
 		}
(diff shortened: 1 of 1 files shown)


erikd:
Interesting, do you have a file which triggers this issue?

MaxKellermann:
On 2016/07/06 21:31, Erik de Castro Lopo <notifications@github.com> wrote:
> Interesting, do you have a file which triggers this issue?

Here's one: https://bugs.musicpd.org/view.php?id=4552

My test source code triggering the bug:
https://paste.debian.net/777929/

erikd:
I compiled your test source code and ran it on a number of FLAC files, including the one at https://bugs.musicpd.org/view.php?id=4552 (4m33s.flac) and none of them seem to trigger any issue.

I was compiling against the Debian 1.3.1-4 packages. I've checked the Debian packaging meta data and there is only one source code patch (taken from Git, commit  43ba7ad05f1656e885ce2f34a9a72494f45705ae) that seems unrelated to the issue you are reporting.

MaxKellermann:
On 2016/07/09 12:09, Erik de Castro Lopo <notifications@github.com> wrote:
> I compiled your test source code and ran it on a number of FLAC files, including the one at https://bugs.musicpd.org/view.php?id=4552 (4m33s.flac) and none of them seem to trigger any issue.
> 
> I was compiling against the Debian 1.3.1-4 packages. I've checked the Debian packaging meta data and there is only one source code patch (taken from Git, commit  43ba7ad05f1656e885ce2f34a9a72494f45705ae) that seems unrelated to the issue you are reporting.

Both 1.3.1 and git master are affected, but the Debian packages were
compiled with assertions disabled, so you don't see the bug there.

erikd:
Yep, with assertions that gets triggered. Thanks.

PR #13, opened 2016-07-08
stream_decoder: fix integer underflow due to malformed wasted_bits

MaxKellermann:
It is pretty easy for a malformed FLAC file to underflow the "bps"
variable.  In the debug build, this results in an assertion failure in
FLAC__bitreader_read_raw_uint32():

    FLAC__ASSERT(bits <= 32);

In non-debug builds, this simply makes
FLAC__bitreader_read_raw_uint32() fail because
bitreader_read_from_client_() doesn't find enough buffer space for
2**32-1 bits.  But since the failing FLAC_ASSERT() is reasonable, this
should be caught in the FLAC__bitreader_read_raw_uint32() caller.

Demo file: http://max.kellermann.name/download/mpd/flac/afl/integer_underflow.flac (bug found with AFL)

### DIFF
--- a/src/libFLAC/stream_decoder.c
+++ b/src/libFLAC/stream_decoder.c
@@ -2481,6 +2481,8 @@ FLAC__bool read_subframe_(FLAC__StreamDecoder *decoder, unsigned channel, unsign
 		if(!FLAC__bitreader_read_unary_unsigned(decoder->private_->input, &u))
 			return false; /* read_callback_ sets the state for us */
 		decoder->private_->frame.subframes[channel].wasted_bits = u+1;
+		if (decoder->private_->frame.subframes[channel].wasted_bits >= bps)
+			return false;
 		bps -= decoder->private_->frame.subframes[channel].wasted_bits;
 	}
 	else


erikd:
What version of FLAC are you using? I tested this with the flac executable in the Debian 1.3.1 package and with git HEAD. Both give identical results:

```
flac 1.3.1
Copyright (C) 2000-2009  Josh Coalson, 2011-2014  Xiph.Org Foundation

integer_underflow.flac: ERROR while decoding data
                        state = FLAC__STREAM_DECODER_READ_FRAME

a.wav: ERROR while decoding metadata
       state = FLAC__STREAM_DECODER_END_OF_STREAM
```

MaxKellermann:
On 2016/07/09 12:08, Erik de Castro Lopo <notifications@github.com> wrote:
> What version of FLAC are you using? I tested this with the flac executable in the Debian 1.3.1 package and with git HEAD. Both give identical results:

Both 1.3.1 and git master are affected, but the Debian packages were
compiled with assertions disabled, so you don't see the bug there.

erikd:
OK, confirmed this one and confirmed the fix. Thanks.

erikd:
Thanks. Applied.

PR #15, opened 2016-10-04
Add draft spec for encapulation in mp4

rillian:
We've been working on an mp4 encapsulation for flac. Record the change history here for broader reference and review.


erikd:
Merged. Thanks.

PR #17, opened 2016-10-27
isoflac: Correct additional 'MEATADATA' typos.

rillian:
Thanks to Matthew Gregan for the report.


erikd:
Manually merges and pushed. Thanks @rillian !

PR #21, opened 2016-12-20
Added CMake build files for libFLAC

dschreij:
Courtesy of @soulik who wrote the initial versions of the CMake build files.
I adapted these to work in an Anaconda build environment (ref https://github.com/conda-forge/staged-recipes/pull/2026/files/817d088b60001ecca8b1259d90f33eecdbb69b51 and https://github.com/conda-forge/staged-recipes/pull/2101), but they should also work outside of it. Maybe these files are useful for you (especially for compilation on Windows) so hence I send you this PR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,191 @@
+# Written by Mário Kašuba (mario.kasuba at it-academy dot sk), 2016
+# Edited by Daniel Schreij (dschreij at gmail dot com)
+
+cmake_minimum_required (VERSION 3.1.0)
+# enable_language(ASM-NASM)
+
+project (FLAC)
+
+# Find required Ogg depencies.
+# Assumes CMAKE_PREFIX_PATH is passed to cmake as a basis to search from. 
+# The paths below are relative to this path.
+
+# Find ogg headers
+FIND_PATH(OGG_INCLUDE_DIR
+    ogg.h
+    PATH_SUFFIXES include/ogg
+)
+
+if(OGG_INCLUDE_DIR)
+    message(STATUS "Found ogg headers at ${OGG_INCLUDE_DIR}")
+    include_directories(OGG_INCLUDE_DIR)
+else()
+    message(FATAL_ERROR "Could not find required OGG headers" )
+endif()
+
+# Find ogg libraries to link to later
+FIND_LIBRARY(OGG_LIBRARIES
+    NAMES ogg libogg
+    PATH_SUFFIXES lib
+)
+
+if(OGG_LIBRARIES)
+    message(STATUS "Found ogg libs at ${OGG_LIBRARIES}")
+else()
+    message(FATAL_ERROR "Could not find required OGG library" )
+endif()
+
+add_definitions(-DVERSION="1.3.1" -DFLAC_API_EXPORTS)
+
+if (NOT FLAC_CPU_X86_64)
+    set(FLAC_USE_ASM OFF CACHE BOOL "Use assembler")
+endif()
+
+set(FLAC_USE_3DNOW OFF CACHE BOOL "Use 3DNow!")
+set(FLAC_USE_AVX OFF CACHE BOOL "Use AVX instructions")
+set(FLAC_USE_SSE OFF CACHE BOOL "Use SSE instructions")
+set(FLAC_USE_SSE2 OFF CACHE BOOL "Use SSE 2 instructions")
+set(FLAC_USE_SSE41 OFF CACHE BOOL "Use SSE 4.1 instructions")
+
+set(FLAC_SRC_FLAC
+    src/flac/analyze.c
+    src/flac/decode.c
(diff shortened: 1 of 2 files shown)


erikd:
There doesn't seem to be any way to run the tests with CMake. I think it would be a bad idea to add a build system that is significantly less full featured than the current ones (FLAC already has three).

dschreij:
True, even though it is possible to add tests in CMake too. They just haven't been implemented yet and sadly I don't have much experience with this either. I thought adding CMake options would be appreciated since the other xiph libraries (ogg and vorbis)  also started to provide these CMake files in their latest snapshots. Additionally, on Windows it is a pain to build from CLI using the provided .sln files, especially because they often depend on the visual studio version they were created with and msbuild doesn't play nice with other versions. With CMake building from CLI is a breeze, and if one does it well, you only have to write one CMake file for all OS's that are around. In the end it's up to you if you find this useful enough to adopt ☺️

dschreij:
A quick search came up with  https://cmake.org/Wiki/CMake/Testing_With_CTest,  so CMake even has a dedicated testing module. Sadly I don't have the time to dive into this anytime soon, so you might want to shelve this until I or someone else has had the time to dive into this testing module.

mancoast:
Greetings @dschreij @erikd 

Please Merge as CMAKE Build required for k1om cross compilation.

Thanks,
Coast

erikd:
I'll merge it when the CMake build system builds the tests.

mancoast:
Greetings,

Implementing testing within CMAKE is no issue.
Please be very specific regarding the testing requirements.

Thanks,
coast

erikd:
The CMake build system has to build and run the same tests as the three other existing build systems.

dschreij:
One needs to use [CTest](https://cmake.org/Wiki/CMake/Testing_With_CTest) for testing in CMake builds. I don't know how easy the existing tests will translate into CTest though, or how much work it is to convert them.

erikd:
> I don't know how easy the existing tests will translate into CTest though, or how much work it is to convert them.

Do *not* convert them. The existing tests should only require minimal changes and should still build with the existing build systems. What is required is that CMake build the test programs and provide a means for running the tests.

Qix-:
```
There doesn't seem to be any way to run the tests with CMake
```

CTest is just an executor - it doesn't require that you change tests as long as they're just command lines.

This is proper, standard CMake to enable/add tests.

```cmake
if (BUILD_TESTING)
    enable_testing ()
    add_test (NAME flac-some-test COMMAND ./flac-tests --some --arg)
    add_test (NAME flac-another-test COMMAND ./flac-tests --some arg)
endif ()
```

```console
$ mkdir build && cd build
$ cmake .. -DCMAKE_BUILD_TYPE=Debug -DBUILD_TESTING=On
$ make
$ ctest
```

`ctest` will run all tests in the build directory. `ctest --help` for all of the options you'd ever need.

dschreij:
I am completely fine with someone else taking this over from me. I'm no CMake expert either, and I composed this CMakeList with the limited knowledge I have of the system (and this was a while ago indeed, haven't worked with it since). So be my guest.

Regarding autotools: this is not really well supported on windows unless you use cygwin or similar, am I right? I composed the CMake buildfile because I wanted to add this library to the Anaconda Python distribution and it seemed the only way to go at the time.

Qix-:
> Regarding autotools: this is not really well supported on windows unless you use cygwin or similar, am I right?

In my experience, this is correct, though I'd imagine pro-autotools users would have some sort of answer for it.

Something to note is that users of IDEs get project generation for free by using CMake, something that autotools can't do. Further, I am not stuck using Makefiles but instead also get support for Ninja, which in my experience compiles large projects much faster.

CMakeLists.txt
```diff
@@ -0,0 +1,191 @@
...
+#       COMPILE_FLAGS "/FI\"${FLAC_SOURCE_DIR}/src/win32_dll.h\""
+#   )
+endif()
+
+target_link_libraries (FLAC
+    ${OGG_LIBRARIES}
+)
+
+set_target_properties(FLAC
+    PROPERTIES PREFIX "" OUTPUT_NAME "libFLAC"
```
dekonvoluted: This seems odd, considering that had you not changed it, the default prefix would have been `lib` and the default output name would have been the name of the target, `FLAC`, so you'd still end up with `libFLAC.so`.

Is there any reason why you wanted to specify this (perhaps because of how things work on windows)?

dschreij: I don't really remember my reasons for these changes, if they were mine to begin with. I largely based this script on one composed by @soulik.

JonnyPtn:
I'd put my support behind this too - building stuff cross-platform is much easier with CMake. I have a fair bit of CMake experience so happy to help/review whatever you put together @Qix-

erikd:
FLAC now has a CMake build system. Will close this.

PR #30, opened 2017-02-16
cpu.h: Another CPP fixup

erikd:
Just a temp PR to force a travis build.

### DIFF
--- a/src/libFLAC/include/private/cpu.h
+++ b/src/libFLAC/include/private/cpu.h
@@ -55,6 +55,9 @@
 
 #endif
 
+#ifndef __has_attribute
+#define __has_attribute(x) 0
+#endif
 
 #if FLAC__HAS_X86INTRIN
 /* SSE intrinsics support by ICC/MSVC/GCC */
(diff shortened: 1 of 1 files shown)


lvqcl:
So the relevant part of the code is basically:
```
#if defined __INTEL_COMPILER
...
#elif defined _MSC_VER
...
#elif defined __clang__ && __has_attribute(__target__) /* clang */
...
#elif !defined __clang__ && (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 9)) /* GCC 4.9+ */
...
#else /* older GCC and clang */
...
#endif /* compiler version */
```

It seems that a compiler other than ICC/MSVC/clang/GCC can complain about undefined    ̲ ̲ GNUC ̲ ̲ variable in the [line 111](https://github.com/erikd/flac/blob/da103e11eb82c98f395bcfcec05be646c8b28eba/src/libFLAC/include/private/cpu.h#L111), so it's better to write it as:
```
#elif defined __GNUC__ && !defined __clang__ && (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 9)) /* GCC 4.9+ */
```
And the comment  _/* older GCC and clang */_  is now incorrect, so it's better to remove it.

Another possibility is to go back to [this version](https://github.com/xiph/flac/blob/3a7884bf532b134804a83632527d2de9a6b25a30/src/libFLAC/include/private/cpu.h) and to add (#ifndef/#define __has_attribute)  to it.

erikd:
The problem with the version you suggest is that it has excessive/un-needed nesting of `#ifdef`. I think added an extra `defined __GNUC__` is a better options.

erikd:
Damn pushed it to Xiph instead of here.

erikd:
Closing.

PR #33, opened 2017-06-22
Update lpc_intrin_sse.c

Shark64:
Minor change, remove redundant assignment before broadcasting into xmm registers


erikd:
@lvqcl What do you think?

lvqcl:
I tested MSVS 2105 and GCC 7.1, and they generate exactly the same code before and after this change, so it should be OK.

erikd:
Thanks @lvqcl for reviewing.

Thanks @Shark64 for the patch. I've manually merged it into the Xiph.org git tree which will be mirrored here in the next 24 or so hours.

PR #34, opened 2017-06-22
Silence warning regarding -finline-functions, it is a gcc-specfic flag

riggs-:
clang issues warnings about -finline-functions being an unrecognised flag. This patch adapts CFLAGS such that it is only used on gcc where it is a recognised flag.


erikd:
Thanks! Applied manually. Should show up here on Github with the next 24 hours.

PR #35, opened 2017-06-22
Add missing prototype for construct_block, silence compilation warning

riggs-:
Tested on clang only so far.


erikd:
Thanks! Applied manually. Should show up here on Github with the next 24 hours.

PR #36, opened 2017-06-29
Correct flattop window coefficients, normalising window amplitude to 1

riggs-:
The flattop window is generated based on the coefficients as described in https://octave.sourceforge.io/signal/function/flattopwin.html, but normalisation is not performed. This results in an amplitude of more than 4.6 at the center of the window. Using the normalised coefficients as given in https://au.mathworks.com/help/signal/ref/flattopwin.html?s_tid=gn_loc_drop, the window is properly normalised with and amplitude of 1.0 at the center.

(All the other windows are already correctly normalised to a maximum amplitude of 1.0 at the highest point(s).)

### DIFF
--- a/src/libFLAC/window.c
+++ b/src/libFLAC/window.c
@@ -109,7 +109,7 @@ void FLAC__window_flattop(FLAC__real *window, const FLAC__int32 L)
 	FLAC__int32 n;
 
 	for (n = 0; n < L; n++)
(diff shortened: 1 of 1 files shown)


erikd:
Merging this to the Xiph.org Git tree. Should be mirrored here within about 24 hours. Thanks!

PR #41, opened 2017-08-20
Proposal: Change apodization parameters for built-in compression levels 7 and 8

riggs-:
I spend some time throwing various content at the encoder with lots of apodization settings and captured how often a particular apodization function resulted in the best encoding.
Long story short: The previous level-8 apodization settings are the best choice in about 75% of all frames.

Due to the fact that even single thread performance of CPUs has improved a bit since the compression levels have been changed last, I propose the new levels given in this patch. 
Rationale:
The proposed apodization settings for level 8 are the best choice in about 80% of all frames in my tests. 80% is a reasonable Pareto based approach. The encoder slowdown for adding those apodizations is very small and likely not an issue for most users, given the current CPUs.

Details:
- Use the apodization parameters previously used for level 8 now in level 7
- Update the level 8 apodization parameters to yield slightly better
  compression
- Update manpage accordingly

### DIFF
--- a/man/flac.1
+++ b/man/flac.1
@@ -264,10 +264,10 @@ Synonymous with -l 8 -b 4096 -m -r 5
 Synonymous with -l 8 -b 4096 -m -r 6 -A tukey(0.5) -A partial_tukey(2)
 .TP
 \fB-7, --compression-level-7\fR
-Synonymous with -l 12 -b 4096 -m -r 6 -A tukey(0.5) -A partial_tukey(2)
+Synonymous with -l 12 -b 4096 -m -r 6 -A tukey(0.5) -A partial_tukey(2) -A punchout_tukey(3)
 .TP
 \fB-8, --compression-level-8\fR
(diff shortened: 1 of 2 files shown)


riggs-:
Actually, let's put this on hold. I have found a small number of edge cases where the new settings perform worse than before. Although the size is only about 0.01% worse, I'd like to investigate. Will update the pull request when I have results.

erikd:
👍 On hold. I've been super busy and not had a proper chance to look at this.

riggs-:
I have done some more testing with various apodization settings and encoding time measurement. While it is always nice to improve compression, the results show that changing the compression presets do note make sense from an economical point of view. In my tests, in order to consistently save ~1 MB compared to ~500 MB (0.2%) flac content compressed with the current '-8' preset, the encoding time would almost **double**. This is much worse than I hoped when working with the patch set.

Summarising, it does not make sense to change the built-in presets -0 to -8 at this point in time.
Users who are willing to accept the tiny gains in compression at the expense of much greater encoding time costs can still use more expensive compression settings via the command line.

Closing this pull request for now.

erikd:
👍 Thanks for investigating this anyway.

PR #47, opened 2017-12-04
libFLAC : new function returns client_data from decoder

ChristopD:
Hello,

As developer, i'm actually writing java interface to use libFLAC with JNA, to reproduce "flac -t" function, needed to validate/unvalidate flac files before sending them in our preservation system.

The most part of implementation with JNA is working fine. Callback are called, and client_data parameter is a smart way to keep custom datas.
My problem is : when error_callback is called, we update client_data to add error status, but client_data reference (java side) is never updated while native code of libFlac receive an updated client_data from callback call.
Use a class variable is working fine, but it's not a safe way to solve our problem in thread context.

The only thing is missing to us, is a function to return client_data from decoder in libFLAC. With this simple function, we can get our client_data, with error status and cause stored in out client_data structure.

If this modification could be added into libFLAC, it should help other JNA implementers, and avoid a patch-task for new libFLAC releases.

Thanks for you attention and for the great work !
Best regards,

Christophe Dumont

### DIFF
--- a/src/libFLAC/include/protected/stream_decoder.h
+++ b/src/libFLAC/include/protected/stream_decoder.h
@@ -57,4 +57,9 @@ typedef struct FLAC__StreamDecoderProtected {
  */
 uint32_t FLAC__stream_decoder_get_input_bytes_unconsumed(const FLAC__StreamDecoder *decoder);
 
+/*
+ * return client_data from decoder
+ */
+FLAC_API void *get_client_data_from_decoder(FLAC__StreamDecoder *decoder);
+
 #endif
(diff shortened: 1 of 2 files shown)


erikd:
Applied to git.xiph.org repo.  It should show up here in Github within 24 hours.

Thanks!

sezero:
This function is marked as public (FLAC_API) but new function's prototype
is added to src/libFLAC/include/protected/stream_decoder.h and not to the
public include stream_decoder.h, therefore applications aren't aware of it.
Also, despite the new function being public, it isn't prefixed by `FLAC_`

See:  https://github.com/xiph/flac/issues/121

PR #49, opened 2018-01-05
UWP build fix

claudiuslollarius:
This is to allow libflac to build in UWP.
`CreateFileA` and `CreateFileW` are not available on UWP, `CreateFile2` needs to be used instead.


erikd:
Merged to https://git.xiph.org/flac.git . It should show up here within about 24 hours.

Thanks!

PR #52, opened 2018-02-05
libFLAC: refactor restart/flush from init/finish

gabrielcorneanu:
As discussed on mailing list, I tried to split those functions. My goal was to allow using the same encoder without re-initializing everything (avoiding lots of malloc, etc).
I have an application compressing hundreds of thousands of small signals (a few k-samples), the time saving are up to 25% on corner cases.
The original API should work as before.
One other small change: default empty vorbis comment is only generated for ogg files; flac streams are valid without this block.
Flac.exe seems to generate it anyway, which I don't really care.


src/libFLAC/stream_encoder.c
```diff
@@ -1146,6 +1092,88 @@ static FLAC__StreamEncoderInitStatus init_stream_internal_(
 	encoder->protected_->seektable_offset = 0;
 	encoder->protected_->audio_offset = 0;
 
+	FLAC__bitwriter_release_buffer(encoder->private_->frame);
+	FLAC__bitwriter_clear(encoder->private_->frame);
+
+#if FLAC__HAS_OGG
+	/* reorder metadata if necessary to ensure that any VORBIS_COMMENT is the first, according to the mapping spec */
+	if(is_ogg && 0 != encoder->protected_->metadata && encoder->protected_->num_metadata_blocks > 1) {
```
moisesmcardona: This fails because `is_ogg` is undefined. I believe it should be `encoder->private_->is_ogg`

ktmf01 (changes requested):
This code doesn't actually prevent a Vorbiscomment from being generated. At this point, there is usually already a Vorbiscomment in `encoder->protected_->metadata[]`, and `metadata_has_vorbis_comment` is already true.  

Besides, not adding a default Vorbis comment by default seems like a very bad idea to me, because the vendor string is added there. Not having that doesn't seem worth saving 40 bytes per file to me.

I would strongly advice against merging this commit in its current form. I think the vendor string has to much value to not enable it by default. For such a specific corner case in usage far out-of-scope for FLAC, I think there is nothing wrong with expert users patching it in themselves.

gabrielcorneanu:
I didn't suggest not having comments by default 😊.
I just need a way to remove/avoid them.

ktmf01:
Yes, but d5b1413 does neither, while its name suggest it disables outputting a vorbiscomment block. 

Anyway, I wrote you a patch that does what I think you want: it adds a define that when enabled stops FLAC from adding any metadata beyond a STREAMINFO block. `make check` fails obviously, because it expects that metadata. You can find it here: https://github.com/ktmf01/flac/commit/334ffe32924c8e41a38cdc3d51f8d645842d9dc2 I hope you like it

PR #60, opened 2018-05-10
Improve SIMD detection with Visual Studio

evpobr:
Visual Studio can use GCC (for Android) and Clang (supported toolset) so they must be checked before native compiler.

### DIFF
--- a/src/libFLAC/include/private/cpu.h
+++ b/src/libFLAC/include/private/cpu.h
@@ -76,21 +76,6 @@
     #define FLAC__AVX2_SUPPORTED 1
     #define FLAC__FMA_SUPPORTED 1
   #endif
-#elif defined _MSC_VER
-  #define FLAC__SSE_TARGET(x)
-  #define FLAC__SSE_SUPPORTED 1
-  #define FLAC__SSE2_SUPPORTED 1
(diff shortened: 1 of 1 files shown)


erikd:
Can't get this to build in CI. It times out on OS X. May need some horrible hack to only run part of the testsuite on OS X.

lvqcl:
IIRC there were `make check` and `make fullcheck`; but somewhere between flac 1.3.1 and 1.3.2 `check` target was removed and `fullcheck` was renamed to `check`.

erikd:
Yes. Back when the `make check` was removed in favor of `make check` Travis builds were not timing out and it was thought that running all the tests all the file, was a good idea.

I'd like to keep `make check` run all the tests and maybe add a `make fastcheck` that does what the old `fullcheck` did. Then in Travis we could run `make check` on Linux and `make fastcheck` on OS X.

enzo1982:
I ran into the same timeout issue in PR #63. Fixed it in commit beb7363.

This PR can be rebased after merging #63 and should build successfully then.

erikd:
#63 has been merged.

willofpath (approved):

PR #57, opened 2018-04-30
Use faster CRC16 algorithm to speed up encoding/decoding

enzo1982:
Hi all,

This PR replaces the single look-up-table CRC16 implementation with the slicing-by-8 algorithm. The new algorithm is approximately 5 times faster than single table lookup CRC and improves FLAC encoding/decoding performance by about 5%.

The PR consists of 3 commits:

- Commit 4ec6cf1 removes some unused CRC8 functions.
- Commit f789008 replaces the CRC16 algorithm and speeds up encoding performance.
- Commit f6a0d22 implements blockwise CRC updates for decoding (instead of wordwise) and speeds up decoding.

The [last of the relevant patents](https://patents.google.com/patent/US20040250193) on this algorithm expired in December 2017. The slicing algorithm is already used by zlib, the Linux kernel and other projects.

See [here](https://freac.org/developer-blog-mainmenu-9/14-freac/277-fastcrc) for my original article on this and [here](https://hydrogenaud.io/index.php/topic,115900.0.html) for the discussion on HydrogenAudio.

It would be great to see this added to the official FLAC source.

Robert

PR #63, opened 2018-05-19
Add unit tests for CRC and BitReader

enzo1982:
This PR adds unit tests for CRC calculation and for the BitReader. See PR #57 for reference.

The CRC tests verify CRC8 and CRC16 calculation against simple reference CRC implementations.
The BitReader test includes code to test unaligned small reads (less than a word) that would trigger the bug found by @lvqcl, fixed in commit d57fb5c in PR #57.

In addition, obsolete CRC8 functions are removed (commit 8cebcf7) and some checks and messages in the BitWriter test are fixed (commit d93eec3).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,32 @@
 language: c
-compiler:
-  - gcc
-  - clang
-os:
-  - linux
-  - osx
+matrix:
+    include:
+        - os: linux
+          compiler: gcc
+
+        - os: linux
+          compiler: clang
+
+        - os: osx
+          compiler: gcc
+
+        - os: osx
+          compiler: clang
+
+        - os: linux
+          compiler: gcc
+          env: CONFIGURE_OPTS=--enable-64-bit-words
+
+        - os: linux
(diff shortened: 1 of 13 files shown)


enzo1982:
The Travis check was terminated because it ran for too long. Everything looked good up to that point though. Is this a configurable timeout or a hard limit?

Do you know if there a way to restart it without pushing new code? For now, I think I'll just try my luck by pushing the last commit again.

.travis.yml
```diff
@@ -1,10 +1,32 @@
...
+
+        - os: osx
+          compiler: gcc
+
+        - os: osx
+          compiler: clang
+
+        - os: linux
+          compiler: gcc
+          env: CONFIGURE_OPTS=--enable-64-bit-words
```
erikd: That's a nice solution to this issue. Thanks!

erikd:
Very useful contribution. Thanks!

willofpath (approved):

willofpath (approved):

PR #64, opened 2018-05-20
Add libogg dependency to Travis configuration

enzo1982:
This adds the libogg dependency to the Travis configuration in order to run CI with Ogg support enabled.

The commit also simplifies the .travis.yml file structure by not using an explicit matrix.


erikd:
Thanks!

PR #65, opened 2018-05-20
Use faster CRC16 algorithm to speed up encoding/decoding

enzo1982:
This is the same as PR #57 which somehow got lost after being merged to xiph/flac.


erikd:
Merged manually and then pushed to https://git.xiph.org/?p=flac.git;a%3Dsummary

enzo1982:
@erikd Thanks, it looks good now. It's great working with you!

PR #66, opened 2018-05-20
Add libogg dependency to Travis configuration

enzo1982:
This is the same as PR #64 which somehow got lost after being merged to xiph/flac.


erikd:
Merged manually and then pushed to https://git.xiph.org/?p=flac.git;a%3Dsummary

PR #70, opened 2018-05-29
Fix --output-prefix with input-files in sub-directories.

orbea:
When `--output-prefix=/foo/bar/` is used with `some/path/to/foo.flac` flac will fail to create the sub-directories instead of just creating the new file in the `/foo/bar/`. So this will strip the unwanted path from `foo.flac` and just successfully create new files to the output directory.

Please review this to make sure I am doing it correctly.

Also please view this issue report which also describes the issue which this patch fixes. I don't have a sourceforge account so someone else will have to respond to that issue report if necessary. :)

https://sourceforge.net/p/flac/bugs/463/.

### DIFF
--- a/src/flac/main.c
+++ b/src/flac/main.c
@@ -2173,12 +2173,30 @@ int decode_file(const char *infilename)
 const char *get_encoded_outfilename(const char *infilename)
 {
 	const char *suffix = (option_values.use_ogg? ".oga" : ".flac");
-	return get_outfilename(infilename, suffix);
+	char *p;
+
+	if(option_values.output_prefix) {
+		p = grabbag__file_get_basename(infilename);
+	}
+	else {
+		p = infilename;
+	}
+
+	return get_outfilename(p, suffix);
 }
 
 const char *get_decoded_outfilename(const char *infilename)
 {
 	const char *suffix;
+	char *p;
+
+	if(option_values.output_prefix) {
+		p = grabbag__file_get_basename(infilename);
+	}
+	else {
+		p = infilename;
+	}
+
 	if(option_values.analyze) {
 		suffix = ".ana";
 	}
@@ -2197,7 +2215,7 @@ const char *get_decoded_outfilename(const char *infilename)
 	else {
 		suffix = ".wav";
 	}
-	return get_outfilename(infilename, suffix);
+	return get_outfilename(p, suffix);
 }
 
 const char *get_outfilename(const char *infilename, const char *suffix)
--- a/test/test_flac.sh
+++ b/test/test_flac.sh
@@ -726,6 +726,43 @@ if [ $is_win = no ] ; then
 	echo OK
 fi
 
+############################################################################
+# test --output-prefix
+############################################################################
+
+in_dir=./tmp_in
+out_dir=./tmp_out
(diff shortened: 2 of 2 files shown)


orbea:
The one travis fail is not related to this change and is failing on a dependency.
```
E: Package 'libogg-dev' has no installation candidate
```

orbea:
I found an issue with the first patch here.

In short before this change.
1. `flac --decode foo/bar.flac`
2. Creates `foo/bar.wav`.

After this change.
1. `flac --decode foo/bar.flac`
2. Creates `bar.wav`.

It seems the whole filename including the path should preserved unless using `--output-prefix`. I added a second commit making this change, if this is agreeable I can squash the commit next. :)

orbea:
@erikd Is there anything more I can do here? Seeing this fixed in the next release would mean a lot to me.

erikd:
Would really like to see a test for this so that it doesn't get broken in future.

orbea:
@erikd Thanks for the feedback! I've done a few things here.

1. I squashed the two existing commits.
2. I've added a new second commit containing a test in `test/test_flac.sh` which will test relative directories with both encoding and decoding.
3. I've tested that all tests pass with `./autogen.sh && ./configure && make && make check`.

Can you review the tests to ensure they make sense and are satisfactory?

The idea is to copy a raw and flac file to `test/tmp_in/` and then test that the files are created and exist in the `test/tmp_out/` directory when encoding or decoding. This will be tested with and without `--ouput-prefix=` to ensure the behavior does not break for either. These two temporary directories are then removed at the end if the tests pass.

Here is the output when the tests pass.
```
testing --output-prefix=./tmp_out/ (encode)... OK
testing --output-prefix=./tmp_out/ (decode)... OK
```

I also tried applying the tests without the first commit to master and found it will correctly error.
```
testing --output-prefix=./tmp_out/ (encode)... ERROR generating FLAC file in ./tmp_out (encode)
make[1]: *** [Makefile:711: check] Error 1
make[1]: Leaving directory '/media/gittings/forks/flac/test'
make: *** [Makefile:491: check-recursive] Error 1
```

orbea:
Is there anything more I can do?

src/flac/main.c
```diff
@@ -2173,12 +2173,30 @@ int decode_file(const char *infilename)
 const char *get_encoded_outfilename(const char *infilename)
 {
 	const char *suffix = (option_values.use_ogg? ".oga" : ".flac");
-	return get_outfilename(infilename, suffix);
+	char *p;
```
erikd: That needs to be `const char`.

src/flac/main.c
```diff
@@ -2173,12 +2173,30 @@ int decode_file(const char *infilename)
...
+		p = infilename;
+	}
+
+	return get_outfilename(p, suffix);
 }
 
 const char *get_decoded_outfilename(const char *infilename)
 {
 	const char *suffix;
+	char *p;
```
erikd: Same here.

erikd:
Actually never mind. I fixed those two minor issues locally and then pushed to the Xiph.org git repo. It should show up here on Github within a couple of hours.

orbea:
Thanks!

PR #78, opened 2018-08-18
move CreateFile function outside of libFLAC

lvqcl:
This patch renames flac_internal_CreateFile_utf8() function to grabbag__CreateFile_utf8() and moves it from /src/libFLAC/windows_unicode_filenames.c into src/share/grabbag/file.c.
This function is not used by libFLAC anyway.

After this, it should be possible to compile libFLAC as UWP.

In the old code there were 3 different #if  conditionals around the code that uses CreateFile_utf8():
1)  #ifdef _WIN32 ... #endif
2)  #if defined _WIN32 && !defined __CYGWIN__ ... #endif
3)  #if defined _MSC_VER || defined __MINGW32__ ... #endif
It seems for me that the 2nd version makes more sense, so I replaced them all with it.

Also this patch removes questionable WINAPI_FAMILY_PARTITION, CreateFile2, etc stuff from the code.
The file src/share/grabbag/file.c contains non-UWP function GetFileInformationByHandle, and nobody complained, so it's OK to use non-UWP CreateFileA/CreateFileW as well.

### DIFF
--- a/include/share/compat.h
+++ b/include/share/compat.h
@@ -175,10 +175,6 @@
 #define flac_fstat fstat
 #endif
 
-#ifndef WINAPI_FAMILY_PARTITION
-#define WINAPI_FAMILY_PARTITION(x) 0
-#endif
-
 #ifdef ANDROID
 #include <limits.h>
 #endif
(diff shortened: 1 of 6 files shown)


erikd:
Thanks!

PR #83, opened 2018-09-19
Replace hadd with shuffle + add

lvqcl:
According to some ASM optimization manuals, shuffle + add is faster than horizontal add (especially on older processors).


erikd:
Thanks!

PR #95, opened 2019-03-02
Prefer old MMX/ASM functions over new SSE4.1 ones.

lvqcl:
My decoding speed tests show that `FLAC__lpc_restore_signal_intrin_sse41()` and `FLAC__lpc_restore_signal_16_intrin_sse41()` are faster than `FLAC__lpc_restore_signal_asm_ia32()` and `FLAC__lpc_restore_signal_asm_ia32_mmx()` on some CPUs, but slower on others. For example, they are faster on older Intel CPUs, but slower on newer Intel CPUs.

Since these new routines aren't always faster, let's use them only if old MMX/ASM ones are unavailable.


erikd:
Thanks!

PR #103, opened 2019-05-11
Fix Visual Studio solution generation with CMake

evpobr:
Rename flac application target to flacapp to avoid name collision with FLAC library target.


krokoziabla:
LGTM

PR #106, opened 2019-05-29
Add more CMake options

evpobr:
* `BUILD_PROGRAMS`
* `INSTALL_MANPAGES`
* `INSTALL_PKGCONFIG_MODULES`
* `INSTALL_CMAKE_CONFIG_MODULE`

Options are always good. For example, MSVC users do not need PkgConfig modules and man pages.


erikd:
The tests fail with this patch.

src/CMakeLists.txt
```diff
@@ -10,8 +10,10 @@ if(BUILD_CXXLIBS)
     add_subdirectory("libFLAC++")
 endif()
 add_subdirectory("share")
-add_subdirectory("flac")
-add_subdirectory("metaflac")
+if(BUILD_PROGAMS)
```
NotTsunami: BUILD_PROGAMS should be BUILD_PROGRAMS

erikd:
Thanks!

Issue #111, opened 2019-06-26
Any plans to make new release?


kloczek:
I think that it would be good to flush away all already committed changes and make new release with updated code base :)

Barough:
I totally agree with you. All these years of waiting for a new release is just ridicules.

erikd:
@Barough "All these years"? Really? The last release was 2.5 years ago. Let me remind you that before I became the maintainer, FLAC has no new release for 5 or 6 years.

What has gone into the code base since then that you are so desperate for that using a git checkout for is inadequate?

kloczek:
Please do not focus on personal aspects of this conversation.
It really doesn't matter who done what and when.

Can you tell a bit about possible next release? :)
Do you have some plans about fixing some other unknown issues which stops make next release?

erikd:
I do not have any outstanding work that is delaying a release.

I also do not do releases simply because its been a certain amount of time since the last release. However, if someone can say feature X or bug Y was fixed in git since the last release and it would be nice to have a release that includes that, then yes, a release is warranted.

kloczek:
You are right that nothing critical have been fixed in already committed changes however few minor fixes have been committed,
IMO it would bi good to push out all of those changes as new updated release to provide updated code base.
Only this and nothing more.

petterreinholdtsen:
As one of the people involved in Debian, I would like to mention that it
would be great to have a new release of flac released with the six
patches carried by Debian included,
<URL: https://sources.debian.org/src/flac/1.3.2-3/debian/patches/ >,
to make sure the package is in sync between Linux distributions.

<URL: https://tracker.debian.org/pkg/flac >

-- 
Happy hacking
Petter Reinholdtsen

erikd:
@petterreinholdtsen That is a great goal, but how about you go through those patches, decide which ones are appropriate for this repo (some may be Debian specific) and raise PRs for the ones that are relevant?

petterreinholdtsen:
[Erik de Castro Lopo]
> @petterreinholdtsen That is a great goal, but how about you go through
> those patches, decide which ones are appropriate for this repo (some
> may be Debian specific) and raise PRs for the ones that are relevant?

Thank you for the feedback.  I just wanted to bring those patches into
the light, to see how interested the xiph group were in looking at them.
I do not know the flac code base nor these patches enough to be able to
figure out which ones are appropriate for this repo.

I notice from the patch file names there are mostly memory leaks and
privacy issues addressed, but that is about it. :)

[...]

PR #114, opened 2019-07-30
Replace xhtml validation links with travis xmllint

rillian:
Apply Fabian Greffrath's [patch](https://sources.debian.org/src/flac/1.3.2-3/debian/patches/privacy-breach-w3c-valid-html.patch/) from the Debian packages to remove the W3C validator links. They were broken and allow tracking of user activity.

Instead, generate and run the documentation through xmllint as part of the travis-ci.org continuous-integration tests. This gives us more active verification than someone clicking the badge and checking manually.

Addresses discussion in #111.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,9 +17,13 @@ env:
   - BUILD_SYSTEM="cmake"     CONFIGURE_OPTS=
   - BUILD_SYSTEM="cmake"     CONFIGURE_OPTS=-DENABLE_64_BIT_WORDS=ON
 install:
-  - if [ $TRAVIS_OS_NAME = linux ]; then sudo apt-get -y install libtool-bin libogg-dev; fi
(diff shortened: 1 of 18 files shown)


rillian:
Two issues I noticed testing:
- Doxygen output passes on travis (Ubuntu 16.04) but fails with duplicate `id` attributes on my Fedora 30 system.
- The docs [don't actually pass](https://validator.w3.org/check?uri=https%3A%2F%2Fxiph.org%2Fflac) the current validator.w3c.org, which wants an xmlns attribute on the top-level element. I'm happy to add that if you like.

I'd also say that the whole XML validation thing wasn't a successful technology, and we'd be better off porting everything to modern html, whose flexibility is more appropriate for document interchange. Nothing wrong with lints either way.

erikd:
@rillian I would be more than happy with either:

> which wants an xmlns attribute on the top-level element. I'm happy to add that if you like.

or porting everything to modern html.

erikd:
Although, I will merge this as it is.

PR #117, opened 2019-08-04
Do not assume that sys/param.h defines MIN and MAX

sezero:
One such example is sys/param.h from MinGW.


erikd:
Going to merge this and make sure it works in a separate PR.

PR #119, opened 2019-08-05
Version 1.3.4

erikd:
Verion 1.3.3 was borked because the version number in CMakeLists.txt did
not get updated.


EatonZ:
@erikd What's the story with this? Looks like v1.3.3 eventually came out.

rillian:
This should be addressed on the 1.3.x branch now.

Issue #120, opened 2019-08-05
exported symbols and dll builds


sezero:
libFLAC.dll built by MSVC exports only the symbols marked as FLAC_API,
because FLAC_API is defined properly as FLAC_API_EXPORTS in export.h.
libFLAC.dll built by MinGW exports all of the symbols it has, because
FLAC_API_EXPORTS is restricted to MSVC in export.h: this is obviously
wrong.

As a test, I tried something like this:
```
diff --git a/include/FLAC/export.h b/include/FLAC/export.h
index d52f0bb..628fe5f 100644
--- a/include/FLAC/export.h
+++ b/include/FLAC/export.h
@@ -59,7 +59,7 @@
 #if defined(FLAC__NO_DLL)
 #define FLAC_API
 
-#elif defined(_MSC_VER)
+#elif defined(_WIN32)
 #ifdef FLAC_API_EXPORTS
 #define	FLAC_API __declspec(dllexport)
 #else
diff --git a/configure.ac b/configure.ac
index 0228a12..c2edf0a 100644
--- a/configure.ac
+++ b/configure.ac
@@ -493,6 +493,17 @@ if test x$enable_stack_smash_protection = "xyes" ; then
 	XIPH_GXX_STACK_PROTECTOR
 	fi
 
+if test x$enable_shared != "xyes" ; then
+	CPPFLAGS="-DFLAC__NO_DLL $CPPFLAGS"
+	fi
+
+AH_VERBATIM([FLAC_API_EXPORTS],
+[/* libtool defines DLL_EXPORT for windows dll builds,
+   but flac code relies on FLAC_API_EXPORTS instead. */
+#ifdef DLL_EXPORT
+# define FLAC_API_EXPORTS
+#endif])
+
 AC_CONFIG_FILES([ \
 	Makefile \
 	src/Makefile \
```

But that doesn't work, because flac.exe and metaflac.exe unfathomably
rely on the private symbols of windows_unicode_filenames.c which only
go into the dll and wrongly exported.  MSVC builds aren't affected by
this, because those exes are linked to static libFLAC and not to the
dll.

What is a good / acceptable solution for this?

erikd:
I don't use Windows. Maybe @lvqcl can help.

lvqcl:
I don't know much how MinGW or configure.ac work, so I cannot help here.

Maybe it makes sense to add FLAC_API_EXPORTS to those functions from windows_unicode_filenames.c ?

Or leave it as is (if it's not a problem)?

[...]

PR #122, opened 2019-08-05
make windows dll builds export only the public symbols.

sezero:
this solves https://github.com/xiph/flac/issues/120

This is an initial patch that solves https://github.com/xiph/flac/issues/120
Tested by cross-compiling on linux using autotools and cmake. MSVC
changes need testing.


lvqcl:
For some reason it breaks encoding of files with non-ASCII filenames:

input filename: RAiNY STARS.wav
output filename: RAiNY STARS.flac

input filename: Иероглиф.wav
output filename: РРµСЂРѕРіР»РёС„.flac

input filename: とびだせ！バンキッキ.wav
output filename: гЃЁгЃігЃ гЃ›пјЃгѓђгѓіг‚­гѓѓг‚­.flac

sezero:
> For some reason it breaks encoding of files with non-ASCII filenames:

I didn't touch any of the code itself.  Did it work fine before?

lvqcl:
Yes, it did work.

windows_unicode_filenames.c contains static variable `utf8_filenames`.

flac.exe calls `get_utf8_argv()`, which in turn calls `flac_set_utf8_filenames()`, which sets this variable to `true`. After this, libFLAC treats filenames as UTF8-encoded.

Maybe currently there are 2 static variables, one in flac.exe and another in libFLAC.dll.

In this case flac.exe changes its own copy, the variable in libFLAC.dll remains unchanged, and libFLAC treats filenames as ANSI-encoded.

sezero:
> windows_unicode_filenames.c contains static variable `utf8_filenames`.
> 
> flac.exe calls `get_utf8_argv()`, which in turn calls `flac_set_utf8_filenames()`, which sets this variable to `true`. After this, libFLAC treats filenames as UTF8-encoded.
> 
> Maybe currently there are 2 static variables, one in flac.exe and another in libFLAC.dll.
> 
> In this case flac.exe changes its own copy, the variable in libFLAC.dll remains unchanged, and libFLAC treats filenames as ANSI-encoded.

Well, I'm stuck there.. Do you have a solution?

sezero:
One solution I can think of is adding a constructor function to dll builds of libFLAC.
Something like:
```
#ifdef FLAC_API_EXPORTS
void __attribute__((constructor)) init_utf8_filenames(void) {
  typedef int (__cdecl *wgetmainargs_t)(int*, wchar_t***, wchar_t***, int, int*);
  wgetmainargs_t wgetmainargs;
  HMODULE handle;

  if ((handle = LoadLibraryW(L"msvcrt.dll")) == NULL) return;
  if ((wgetmainargs = (wgetmainargs_t)GetProcAddress(handle, "__wgetmainargs")) != NULL) {
      flac_set_utf8_filenames(true);
  }
  FreeLibrary(handle);
}
#endif
```

Would this work?  (not tested..)

Yes, get_utf8_argv() in win_utf8_io.c refuses to set utf8_filenames
depending on more failures (such as malloc failure which should 
actually result in an abort itself but whatever), but this should be
enough for a dll if it works.  (and if it works, we'll need an MSVC
equivalent of constructor attribute.)

sezero:
The above seems to work for me (TM).

lvqcl:
TWIMC - the first message about Unicode support for Windows version: http://lists.xiph.org/pipermail/flac-dev/2013-March/003889.html

UTF8 mode was made a runtime choice: http://lists.xiph.org/pipermail/flac-dev/2013-April/003973.html

Basically, the reason why `utf8_filenames` is not `true` by default is: 

> The reason I had them this way is that FLAC API has public functions that work with filenames. Default compile options for FLAC would be unaffected and existing programs that rely on ANSI filenames would continue to work. The new configuration option "Release (UTF-8)" was meant to isolate the changes so that only official bundled frontends that know about the changes would use them.

sezero:
So, what is it that you are proposing?

Should we add a public libFLAC api like `FLAC_set_utf8_filenames(FLAC__bool)`
or something?

lvqcl:
Honestly, I don't know, and I don't care much about it. For me, everything is OK as long as it doesn't break Unicode support.

I also don't care much about FLAC API - it's a mess already... :(

sezero:
OK then.  Well the main idea is here, @erikd should decide what to do.
Or I can can this and let things stay as they are.

sezero:
I did as best as I could, so I'm open to ideas.

PR #124, opened 2019-08-08
libFLAC: Rename get_client_data_from_decoder function

erikd:
This function was added after the 1.3.2 release, but the name was bad
(missing "flac" at the start) and the pointer returned should be 'const'.

### DIFF
--- a/src/libFLAC/include/protected/stream_decoder.h
+++ b/src/libFLAC/include/protected/stream_decoder.h
@@ -53,13 +53,13 @@ typedef struct FLAC__StreamDecoderProtected {
 } FLAC__StreamDecoderProtected;
 
 /*
- * return the number of input bytes consumed
+ * Return the number of input bytes consumed
  */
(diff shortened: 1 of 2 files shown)


sezero:
Is this not public (it is marked `FLAC_API`)?  If so,
- the prefix should be `FLAC__`  (with two underscores) instead of just `flac_`
- the proto should be moved to the top-level public include/FLAC/stream_decoder.h
  so that the users would be aware of it.

As for const return, I hope that matches the intended use by @ChristopD

erikd:
@ChristopD needs a public functions, so it needs to be one. Its `const` because while its fine to hold a pointer to and internal FLAC data structure, its not fine to modify anything it points to.

I will fix the naming.

ChristopD:
I just check the updated master/321f9ea, and everything is working fine for us, with a minimal renaming task. We are ready for next release.
Thank you.

tledoux:
This new method should probably be documented in the changelog in the library section.

PR #126, opened 2019-08-15
Fix build on FreeBSD ppc

pkubaj:
FreeBSD uses elf_aux_info function.

### DIFF
--- a/src/libFLAC/cpu.c
+++ b/src/libFLAC/cpu.c
@@ -250,6 +250,8 @@ ppc_cpu_info (FLAC__CPUInfo *info)
 	hwcaps = getauxval(AT_HWCAP2);
 #elif defined (__FreeBSD__)
 	elf_aux_info(AT_HWCAP2, &hwcaps, sizeof(hwcaps));
+#else
+#error Unsupported platform! Please add support for reading ppc hwcaps.
 #endif
 	
 	if (hwcaps) & PPC_FEATURE2_ARCH_3_00) {
--- a/src/libFLAC/cpu.c
+++ b/src/libFLAC/cpu.c
@@ -245,11 +245,21 @@ ppc_cpu_info (FLAC__CPUInfo *info)
 #define PPC_FEATURE2_ARCH_2_07		0x80000000
 #endif
 
+#ifdef __linux__
 	if (getauxval(AT_HWCAP2) & PPC_FEATURE2_ARCH_3_00) {
 		info->ppc.arch_3_00 = true;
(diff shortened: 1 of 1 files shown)


src/libFLAC/cpu.c
```diff
@@ -245,9 +245,16 @@ ppc_cpu_info (FLAC__CPUInfo *info)
 #define PPC_FEATURE2_ARCH_2_07		0x80000000
 #endif
 
-	if (getauxval(AT_HWCAP2) & PPC_FEATURE2_ARCH_3_00) {
+	long hwcaps;
+#ifdef __linux__
+	hwcaps = getauxval(AT_HWCAP2);
+#elif defined (__FreeBSD__)
+	elf_aux_info(AT_HWCAP2, &hwcaps, sizeof(hwcaps));
+#endif
```
rillian: What about adding
```
#else
        #error Unsupported platform! Please add support for reading ppc hwcaps.
```
so the code doesn't proceed with uninitialized data if someone happens to build on a third ppc platform?

rillian:
The travis test failure is unrelated. I've submitted #127 to address it.

src/libFLAC/cpu.c
pkubaj: OK, that will it easier for users of other systems.

rillian: Thanks!

src/libFLAC/cpu.c
```diff
@@ -245,9 +245,18 @@ ppc_cpu_info (FLAC__CPUInfo *info)
...
+	long hwcaps;
+#ifdef __linux__
+	hwcaps = getauxval(AT_HWCAP2);
+#elif defined (__FreeBSD__)
+	elf_aux_info(AT_HWCAP2, &hwcaps, sizeof(hwcaps));
+#else
+#error Unsupported platform! Please add support for reading ppc hwcaps.
+#endif
+	
+	if (hwcaps) & PPC_FEATURE2_ARCH_3_00) {
```
rillian: It looks liek there's an unbalanced `)` here. Does this compile?

pkubaj: You're right. Github makes it hard to paste code (since some time, it worked fine before) in its editing window so I just type code manually for short PR's...

rillian: I see, yes. Normally travis would catch that sort of thing, but we don't test building for ppc.

erikd:
Squashed these commits in a single one and improved the commit message.

PR #128, opened 2019-08-18
Switch to utimensat for newer POSIX versions

neheb:
Some libcs like uClibc-ng can optionally disable deprecated functions.
utime is one of them. When done so, both the header and the function go
missing.

This fixes flac_utime to work in such a situation.

### DIFF
--- a/include/share/compat.h
+++ b/include/share/compat.h
@@ -112,8 +112,12 @@
 #include <sys/utime.h> /* for utime() */
 #endif
 #else
+#if defined(_POSIX_C_SOURCE) && (_POSIX_C_SOURCE >= 200809L)
+#include <fcntl.h>
+#else
 #include <sys/types.h> /* some flavors of BSD (like OS X) require this to get time_t */
 #include <utime.h> /* for utime() */
+#endif
 #endif
 
 #if defined _MSC_VER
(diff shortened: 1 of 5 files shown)


neheb:
Hmm weird. _POSIX_C_SOURCE is undefined in metadata_manip.c. Doesn't happen locally.

erikd:
Build failure https://travis-ci.org/xiph/flac/jobs/573487312 :
```
In file included from ../../include/share/grabbag/file.h:35:0,
                 from ../../include/share/grabbag.h:25,
                 from metadata_manip.c:39:
metadata_manip.c: In function ‘set_file_stats_’:
../../include/share/compat.h:172:50: error: incompatible type for argument 3 of ‘utimensat’
 #define flac_utime(a, b) utimensat (AT_FDCWD, a, *b, 0)
                                                  ^
metadata_manip.c:267:8: note: in expansion of macro ‘flac_utime’
  (void)flac_utime(filename, &srctime);
        ^
In file included from metadata_manip.c:35:0:
/usr/include/x86_64-linux-gnu/sys/stat.h:363:12: note: expected ‘const struct timespec *’ but argument is of type ‘struct utimbuf’
 extern int utimensat (int __fd, const char *__path,
            ^
Makefile:470: recipe for target 'metadata_manip.o' failed
```

neheb:
I'm actually testing on Linux with musl and uclibc-ng. Not with glibc. I have no idea why it's failing.

erikd:
Well the Travis VMs all use glibc. Obviously you need to test with glibc. Maybe you could spin up glibc based Linux system in a VM.

neheb:
I see the problem. I don't build tests.

erikd:
Oh, cool. If you fix this, please make sure to rebase your commits into single commit.

neheb:
Do the C++ bindings target C++98 or C++11? The utime stuff can be replaced with https://en.cppreference.com/w/cpp/thread/sleep_for if the latter.

erikd:
This change should be as small as possible and only address the actual issue.

neheb:
Should build now.

Replacing utime/utimensat with https://en.cppreference.com/w/cpp/thread/sleep_for is simpler but requires C++11.

erikd:
Lets see what CI says :)

neheb:
Passes

PR #130, opened 2019-08-29
Run libFLAC++ through clang-tidy.

neheb:
Applied the following suggestions:

modernize-use-auto
modernize-use-equals-default
modernize-deprecated-headers
google-readability-casting
google-readability-namespace-comments
readability-else-after-return


neheb:
Succeeds now. Removed a few of the changes since they require C++11.

erikd:
Thanks!

PR #131, opened 2019-08-29
window: fabs > fabsf and cos > cosf where it makes sense

neheb:
The f functions are faster. It seems the arguments are all floats.


erikd:
Thanks!

PR #132, opened 2019-08-30
Change FLAC__bool to bool for MSVC 2015 an above and elsewhere.

neheb:
MSVC 2015 introduced bool. Use it.

### DIFF
--- a/include/FLAC/ordinals.h
+++ b/include/FLAC/ordinals.h
@@ -66,7 +66,12 @@ typedef uint64_t FLAC__uint64;
 
 #endif
 
+#if defined(_MSC_VER) && _MSC_VER < 1900
 typedef int FLAC__bool;
+#else
+#include <stdbool.h>
+typedef bool FLAC__bool;
+#endif
 
 typedef FLAC__uint8 FLAC__byte;
 
--- a/src/metaflac/options.c
+++ b/src/metaflac/options.c
@@ -127,9 +127,7 @@ void init_options(CommandLineOptions *options)
 {
 	options->preserve_modtime = false;
 
-	/* '2' is a hack to mean "use default if not forced on command line" */
-	FLAC__ASSERT(true != 2);
-	options->prefix_with_filename = 2;
+	options->prefix_with_filename = true;
 
 	options->utf8_convert = true;
 	options->use_padding = true;
@@ -176,7 +174,7 @@ FLAC__bool parse_options(int argc, char *argv[], CommandLineOptions *options)
 		}
 	}
 
-	if(options->prefix_with_filename == 2)
+	if(options->prefix_with_filename == true)
 		options->prefix_with_filename = (argc - share__optind > 1);
 
 	if(share__optind >= argc && !options->show_long_help && !options->show_version) {


neheb:
Fails because of a "hack":

options.c:179:35: warning: comparison of constant ‘2’ with boolean expression is always false [-Wbool-compare]

  if(options->prefix_with_filename == 2)

sezero:
If you do this, you will possibly make FLAC__bool to be different in C and C++
and also different across different platforms.

neheb:
book is an int in C++?

neheb:
Actually...I first saw this problem when running clang-tidy on libFLAC++. It was telling me to cast true/false to FLAC__bool, which is an int.

sezero:
> bool is an int in C++?

That came out wrong.  What I mean is what we used to have is
an int==4 bytes, where as stdbool.h bool is very most likely is an
unsigned char==1 byte: this will change the api badly.

neheb:
bool is actually one bit, not byte. I see it all the time when turning on -Wpadded in other projects.

sezero:
> bool is actually one bit, not byte. I see it all the time when turning on -Wpadded in other projects.

Printing sizeof(bool) output 1.  (And what you say doesn't excuse api breakage.)

erikd:
> bool is actually one bit, not byte. I see it all the time when turning on -Wpadded in other projects.

Logically `bool` is only one bit, but the in memory representations of `bool` are probably never as a single bit.

neheb:
I just checked again using pahole. Seems you're right. This is output from a linux kernel driver:

```
struct ag71xx_dcfg {
	u32                        max_frame_len;        /*     0     4 */
	const u32  *               fifodata;             /*     4     4 */
	u16                        desc_pktlen_mask;     /*     8     2 */
	bool                       tx_hang_workaround;   /*    10     1 */

	/* XXX 1 byte hole, try to pack */
```

or maybe that just bool in the kernel.

erikd:
@nehad:
> Actually...I first saw this problem when running clang-tidy on libFLAC++. It was telling me to cast true/false to FLAC__bool, which is an int.

Tools like that can get it wrong, particularly when on mixed C / C++ code (eg C++ code calling C code). You should use what it says as a guide and figure why its say what it is saying.

@sezero:
> What I mean is what we used to have is an int==4 bytes, where as stdbool.h bool is very most likely is an unsigned char==1 byte: this will change the api badly.

I think you are getting A*P*I confused with A*B*I. The former says what the library does on a logical basis while the later is makes assumptions about memory layout and data type sizes.

> or maybe that just bool in the kernel.

All C types have a `sizeof` at least 1 which means 1 byte.

sezero:
> I think you are getting A_P_I confused with A_B_I. The former says
> what the library does on a logical basis while the later is makes
> assumptions about memory layout and data type sizes.

I'm aware, but was lazy to edit my post.

So, what is your decision? I hope this doesn't go in.

erikd:
I have removed my approval. I will not re-approve unless one of the following happens:

* We decide that an ABI change is acceptable and bump the library version number appropriately (probably some other things we should fix in the API/ABI at that point as well).
* We can prove that merging this is does not result in an ABI change.

PR #133, opened 2019-08-30
flac/utils: Add missing define

neheb:
wcswidth needs _XOPEN_SOURCE defined. Otherwise, no declaration.


neheb:
/home/mangix/devstuff/flac/src/flac/utils.c:210:8: warning: implicit declaration of function 'wcswidth' is invalid in C99 [clang-diagnostic-implicit-function-declaration]
        len = wcswidth(wtmp, len);

erikd:
Thanks!

Issue #44, opened 2017-10-20
Conflicting types in source and header files


razed11:
I posted this last night to the mailing list. Reposting here:

I'd like to compress a 32 kHz 16-bit mono using an STM32F4 microcontroller (RTOS not Linux). My plan is to capture from 500-1000 ms of data, compress, and send the result over a UART.

I have the arm-none-eabi tools installed on MacOS. I used homebrew to install autoconf, automake, and libtool. autogen.sh succeeded and I was able to create my makefiles using the following:

```
./configure --build=x86_64-apple-darwin16.7.0 --host=arm-none-eabi --prefix=$PWD/export CFLAGS="-O3 -mthumb -mcpu=cortex-m4 -lc -lrdimon -specs=rdimon.specs" --disable-cpplibs --disable-ogg --disable-oggtest --disable-examples --disable-xmms-plugin
```

```
  Configuration summary :

    FLAC version : ........................ 1.3.2

    Host CPU : ............................ arm
    Host Vendor : ......................... none
    Host OS : ............................. eabi

    Compiler is GCC : ..................... yes
    GCC version : ......................... 5.4.1
    Compiler is Clang : ................... no
    SSE optimizations : ................... yes
    Asm optimizations : ................... no
    Ogg/FLAC support : .................... no
```

I'm getting some errors around type mismatches between the declarations and definitions for example:

```
grabbag/cuesheet.c:32:10: error: conflicting types for 'grabbag__cuesheet_msf_to_frame'
 uint32_t grabbag__cuesheet_msf_to_frame(uint32_t minutes, uint32_t seconds, uint32_t frames)
          ^
In file included from ../../include/share/grabbag.h:24:0,
                 from grabbag/cuesheet.c:29:
../../include/share/grabbag/cuesheet.h:32:10: note: previous declaration of 'grabbag__cuesheet_msf_to_frame' was here
 unsigned grabbag__cuesheet_msf_to_frame(unsigned minutes, unsigned seconds, unsigned frames);
```

I can see that the variables in the header are not declared using the <stdint.h> values.

[...]

PR #135, opened 2019-09-02
Fix function redefinition on arm-none-eabi compilers

Oreo639:
This should fix https://github.com/xiph/flac/issues/44
While on most systems unsigned does equal 4bytes, this is not true on all systems which is why arm-none-eabi complains about a function redefinition. Not sure if it happens with other compilers, but this is the one I ran into the issue on.

Feel free to let me know if there are any issues with this pr.

### DIFF
--- a/include/share/grabbag/cuesheet.h
+++ b/include/share/grabbag/cuesheet.h
@@ -29,10 +29,10 @@
 extern "C" {
 #endif
 
-unsigned grabbag__cuesheet_msf_to_frame(unsigned minutes, unsigned seconds, unsigned frames);
-void grabbag__cuesheet_frame_to_msf(unsigned frame, unsigned *minutes, unsigned *seconds, unsigned *frames);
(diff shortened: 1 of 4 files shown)


neheb:
What does int equal on armeb out of curiosity?

Oreo639:
The target arch I am compiling for has unsigned equal 4bytes. (Since it's a 32bit system)

erikd:
The vast majority of current systems use 32 bit `int`. However the actual size of `int` is not specified in the standard. This means that use of types like `int32_t` and `uint32_t` make sense because they specify *exactly* how many bits the type has.

Systems with `int`s smaller than 32 bits are quite rare now, but all Windows versions before Windows 95 had 16 but `int` as did most 16 bit embedded OSes.

neheb:
makes sense then i guess. MIPS16 has int at 32 bits i believe.

erikd:
Thanks!

PR #142, opened 2019-09-29
configure.ac: check for sys/auxv.h before defining FLAC__CPU_PPC

pseiderer:
Signed-off-by: Peter Seiderer <ps.report@gmx.net>
---
Note: One of the buildroot autobuilder provided powerpc toolchains
seem to not provide the sys/auxv.h header file, resulting in a
compile failure (see [1] for details).

[1] http://lists.busybox.net/pipermail/buildroot/2019-September/259732.html


erikd:
Thanks!

PR #146, opened 2019-10-08
cmake: fix nasm sources build when building in a separate directory

sezero:
When building in the same directory as the source, it works, but not in
a separate directory.  The trailing slash is important with nasm.  Note
that: include_directories("${CMAKE_CURRENT_SOURCE_DIR}/") does not work
because cmake seems to strip the trailing slash.

Also relax the cmake version requirement from 3.12 to 3.9 as everywhere
else in the tree: it works just fine.


Chocobo1:
I'm having build issues related to this PR, see #220.

PR #147, opened 2019-10-08
make dllexport work with compilers other than MSVC

sezero:
This is a minimal solution for https://github.com/xiph/flac/issues/120:
@lvqcl, @erikd: please review and test.

This changeset makes dllexport work with compilers other than MSVC.
The issue is, flac and metaflac exes rely on flac_internal_???_utf8()
procedures from windows_unicode_filenames.c and there is no easy way
to exclude them from exports without breaking things. So export them
explicitly (they are exported anyway w/o this patch), but add a FIXME
note about the kludge in windows_unicode_filenames.c.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -493,6 +493,22 @@ if test x$enable_stack_smash_protection = "xyes" ; then
 	XIPH_GXX_STACK_PROTECTOR
 	fi
 
+AH_VERBATIM([FLAC_API_EXPORTS],
+[/* libtool defines DLL_EXPORT for windows dll builds,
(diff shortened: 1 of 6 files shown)


sezero:
> Please squash the top commit (fix libFLAC++ dll build) down into
> whichever of the commits seems most appropriate.

Done.

sezero:
NOTE: this changeset can theoretically be propagated to cygwin too, e.g.
`#elif defined(_WIN32) || defined(__CYGWIN__)` in export.h, but I can't
test it.

erikd:
If it can't be tested its probably now worth it. I think Cygwin is pretty much dead these days.

PR #148, opened 2019-10-08
configury: allow several compiler flags for clang

sezero:
The flac configury marked clang as 'not being gcc' and excluded
a lot of compiler switches, most importantly the visibility flags,
from being used with it. This was done possibly after a problem
reported at: https://github.com/erikd/libsndfile/issues/49 .

This patch does the following:

- m4/gcc_version.m4 (XIPH_GCC_VERSION): set GCC_MAJOR_VERSION and
  GCC_MINOR_VERSION to 0 for non-gcc. Previously, they were left
  unset.

- configure: the gcc version checks are, naturally, against non-
  zero values, so, allow many compiler switches to be used with
  clang without affecting real-gcc cases.

- configure: When setting CFLAGS="-O3 -funroll-loops", also set
  CXXFLAGS="-O3".  Prevents g++ warnings with _FORTIFY_SOURCE, i.e.:
 '_FORTIFY_SOURCE requires compiling with optimization (-O)'

Tested compilation using gcc-7.3.1 and clang-5.0.2 on x86_64-linux,
and gcc-4.4.7 and clang-3.4.2 on an i686-linux.  Also tested cross-
compiling for Mac OS X using clang-5.0.2.


sezero:
With this, the only warnings from clang-5.0.2 are -Wcast-align warnings.
The stderr output for it is attached here, in case anyone needs it.

[clang-Wcast-align.txt](https://github.com/xiph/flac/files/3703770/clang-Wcast-align.txt)

sezero:
> This looks good, but can you fix the commit message "configury" -> "configure.ac".

Done.

sezero:
> > This looks good, but can you fix the commit message "configury" -> "configure.ac".
> 
> Done.

@erikd: do you need anything else with this?

erikd:
Nothing to be done. I just want to batch this with some other FLAC PRs. That will happen over the weekend.

Issue #152, opened 2019-10-17
FLAC failed to build due to error C2375 on windows with MSVC


spacelg:
Environment:
Windows Server 2016 + VS2017 + flac master branch latest source code.

Issue description:
I tried to build flac using VS2017 on Windows Server 2016. We found that flac failed to build due to error C2375: 'flac_internal_set_utf8_filenames': redefinition; different linkage.  It can be first reproduced on b917d45 revision. Could you please help take a look at this? Thanks in advance!

Repro steps:
1. git clone https://github.com/xiph/flac D:\FLAC\src
2. Open a VS 2017 x86 command prompt as admin and browse to D:\FLAC\src
3. devenv /upgrade FLAC.sln
4. msbuild FLAC.sln /t:Rebuild /p:Configuration=Release;Platform=Win32 /p:WindowsTargetPlatformVersion=10.0.17134.0

[log_x86_build.log](https://github.com/xiph/flac/files/3737259/log_x86_build.log)

Actual result:
D:\FLAC\src\src\libFLAC\windows_unicode_filenames.c(67): error C2375: 'flac_internal_set_utf8_filenames': redefinition; different linkage [D:\FLAC\src\src\libFLAC\libFLAC_dynamic.vcxproj]
D:\FLAC\src\src\libFLAC\windows_unicode_filenames.c(72): error C2375: 'flac_internal_get_utf8_filenames': redefinition; different linkage [D:\FLAC\src\src\libFLAC\libFLAC_dynamic.vcxproj]
D:\FLAC\src\src\libFLAC\windows_unicode_filenames.c(79): error C2375: 'flac_internal_fopen_utf8': redefinition; different linkage [D:\FLAC\src\src\libFLAC\libFLAC_dynamic.vcxproj]
D:\FLAC\src\src\libFLAC\windows_unicode_filenames.c(101): error C2375: 'flac_internal_stat64_utf8': redefinition; different linkage [D:\FLAC\src\src\libFLAC\libFLAC_dynamic.vcxproj]
D:\FLAC\src\src\libFLAC\windows_unicode_filenames.c(117): error C2375: 'flac_internal_chmod_utf8': redefinition; different linkage [D:\FLAC\src\src\libFLAC\libFLAC_dynamic.vcxproj]
[...]

PR #153, opened 2019-10-18
fix msvc build after commit b917d456 (closes github bug #152)

sezero:
Tested build with Visual Studio 2017 (version 15.9.17) and MinGW-w64.

### DIFF
--- a/include/share/windows_unicode_filenames.h
+++ b/include/share/windows_unicode_filenames.h
@@ -39,21 +39,24 @@
 #include <sys/utime.h>
 #include "FLAC/ordinals.h"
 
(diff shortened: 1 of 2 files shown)


include/share/windows_unicode_filenames.h
```diff
@@ -39,21 +39,24 @@
 #include <sys/utime.h>
```
erikd: Looks good. Can you just change the title to be just "Fix msvc build after commit b917d45" and then add:
```
Closes: https://github.com/xiph/flac/issues/152
```
to the commit message body?

sezero: Done.

PR #154, opened 2019-10-23
libFLAC/stream_decoder.c: fix integer overflow on corrupt file

hnktmobile:
Corrupt metadata could make the length calculation overflow.

This issue was found when reading a corrupt flac file when ubsan (undefined behaviour sanitizer) is enabled for the compiler.


hnktmobile:
In our case length was 0, causing the subtraction to result in a negative number.
Storing a negative number in an unsigned variable caused a crash since ubsan was enabled.

erikd:
Nice work. Thanks!

PR #159, opened 2019-11-11
libFLAC/bitreader.c: Fix shift invoking undefined behaviour

erikd:
Credit: Oss-Fuzz
Issue: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=18535
Testcase: fuzzer_decoder-6573800707063808


erikd:
If I build this locally with GCC 9, all the tests pass. The failing tests in CI are being built with gcc-5.4.0 and some unknown version of clang.

erikd:
This is only a problem when using the `--enable-64-bit-words` configure option. :(

PR #161, opened 2019-11-13
cmake_minimum_required() cleanup:

sezero:
- document version requirements in the top-level file.
- remove incorrect version requirements from sub-dirs.
- set min. required version to 3.5 in top-level file.
- set minimun required version to 3.9 under docs/ .
- make documents building an option (on by default.)
  so, the tree can be built using cmake-3.5 using
 -DBUILD_DOCS=0 on the command line.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,8 @@
-cmake_minimum_required(VERSION 3.9)
+# 3.1 is OK for most parts. However:
+# 3.3 is needed in src/libFLAC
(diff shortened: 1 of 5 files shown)


sezero:
Build tested on CentOS-6.10/i686 using cmake 3.5.0 (w/o docs) and 3.9.0,
and on Fedora 27/x86_64 using cmake 3.11.2.

erikd:
Thanks!

PR #165, opened 2019-11-17
cmake: use proper pkg-config install path variable

### LINKED ISSUE #160: Question about the pkg-config files install paths when using CMake
In the cmake config for libflac++ and libflac the .pc files get installed to:
`${CMAKE_INSTALL_DATAROOTDIR}/pkgconfig`
`${CMAKE_INSTALL_DATADIR}/pkgconfig`

Shouldn't it be `${CMAKE_INSTALL_LIBDIR}/pkgconfig`
Example from your other project `opus`:
```
 install(FILES ${CMAKE_CURRENT_BINARY_DIR}/opus.pc
          DESTINATION ${CMAKE_INSTALL_LIBDIR}/pkgconfig) 
```
Seems datarootdir and datadir resolve to `share/` instead of the needed `lib/` where everyone else stores their pkgconfig files, including opus..

Sources:
https://github.com/xiph/flac/blob/master/src/libFLAC%2B%2B/CMakeLists.txt#L53
https://github.com/xiph/flac/blob/master/src/libFLAC/CMakeLists.txt#L129

Patch:
https://github.com/DeadSix27/flac/commit/d8f189ac0e5d0f051e5d844bc21c253ce505e18e

If this is indeed an issue I can provide a PR if wanted.

DeadSix27:
As discussed in #160 
This changes the target location variable for both flac++.pc and flac.pc to the ~~proper~~ more common one.

Resolves #160


erikd:
@evpobr @sezero @krokoziabla I would appreciate a review from one or more of you.

DeadSix27:
Fixed commit message.

neheb:
This change is correct.

evpobr:
From `man pkg-config`:

> pkg-config retrieves information about packages from special metadata files. These files are named after the package, and has a .pc extension. On most systems, pkg-config looks in /usr/lib/pkgconfig, /usr/share/pkgconfig, /usr/local/lib/pkgconfig and /usr/local/share/pkgconfig for these files. It will additionally look in the colon-separated (on Windows, semicolon-separated) list of directories specified by the PKG_CONFIG_PATH environment variable

So both old and new paths are correct. But I vote for this, `lib`  is much more popular.

PR #166, opened 2019-11-18
doc: Update to Doxygen 1.7.6.1

NotTsunami:
Brings us a bit closer to the modern world and silences some warnings in doxygen output.


NotTsunami:
An alternative to the #define changes is to wrap them in @code and @endcode, in which case Doxygen will ignore the number signs.

erikd:
I don't think enclosing them in `code` blocks is the right thing to do. @rillian any thoughts?

NotTsunami:
I don't like that either which is why I proposed it just removing the number sign. Neither of these changes are even close to mandatory, but nice to have to silence Doxygen output and bring things forward a bit.

erikd:
I think it would be better with this: http://www.doxygen.nl/manual/commands.html#cmdhash :)

NotTsunami:
!! Learned something new, I'll amend and push after uni.

erikd:
I thought there had to be a way to do it and that it has to be documented :).

NotTsunami:
1.8 deprecates further functions so some deprecation warnings still show, but fewer than before:
```
warning: Tag `SYMBOL_CACHE_SIZE' at line 328 of file `Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag `SHOW_DIRECTORIES' at line 548 of file `Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag `HTML_ALIGN_MEMBERS' at line 943 of file `Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag `USE_INLINE_TREES' at line 1134 of file `Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag `XML_SCHEMA' at line 1397 of file `Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag `XML_DTD' at line 1403 of file `Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
```

The warnings in the second commit no longer show. Nothing immediate, but maybe in the near future we can upgrade to 1.8.x and clean up documentation a bit more.

erikd:
Thanks!

PR #167, opened 2019-11-19
cmake/configure.ac: Enable -fstack-protector-strong by default

NotTsunami:
Windows is excluded from this (MSVC in CMake).

If the general idea is accepted, I'll test across systems and clean up any fuzz.


erikd:
If this passes in CI I would be in favor.

NotTsunami:
> If this passes in CI I would be in favor.

It did just pass Travis. In the morning I'll add the missing console output for its status since it's a global opt-out now.

NotTsunami:
> Please squash this PR down to a single commit, line wrap commit commets to < 70 chars (other than URLs). Feel free to extent the commit message with the other commit messages as points (this is better done manually because GH does a rubbish job of doing this automatically).

Done.

erikd:
Thanks!

PR #168, opened 2019-11-19
lib.mk: Remove incorrect u flag from ar

NotTsunami:
We work in deterministic mode by default, so timestamps are zeroed, thus it is impossible to check timestamps and insert only newer members. Silences the following autotools warning:

ar: `u' modifier ignored since `D' is the default (see `U')

https://sourceware.org/binutils/docs/binutils/ar-cmdline.html


erikd:
Thanks!

PR #174, opened 2019-11-30
Fixed error condition processing

elnull:
Value of variable `ok` was always overwritten with result of `encoder.set_metadata(metadata, 2)` so it may never set to `false` here.


rillian (approved):

rillian:
Thanks for the patch. I couldn't figure out how to download it, since the submission repo has been deleted, but I pushed a reconstructed version as a9557c13fa78d73bee0ef44a08057bff6d88acee.

PR #175, opened 2019-11-30
Input file name size calculated just one time

elnull:


erikd:
Merged in #184. Thanks!

PR #176, opened 2019-11-30
Corrected printf specifiers to unsigned int where needed

elnull:


erikd:
Merged in https://github.com/xiph/flac/pull/185 . Thanks!

PR #179, opened 2019-11-30
Delete decoder and encoder before return if they're not valid

elnull:


erikd:
Thanks!

PR #180, opened 2019-11-30
Fixed usage of gainfilter variable after free

elnull:


erikd:
Thanks!

PR #181, opened 2019-11-30
Variable removed from expression because it's always equals to false

elnull:
Variable `got_data_chunk` was already checked at line https://github.com/xiph/flac/blob/6455e477218360899c55f8dbd06c6628260d4123/src/flac/encode.c#L207
then it must be always false at this line https://github.com/xiph/flac/blob/6455e477218360899c55f8dbd06c6628260d4123/src/flac/encode.c#L226 so it can be safely removed.


erikd:
Thanks.

PR #182, opened 2019-11-30
Removed useless comparison

elnull:
Line https://github.com/xiph/flac/blob/6455e477218360899c55f8dbd06c6628260d4123/src/libFLAC/stream_decoder.c#L3303 have comparison which negative condition equals to https://github.com/xiph/flac/blob/6455e477218360899c55f8dbd06c6628260d4123/src/libFLAC/stream_decoder.c#L3320


erikd:
Thanks!

PR #188, opened 2019-12-21
CMake: a handier way to find Iconv

krokoziabla:


erikd:
@evpobr This seem sensible to you?

@krokoziabla Please prepend the commit message with "cmake:".

evpobr:
It is correct.

krokoziabla:
@erikd Done

erikd:
Thanks @krokoziabla

sezero:
Addition of cmake_minimum_required(VERSION 3.11) to src/CMakeLists.txt seems bogus.

Other than that, because of TARGET_NAME_IF_EXISTS thing, this is shoving cmake-3.12 requirement to my throat. Isn't there a cheaper way of doing that?

sezero:
> Addition of cmake_minimum_required(VERSION 3.11) to src/CMakeLists.txt seems bogus.
> 
> Other than that, because of TARGET_NAME_IF_EXISTS thing,
> this is shoving cmake-3.12 requirement to my throat. Isn't there
> a cheaper way of doing that?

I mean, the whole tree could be happily built by cmake-3.5 before this
(provided that docs is excluded from build.)

krokoziabla:
@sezero 3.12 released a year ago already. What stops you from upgrading your CMake?

sezero:
> @sezero 3.12 released a year ago already. What stops you from upgrading your CMake?

1. Why should I have to?  There is no _actual_ and _absolute_ need
  for the functionality provided by newer cmake.
2. I'm on CentOS-6.10, still a supported OS, where the default compiler
  is gcc-4.4.7, whereas cmake newer than 3.9 (IIRC) require gcc >= 4.8
  due to increased C++ standart requirements, which I shouldn't have to
  depend on.

krokoziabla:
Well, if you use such an old GCC release probably you don't need the most recent FLAC version. If you do, you still can revert this patch and it will go well for you, since this is just a piece of refactoring which doesn't introduce new features but makes searching for ICONV more robust.

zvezdochiot:
@krokoziabla say:
>you still can revert this patch

No good.

sezero:
>   probably you don't need the most recent FLAC version.

And now you're deciding what I'd need. Great going.

krokoziabla:
I'm just trying to help you :-)

erikd:
CentOS 7 was released in 2014, so CentOS 6 must be even older.  Sometimes things will break on older OS versions. @sezero you (and anyone else) has a couple of options:

* Use a newer OS version.
* Use an alternative build system like the autotool or `Makefile.lite` build systems on your existing OS.
* Stay on top of CMake build system related PRs and speak up before PRs like this get merged.

PR #189, opened 2019-12-26
Fix nullptr dereference in FLAC++

tzahola:
I've encountered a crash due to dereferencing a nullptr when calling `SimpleIterator::get_block()`. It turned out that the C API (`FLAC__metadata_simple_iterator_get_block`) can return null which wasn't propagated through the C++ API. This PR fixes the issue by ensuring that we don't dereference the `FLAC__StreamMetadata` pointer if it's null, in which case the C++ API will return null like the C function. 

I've encountered this issue when trying to open a file that got truncated in the middle of a PICTURE block.


erikd:
Thanks!

PR #194, opened 2020-02-09
fix build under Mac OS X with PowerPC CPU

miniupnp:
revert 44036c9a9b45d03373fe90e9c112852bfc054c51 because auxv.h is only
available under Linux and FreeBSD.

tested on OS X 10.4 with a G4 PowerPC CPU (7400)


miniupnp:
reference to macports bug : https://trac.macports.org/ticket/58947

PR #195, opened 2020-03-04
link lm publicly to FLAC

ericLemanissier:
without this, cmake compilation of shared library fails with these error:
```
[ 92%] Linking C executable ../../../bin/metaflac
/usr/bin/ld: ../../../lib/libreplaygain_analysis.a(replaygain_analysis.c.o): undefined reference to symbol 'log10@@GLIBC_2.2.5'
//lib/x86_64-linux-gnu/libm.so.6: error adding symbols: DSO missing from command line
```
and
```
[100%] Linking C executable ../../../bin/flac
/usr/bin/ld: ../../../lib/libreplaygain_analysis.a(replaygain_analysis.c.o): undefined reference to symbol 'log10@@GLIBC_2.2.5'
//lib/x86_64-linux-gnu/libm.so.6: error adding symbols: DSO missing from command line
```

### DIFF
--- a/src/libFLAC/CMakeLists.txt
+++ b/src/libFLAC/CMakeLists.txt
@@ -102,7 +102,7 @@ endif()
 target_include_directories(FLAC INTERFACE
     "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>"
     "$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>")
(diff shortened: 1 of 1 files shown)


erikd:
@evpobr Does this CMake change make sense?

evpobr:
@krokoziabla is CMake developer here.

According to output, math library is dependency of `libreplaygain_analysis` and should be added to it.

evpobr:
[Confirm the bug](https://github.com/evpobr/flac/runs/489368610).

PR #197, opened 2020-03-09
Integrate Github Actions

evpobr:
Travis CI tests are incredible slow. This PR adds [Github Actions](https://help.github.com/en/actions) workflow pipeline to run tests.

Note macOS GCC tests are removed, it seems like [Apple's GCC is actually Clang](https://www.quora.com/Why-does-the-gcc-command-on-MacOS-execute-clang),

Travis CI is not deleted in this PR just to be sure everything works as expected.

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -1,0 +1,153 @@
+name: GitHub Actions
+
+on: [push]
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        name:
+          [
+            ubuntu-latest-gcc-autotools,
+            ubuntu-latest-clang-autotools,
+            ubuntu-latest-gcc-cmake,
+            ubuntu-latest-clang-cmake,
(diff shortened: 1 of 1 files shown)


Makefile.am
```diff
@@ -38,6 +38,7 @@ SUBDIRS += examples
 endif
 
 EXTRA_DIST = \
+	.github/workflows/action.yml \
```
erikd, line 41: Is that really something that belongs in the distribution source code tarball? Its Github specific.

evpobr: Probably not. Remove this?

erikd: Yes please.

evpobr: @erikd, complete.

erikd:
Thanks!. Are these now enabled?

evpobr:
Yes, the same as Travis CI, see also [Actions](https://github.com/xiph/flac/actions) pane.

PR #201, opened 2020-04-07
cmake: Fix pkg-config install path

evpobr:


erikd:
Thanks @evpobr  and thanks @DeadSix27 for notifying us.

erikd:
I *think* I should have let CI finish before I merged that. That might have avoided the failed Github Actions.

evpobr:
[Configuring protected branches](https://help.github.com/en/github/administering-a-repository/configuring-protected-branches).

But be aware of Travis CI - sometimes it fails just because job time limit reached,

PR #202, opened 2020-04-07
Add CMake and VS Code Git ignore rules

evpobr:


evpobr:
Repository has build directory, we don't want to exclude it.

PR #210, opened 2020-05-09
CMake polishing

evpobr:
* Fix x86 and AMD64 detection
* Improve Doxygen docs generation
* Fix Windows UWP build
* Minor bugfixes


evpobr:
@erikd?

erikd:
Sorry, super busy with my day job. However these should be squashed down to a single commit.

evpobr:
@erikd, ready.

CMakeLists.txt
```diff
@@ -17,6 +17,9 @@ option(BUILD_PROGRAMS "Build and install programs" ON)
 option(BUILD_EXAMPLES "Build and install examples" ON)
 option(BUILD_DOCS "Build and install doxygen documents" ON)
 option(WITH_STACK_PROTECTOR "Enable GNU GCC stack smash protection" ON)
+if(MINGW)
```
NotTsunami, line 20: This should be moved down to L75 if anything. Curious as to why you'd disable on MINGW? It shouldn't be adding the flag if the compiler doesn't support it.

NotTsunami: If it comes down to having to disable it for MingW the option shouldn't be exposed for MingW. There are many cleaner ways to do this.

src/libFLAC/CMakeLists.txt
```diff
@@ -31,7 +46,8 @@ if(NOT WITH_ASM)
     add_definitions(-DFLAC__NO_ASM)
 endif()
 
-if(IA32)
+if(FLAC__CPU_IA32)
+
```
NotTsunami, line 50: Extra blank line

src/libFLAC/CMakeLists.txt
```diff
@@ -10,14 +10,29 @@ check_include_file("sys/param.h" HAVE_SYS_PARAM_H)
 set(CMAKE_REQUIRED_LIBRARIES m)
 check_function_exists(lround HAVE_LROUND)
 
-if(CMAKE_SYSTEM_PROCESSOR MATCHES "[xX]86_64|(AMD|amd)64")
-    if(CMAKE_SIZEOF_VOID_P EQUAL 4)
-        set(IA32 TRUE)
-    endif()
-    add_definitions(-DFLAC__CPU_X86_64 -DFLAC__ALIGN_MALLOC_DATA)
-elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "([xX]|i[346])86")
```
NotTsunami, line 18: imo this regex could be improved:
"[xX]86|i.86"
would catch the currently missing i586 architecture

src/libFLAC/CMakeLists.txt
```diff
@@ -10,14 +10,29 @@ check_include_file("sys/param.h" HAVE_SYS_PARAM_H)
 set(CMAKE_REQUIRED_LIBRARIES m)
 check_function_exists(lround HAVE_LROUND)
 
-if(CMAKE_SYSTEM_PROCESSOR MATCHES "[xX]86_64|(AMD|amd)64")
-    if(CMAKE_SIZEOF_VOID_P EQUAL 4)
```
NotTsunami, line 14: Is the issue CMAKE_SIZEOF_VOID_P not working? I'd rather continue seeking alternatives than introduce C to our CMake files.

CMakeLists.txt
evpobr: More info here: msys2/MINGW-packages#5803

This works for me. I you have clever solution i whould like to see it.

src/libFLAC/CMakeLists.txt
evpobr: `CMAKE_SIZEOF_VOID_P` returns value of system, not target. `CMAKE_SYSTEM_PROCESSOR` is also useless. More info: [How to detect target architecture using CMake?
](https://stackoverflow.com/questions/11944060/how-to-detect-target-architecture-using-cmake), [https://gitlab.kitware.com/cmake/cmake/issues/15170](CMAKE_SYSTEM_PROCESSOR is broken on Windows).

For `CMAKE_SYSTEM_PROCESSOR` is `AMD64` when i build in Visual Studio + Ninja + MSVC ARM64 toolchain.

CMakeLists.txt
NotTsunami: -fstack-protector-strong should cause libssp to get linked, right? Wouldn't it only be an issue with __FORTIFY_SOURCE?

Anyway, the quickest band-aid is changing L76 of the old CMakeLists.txt from:
`if(NOT MSVC)`
to
`if(NOT MSVC AND NOT MINGW)`

src/libFLAC/CMakeLists.txt
NotTsunami: According to the documentation, [CMAKE_SIZEOF_VOID_P](https://cmake.org/cmake/help/latest/variable/CMAKE_SIZEOF_VOID_P.html) has been returning the value of the target machine since CMake 3.2. As far as CMAKE_SYSTEM_PROCESSOR goes, I wouldn't call it useless, we can still work around that bug IMO. In the case that we hit AMD64, we can also check against CMAKE_CL_64 and if it is false, then set IA32. Google does a [similar](https://boringssl.googlesource.com/boringssl/+/refs/heads/master/CMakeLists.txt#472) check in BoringSSL.

evpobr: Updating Vcpkg package manager FLAC port. a've found `arm-windows` and `arm64-windows` builds both failed.

I've tryed all known checks: `CMAKE_SYSTEM_PROCESSOR`, `CMAKE_GENERATOR_PLATFORM` , spent three days dearching answer - no success.

The only working solution is here in this PR. This works allways. Doesn't break anything.

CMakeLists.txt
evpobr: I've decided to remove this hack. No idea how to detect bugged MinGW. At least we have option to disable stack protector.

src/libFLAC/CMakeLists.txt
evpobr: Fixed.

evpobr:
Small fix: use `FLAC__CPU_X86_64` and `FLAC__CPU_IA32` to add `WITH_AVX` option.

evpobr:
Ok, checks are hidden behind `check_cpu_arch_x64` and `check_cpu_arch_x86`. (`cmake/CheckCPUArch.cmake` module`). @NotTsunami , are we happy now?

evpobr:
@erikd, merge it, please.

Issue #208, opened 2020-05-04
clang-format of the tree


neheb:
While running clang-tidy on the tree, I've noticed inconsistencies in the codebase (mixed TAB and space, various code styles used, etc...).

How about running clang-format through the whole codebase in order to have a consistent style?

The main question would be which style is desired.

erikd:
The most import thing we have in the repository other than the code itself is the change history.

I an see the benefit of applying an auto formater at the start of a project, applying an auto formater to an existing code base destroys the change history for close to zero benefit.

If auto formatting could be applied to the additions of diffs, they this may be worth considering.

neheb:
There is integration with git clang-format, sure.

This is another project's approach to the problem: https://github.com/gerbera/gerbera/commit/0f1083e1dd4f390a548c3e4fdc5dc29dc159b4f6

No idea how this could be integrated with travis.

erikd:
Having it as a CI check is fine, but how do people ensure its correct locally before pucking code to github?

Installing the required tools is probably easy to do on Linux but I wonder how easy this would be on Mac, Windows and other platforms.

@evpobr ?

erikd:
Also I am not trying to be an asshat here, but I would like point out that the usefulness of this proposal is pretty minor in comparison to tackling the issues like #164 and #192 which I have not tackled due to lack of free time.

Discussing formatting issues and how to fix them are pretty close to the definition of bike shedding.

neheb:
There's git integration: git clang-format HEAD~1

The real question is what style should be used. https://clang.llvm.org/docs/ClangFormatStyleOptions.html#configurable-format-style-options

erikd:
For this to be considered at all we need:

* Formatting to only be applied to the changes to a file in a commit, not the whole file and not the whole project (for reasoning see https://github.com/xiph/flac/issues/208#issuecomment-623774148 ).
* It must be something that can be checked/validated in CI (if it is not checked it will get screwed up).
* It must be something that any dev can apply locally (for all there major OSes) before they push to GH (waiting for CI to provide a patch is too much friction on the development process).

neheb:
All of that is true.

erikd:
Ok, so it now depends on how easy this is for me to set up. The easier it is, the more likely it is to happen.

ValZapod:
> base destroys the change history for close to zero benefit.

No, it does not. You will just have to do blame before the tidy commit. Github permits thag woth no problems.

PR #213, opened 2020-05-31
[RFC] clang-format

neheb:


ktmf01:
@neheb I guess this PR is no longer applicable? Travis is no longer in use here. Should this PR be closed?

neheb:
the first commit is somewhat valid. The second is completely useless. I'll rebase.

.clang-format
```diff
@@ -0,0 +1,19 @@
+BasedOnStyle: WebKit
+AlignAfterOpenBracket: Align
+AlignEscapedNewlines: DontAlign
+AlignOperands: true
+BraceWrapping:
+  AfterFunction: true
```
ktmf01, line 6:
```suggestion
  AfterFunction: true
  BeforeElse: true

```

.clang-format
```diff
@@ -0,0 +1,19 @@
...
+BreakBeforeBinaryOperators: None
+BreakBeforeBraces: Custom
+BreakBeforeInheritanceComma: true
+BreakConstructorInitializers: BeforeComma
+DerivePointerAlignment: false
```
ktmf01, line 11:
```suggestion
DerivePointerAlignment: false
IndentCaseLabels: true
```

ktmf01 (commented):
Here are two additions.

Still, this doesn't really fit with FLACs coding style. This is apparently a problem with clang-format, see [here](https://stackoverflow.com/questions/67245250/can-i-tell-clang-format-to-not-add-space-around-arithmetic-operators) and [here](https://stackoverflow.com/questions/44259984/how-to-tell-clang-format-to-keep-whitespace-between-binary-operators-as-they-are).

Also, when formatting lpc.c, clang-format IMO makes a lot of fall-through switch-case less easy to read.

neheb:
Added changes. git diff on stream_decoder.c goes down by half. cool.

ktmf01:
Are you planning on improving this even further, or can this be merged? It is still marked 'draft'.

I agree with erikd in #208 though, I don't think it is a good idea to clang-format the whole tree. However, this file is nice to have to use with git-clang-format.

neheb:
Marked as ready.

PR #214, opened 2020-06-02
fix incorrect usage of realloc in bitwriter_grow_

ltx2018:


petterreinholdtsen:
The patch seem to assume that if realloc() return NULL, the original memory is lost, but the manual page for realloc() at least on Linux state that "If realloc() fails, the original block is left untouched; it is not freed or moved."  The C standard states that "If memory for the new object cannot be allocated, the old object is not deallocated and its value is unchanged.".

This make me believe the proposed change is not correct.

ltx2018:
> The patch seem to assume that if realloc() return NULL, the original memory is lost, but the manual page for realloc() at least on Linux state that "If realloc() fails, the original block is left untouched; it is not freed or moved." The C standard states that "If memory for the new object cannot be allocated, the old object is not deallocated and its value is unchanged.".
> 
> This make me believe the proposed change is not correct.

if realloc() return NULL, original memory will be freed in safe_realloc_
image

ltx2018:
similar to this patch,
in FLAC__stream_decoder_set_metadata_respond_application
image

in read_metadata_seektable_
image

ltx2018:
the problem is 'size1 > SIZE_MAX / size2' in safe_realloc_mul_2op_ , if that happens original memory not freed, so maybe free(ptr) should be called before return.  
any idea?

erikd:
@ltx2018 please do not post screen captures, my screen resolution makes them unreadably small. Instead copy and paste and learn  how to use GH markup.

ltx2018:
> @ltx2018 please do not post screen captures, my screen resolution makes them unreadably small. Instead copy and paste and learn how to use GH markup.

emm..updated

ktmf01:
The problem that the first commit tries to fix appeared today during fuzzing, so it seems this fix is still needed. I will take some time to review it soon.

ktmf01:
This is fixed with the merge of #419

PR #216, opened 2020-06-05
cmake: fix c/cxx compile option used with asm code

ledocc:


NotTsunami:
You can further simplify it with something like:
`$<$<NOT:$<COMPILE_LANGUAGE:ASM>>`

PR #223, opened 2020-07-09
Fix compression ratio display for very small files

ktmf01:
Because the compression ratio was calculated before processing
the input of the last frame, it did not include the size of this
last frame. This patch moves the calculation of the compression
ratio after the new input has been processed. Now the compression
ratio on very small files is correctly displayed.


erikd (approved):

erikd:

PR #224, opened 2020-07-09
Fix possible overflows in stream_encoder.c and analyse.c

ktmf01:
Recently I was trying some new approaches to improve FLAC compression, when I stumbled on a possible overflow. The reason this has not come up earlier is because the encoder only hits this point when the estimate of the rice_parameter is very much off.

To trigger this overflow, one has to force rice_parameter to 0 in for example the function evaluate_lpc_subframe in libFLAC/stream_encoder.c. When encoding noisy material, which needs a high rice parameter, it can happen that the return value of that function overflows. When this happens, the wrong order might be picked, and the file blows up to enormous proportions. In my case, about 30 times the size of the original WAV file.

When analysing this problem, I found more or less the same overflow in flac/analyse.c, where in flac__analyse_frame, frame_bytes is multiplied by 8. 

These patches alleviate these problems. Below two links with results

[Graph with sample set 1](http://www.audiograaf.nl/misc_stuff/check-ricebitscalculation-overflow-patch1.pdf)
[Graph with sample set 2](http://www.audiograaf.nl/misc_stuff/check-ricebitscalculation-overflow-patch1.pdf)

### DIFF
--- a/src/flac/analyze.c
+++ b/src/flac/analyze.c
@@ -62,7 +62,7 @@ void flac__analyze_init(analysis_options aopts)
 	}
 }
 
-void flac__analyze_frame(const FLAC__Frame *frame, uint32_t frame_number, FLAC__uint64 frame_offset, uint32_t frame_bytes, analysis_options aopts, FILE *fout)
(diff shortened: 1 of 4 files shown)


erikd (approved):

erikd:

PR #229, opened 2020-10-26
libFLAC/stream_decoder.c: Fix divide by zero error in FLAC__stream_de…

NeelkamalSemwal:
…coder_seek_absolute

In stream_decoder.c, there are three equations of finding the position in seek_to_absolute_sample_ function. The third equation i.e. when FLAC__INTEGER_ONLY_LIBRARY is defined and (upper_bound - lower_bound < 0xffffffff), can cause a divide by zero error if (upper_bound_sample - lower_bound_sample) is less than 2^16. This condition can arise due to an invalid seek table.

The fix mitigates the issue and ensures that the 1st equation and the 3rd equation give the same position under this condition.

### DIFF
--- a/src/libFLAC/stream_decoder.c
+++ b/src/libFLAC/stream_decoder.c
@@ -3119,8 +3119,10 @@ FLAC__bool seek_to_absolute_sample_(FLAC__StreamDecoder *decoder, FLAC__uint64 s
 		/* a little less accurate: */
 		if(upper_bound - lower_bound < 0xffffffff)
 			pos = (FLAC__int64)lower_bound + (FLAC__int64)(((target_sample - lower_bound_sample) * (upper_bound - lower_bound)) / (upper_bound_sample - lower_bound_sample)) - approx_bytes_per_frame;
-		else /* @@@ WATCHOUT, ~2TB limit */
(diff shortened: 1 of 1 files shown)


asanoic (commented):
the branch goes to else of `upper_bound - lower_bound < 0xffffffff` which means `upper_bound - lower_bound >= 0xffffffff`, so `(upper_bound_sample - lower_bound_sample)>>16` should be greater than 0.
and your change is on the code when FLAC__INTEGER_ONLY_LIBRARY  is not defined.

NeelkamalSemwal:
The values of bounds(upper/lower) and samples(upper/lower) come from the seek table. In a correct seek table, there will be a linear relationship between upper_bound and upper_bound_sample i.e. upper_bound_sample will increase with an increasing upper_bound. But, if the seek table is incorrect, and since there are no checks after parsing the seek table for this, hence the condition when (upper_bound - lower_bound) is greater than 0xffffffff but upper_bound_sample and lower_bound_sample don't change accordingly and can get values that can cause (upper_bound_sample - lower_bound_sample)>>16 = 0 and hence a divide by 0 error.

read_metadata_seektable_ function in stream_decoder.c parses the seek table values.

erikd (approved):

erikd (approved):

erikd:

PR #230, opened 2020-10-29
libFLAC/stream_decoder.c: Use current position as bound when seeking

enzo1982:
I ran into an issue today when seeking forward or backward small amounts of samples in FLAC files without seek-tables (or with very widely spaced seek points, e.g. every 10 seconds).

Turns out the decoder always searches the whole file when seeking. That the files at hand contained a lot of silence made it worse, as it kind of defeated the pivot point calculation based on average frame size throughout the file.

This PR greatly speeds up seeking in such cases by using the current position as either the lower or upper bound, depending on where the target position happens to be.


sezero:
Has this patch ever been reviewed, any intentions to apply it?

erikd:

PR #235, opened 2021-03-14
Consolidate a bunch of PRs

erikd:


erikd:
Doxygen failure :cry: .

PR #238, opened 2021-04-25
cpu.h: detect AVX/FMA intrinsics availability on clang

tmkk:
Currently `FLAC__AVX2_SUPPORTED` macro is defined, but actually AVX2 functions are not used when built with clang because flac only uses them when `FLAC__AVX_SUPPORTED` macro is set.


NotTsunami:
@rillian This should be able to landed safely.

rillian:
Thanks for reviewing, @NotTsunami. [Green ci run](https://github.com/xiph/flac/actions/runs/1063946185). Merged in b358381a102a2c1c153ee4cf95dfc04af62faa1a.

PR #241, opened 2021-05-26
Fix version & requires fields in pkg config files

tg-m:
Details:
 - When building with CMake, pkg-config files flac.pc and flac++.pc
   are not updated with Version information. This causes a failure
   when running a configure script in opus-tools
   (https://github.com/xiph/opus-tools).

   Note that OGG_PACKAGE is set to "ogg" exactly as in the configure.ac,
   but it is only set when the building explicitly with libogg.

BTW, I have no idea what libogg is, so this might be an "over-patch". :)


c72578 (approved):
LGTM.
`Version` is currently empty in the **`*.pc`** files. Furthermore, `ogg`should be added to `Requires` in case of `WITH_OGG`. Both has been addressed by this PR.

rillian:
Thanks for reviewing. Can you please submit a follow-up witht he `Requires` change? I don't know cmake well enough to follow what you meant.

PR #242, opened 2021-05-26
Fix building/installing docs with CMake

tg-m:
Details:
 - During the installation of the package with CMake, it stops
   (via fail) when doc files should be copied. It was caused by
   changing current binary to current source CMake directory.

Note that it could also be solved by copying files from the source to the binary dir in the first place, but I didn't have time to inspect that.


rillian:
Thanks!

PR #244, opened 2021-06-22
Fix encoding of contributor name.

maxz:
The name of Håvard Kvålen was wrongly encoded.

I came across this while checking which environment variables `flac` evaluates and decided to fix it.


rillian:
Thanks!

Issue #199, opened 2020-04-04
[regression, bisected] 1.3.3 fails to build on non-VSX ppc64 CPUs


ernsteiswuerfel:
`./configure && make` works fine with 1.3.2.

With 1.3.3 I get:
```
make[4]: Entering directory '/root/build/flac/src/libFLAC'
  CC       bitmath.lo
  CC       bitreader.lo
  CC       bitwriter.lo
  CC       cpu.lo
  CC       crc.lo
  CC       fixed.lo
  CC       fixed_intrin_sse2.lo
  CC       fixed_intrin_ssse3.lo
  CC       float.lo
  CC       format.lo
  CC       lpc.lo
  CC       lpc_intrin_sse.lo
  CC       lpc_intrin_sse2.lo
  CC       lpc_intrin_sse41.lo
  CC       lpc_intrin_avx2.lo
  CC       lpc_intrin_vsx.lo
lpc_intrin_vsx.c: In function 'FLAC__lpc_compute_autocorrelation_intrin_power8_vsx_lag_16':
lpc_intrin_vsx.c:94:7: warning: implicit declaration of function 'vec_vsx_ld'; did you mean 'vec_vslh'? [-Wimplicit-function-declaration]
   94 |  d0 = vec_vsx_ld(0, base);
      |       ^~~~~~~~~~
      |       vec_vslh
lpc_intrin_vsx.c:94:7: warning: nested extern declaration of 'vec_vsx_ld' [-Wnested-externs]
lpc_intrin_vsx.c:94:7: error: incompatible types when assigning to type '__vector float' {aka '__vector(4) float'} from type 'int'
lpc_intrin_vsx.c:95:7: error: incompatible types when assigning to type '__vector float' {aka '__vector(4) float'} from type 'int'
   95 |  d1 = vec_vsx_ld(16, base);
      |       ^~~~~~~~~~
lpc_intrin_vsx.c:96:7: error: incompatible types when assigning to type '__vector float' {aka '__vector(4) float'} from type 'int'
   96 |  d2 = vec_vsx_ld(32, base);
      |       ^~~~~~~~~~
lpc_intrin_vsx.c:97:7: error: incompatible types when assigning to type '__vector float' {aka '__vector(4) float'} from type 'int'
   97 |  d3 = vec_vsx_ld(48, base);
      |       ^~~~~~~~~~
[...]

PR #191, opened 2020-01-05
Add Arm64/Aarch64 support

erikd:

PR #183, opened 2019-12-02
Add Aarch64 support

coreyjjames:
This pull request adds Aarch64 support to the FLAC project.
fixes #156  

What is included in this pull request:

- Edited configure script to check for aarch64 CPU and define FLAC__CPU_AARCH64 if aarch64 is detected.
- Edited configure script to check for arm_neon.h and define FLAC__HAS_NEONINTRIN if arm_neon.h  is detected.
- Add the logic required to choose the aarch64 versions of the lpc_compute_autocorrelation function. When on an aarch64 machine.
- Translated SSE intrinsics from the lpc_intrin_sse.c file into neon intrinsics and put them into a file called lpc_intrin_neon.c. 
- Added the new file lpc_intrin_neon.c to the src/libFLAC/Makefile.am.

Performance Boost to encoding:
I tested the performance with two aarch64 machines. The test I ran was encoding a .wav file to .flac. The size of the wave file was 1.57 gigabytes.
The machine, with a cortex-a57 8 threads, I got a performance increase of 106.57% for the compute autocorrelation function.
A savings of 0m8.281 seconds.
The machine, with a cortex-a53 24 threads, I got a performance increase of 254.6% for the compute autocorrelation function.
A savings of 1m40.166 seconds.

PR #245, opened 2021-06-30
Change autoc[] to double

ktmf01:
This changes the computation of autocorrelation from single precision to double precision. This results in a compression gain of on average 0.3%point or 0.5% with compression setting -4, but on some tracks this can reach 2.2%point or 5% for
setting -4. See the graphs below

http://www.audiograaf.nl/misc_stuff/double-autoc-with-sse2-intrinsics.pdf
http://www.audiograaf.nl/misc_stuff/double-autoc-with-sse2-intrinsics-per-track.pdf

More information and explanation is on the flac-dev mailing list: http://lists.xiph.org/pipermail/flac-dev/2021-June/006470.html


ktmf01:
The second commit, "Add check for vec_doubleh in altivec.h to configure" should also fix issue #199

ktmf01:
This is driving me crazy. I cannot reproduce the failing CI checks. I've ran the full test suite over 1000 times on my Linux x86 system, (800 times with configure, 200 times with CMake -DENABLE_64_BIT_WORDS), but not fails. I've also borrowed a Macbook from a friend of mine, installed XCode 12.4 and ran CMake -DENABLE_64_BIT_WORDS 3 times, but no fails. Is there any way the CI test could be made a little more verbose?

rillian:
The failure is with the macos-latest-clang-cmake-64-bit-words job:

> Operating System
>  - Mac OS X
>  - 10.15.7
>  - 19H1217
>
> Virtual Environment
>  - Environment: macos-10.15
>  - Version: 20210626.1
>  - Included Software: https://github.com/actions/virtual-environments/blob/macOS-10.15/20210626.1/images/macos/macos-10.15-Readme.md
>  - Image Release: https://github.com/actions/virtual-environments/releases/tag/macOS-10.15%2F20210626.1
>
> Build with CMake
>  - cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON -DENABLE_64_BIT_WORDS=ON -DCMAKE_FIND_FRAMEWORK=NEVER
>  - cmake --build .
>  - ctest -V
>
> 8: noise (--channels=1 --bps=8 -6 -e --lax -b 32768 ): encode... ERROR during encode of noise
> 
> The following tests FAILED:
> 	  8 - streams (Failed)

I don't know if there's a way to make the build logs public.

rillian:
Asking github to archive the logs, it appears to be a segfault:
```
### ENCODE noise #######################################################
###    cmd=run_flac --verify --silent --force --force-raw-format --endian=little --sign=signed --sample-rate=44100 --bps=16 --channels=2 -8  --lax -b 65535  --no-padding noise.raw
/Users/runner/work/flac/flac/test/test_streams.sh: line 34: 35087 Segmentation fault: 11  flac${EXE} --no-error-on-compression-fail $*
```

rillian:
I spent some time trying to script invoking lldb on failure.

The good new is, if I break the `|| die` on test failure, it fails on about a dozen different variations of the `--channels=2 -8 --lax -b 65535` encoder options on `noise.raw`. Passing `-DENABLE_64_BIT_WORDS=ON` to cmake doesn't make a differnce.

The bad news is, [it doesn't seem to fail under lldb](https://github.com/xiph/flac/runs/2983927674?check_suite_focus=true), so I couldn't get a backtrace.
```
8: noise (--channels=2 --bps=16 -8  --lax -b 65535 ): encode... FAIL
8: (lldb) target create "flac"
8: Current executable set to 'flac' (x86_64).
8: (lldb) run --verify --silent --force --force-raw-format --endian=little --sign=signed --sample-rate=44100 --bps=16 --channels=2 -8  --lax -b 65535  --no-padding noise.raw
8: Process 36323 exited with status = 0 (0x00000000) 
```
Have you tried running the tests under valgrind? I think you have to set `FLAC__TEST_WITH_VALGRIND=yes` explicitly in the environment with the cmake build. The autotools build has an `--enable-valgrind-testing` switch.

ktmf01:
Many, many thanks for taking so much of your time to have an in-depth look into this. I've just started a run with Valgrind, and I see there is indeed something wrong with the new SSE code. Thanks for the tip, I should have thought of that sooner :)

ktmf01:
I've merged and rewritten the additions from pull request #183 (and thereby also #191). Perhaps @coreyjjames might want to take a look at it?

Also, I've done some testing on Travis CI with an altered test_compression.sh (not included in these commits) to check compression and speed, see the very bottom of the logs here: https://travis-ci.com/github/ktmf01/flac/builds/232024699 Results are: VSX code is at least 2x as fast as non-VSX code, compression is exactly the same. NEON code is 30% faster than non-NEON code, compression is exactly the same.

Also, the problem mentioned in #191 should be fixed.

src/libFLAC/lpc_intrin_neon.c
```diff
@@ -0,0 +1,324 @@
...
+    vst1q_f64(autoc, sum0);
+    vst1q_f64(autoc + 2, sum1);
+    vst1q_f64(autoc + 4, sum2);
+    vst1q_f64(autoc + 6, sum3);
+    vst1q_f64(autoc + 8, sum4);
+    vst1q_f64(autoc + 10, sum5);
+    vst1q_f64(autoc + 12, sum6);
+    vst1q_f64(autoc + 14, sum7);
+}
+#endif /* FLAC__CPU_ARM && FLAC__HAS_ARCH64INTRIN */
```
NotTsunami, line 322:
```suggestion
#endif /* FLAC__CPU_AARCH64 && FLAC__HAS_ARCH64INTRIN */
```

src/libFLAC/cpu.c
```diff
@@ -298,6 +300,7 @@ void FLAC__cpu_info (FLAC__CPUInfo *info)
 	case FLAC__CPUINFO_TYPE_PPC:
 		ppc_cpu_info (info);
 		break;
+	case FLAC__CPUINFO_TYPE_ARM: /* fallthrough */
```
NotTsunami, line 303:
```suggestion
	case FLAC__CPUINFO_TYPE_AARCH64: /* fallthrough */
```

src/libFLAC/cpu.c
```diff
@@ -286,6 +286,8 @@ void FLAC__cpu_info (FLAC__CPUInfo *info)
 	info->type = FLAC__CPUINFO_TYPE_X86_64;
 #elif defined FLAC__CPU_PPC
 	info->type = FLAC__CPUINFO_TYPE_PPC;
+#elif defined FLAC__CPU_ARM
+	info->type = FLAC__CPUINFO_TYPE_ARM;
```
NotTsunami, line 290:
```suggestion
#elif defined FLAC__CPU_AARCH64
	info->type = FLAC__CPUINFO_TYPE_AARCH64;
```

src/libFLAC/include/private/cpu.h
```diff
@@ -154,6 +154,7 @@ typedef enum {
 	FLAC__CPUINFO_TYPE_IA32,
 	FLAC__CPUINFO_TYPE_X86_64,
 	FLAC__CPUINFO_TYPE_PPC,
+	FLAC__CPUINFO_TYPE_ARM,
```
NotTsunami, line 157:
```suggestion
	FLAC__CPUINFO_TYPE_AARCH64,
```

NotTsunami (changes requested):
At some point `FLAC__CPU_ARM` was changed to `FLAC__CPU_AARCH64` in the old PR, and there are still a few remnants of that here. I'll probably throw out another review once I take a more in-depth look.

NotTsunami:
None of those changes are functional changes (in the above review), but it'll keep naming consistent.

ktmf01:
I've just taken a look myself, I think those are indeed all mentions of 'ARM' in the code. However, now that I think of it, wouldn't it be better to completely remove the ARM-related code in cpu.c, as it does not serve any purpose right now? Or should it be kept around in case anyone wants to write code with SVE/SVE2 intrinsics? What do you think @NotTsunami ?

NotTsunami:
> I've just taken a look myself, I think those are indeed all mentions of 'ARM' in the code. However, now that I think of it, wouldn't it be better to completely remove the ARM-related code in cpu.c, as it does not serve any purpose right now? Or should it be kept around in case anyone wants to write code with SVE/SVE2 intrinsics? What do you think @NotTsunami ?

It would be nice to have some ARMv7 compatible NEON intrinsics, but since this switches autoc[] to double-precision, it would be more of a hassle for that. For simplicity's sake, you could probably remove it for now, and if anyone picks up any further intrinsics, they could probably add it back in.

ktmf01:
I've removed the ARM parts of cpu.c. As NEON is optional in ARMv7 and requires OS-specific code to check for support (unlike x86 feature detection), this would need quite a bit of rework anyway.

ktmf01:
Sorry I keep adding commits to this pull request, but I saw quite an opportunity further optimizing the code I already touched. I rewrote the VSX and NEON code to be much cleaner and easier to read, and added a check to the VSX code for very small data_len. This gives a speed increase for preset 8 of about 40%.

I also rewrote the SSE2 code to be much cleaner and easier to read, but sadly the gain is only 8%.

ktmf01:
Earlier it was mentioned that this change results in on average 0.5% compression improvement, with outliers on specific material of up to 5%, but it turns out the gains are much, much larger on files where high-frequency signals are missing, for example band-passed files, 78rpm vinyl rips or high samplerate (>96kHz) material that contains only audible frequencies. As noted in https://hydrogenaud.io/index.php?topic=120158.msg1001482#msg1001482, gains of 10.7%-point (or 19.5%) have been seen.

ValZapod:
Two questions: afaik, 32 bit audio with all bits set can be put into 24 bit flac losslessly. But what about adding actual support for 32-bit flac? Everybody says there are no implementations of 32 bit flac even here in the reference encoder.

Second: I see that in the link ahove (https://hydrogenaud.io/index.php?topic=120158.msg1001482#msg1001482) somebody brute-forces flac for 5 days to produce even smaller samples. Did you think about creating an algorithm that will converge faster or an AI approach with then reverse ingeenering the algorithm?

ktmf01:
@ValZapod The two question you are asking have nothing to do with this pull request. Concerning the 32-bit audio, you could open a new issue, but please provide a clear use case. Concerning the second question, please read the first few posts of that thread carefully. You'll see that there are still people trying new algorithms, but these developments take **a lot** of time. Please keep discussion in this thread about this pull request.

ktmf01:
@antonblanchard I haven't been able to contact you by mail, so I try it over github: Could you perhaps spare some time to check the changes this pull request makes to the PPC code you added a few years back? The performance claims I made concerning the speedups were based on limited data supplied by the CI running on PowerPC, but perhaps you can verify those too?

ktmf01:
I'm going to split up this PR because it does way to much at once. I will split off the ARM64 part and rework the SSE2/PPC parts

vrubleg:
Please mention this PR in the new PRs, so everybody who are following it would know what to follow now. I hope that this cool thing will be merged soon :)

ktmf01:
@vrubleg most is now in #317 and #318. For the ARM intrinsics part, that still needs some more work.

PR #246, opened 2021-07-04
Github actions improvements

rillian:
Two changes, grouped here for testing:
- Upload logs and a artifact if the build fails
  Can see this [working here](https://github.com/xiph/flac/actions/runs/998591104).
- Try to add a windows job

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -149,3 +149,10 @@ jobs:
           cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON ${{ matrix.configure-opts }} -DCMAKE_FIND_FRAMEWORK=NEVER
           cmake --build .
           ctest -V
+
+      - name: Upload logs on failure
+        uses: actions/upload-artifact@v2
+        if: failure()
+        with:
+          name: flac-${{ github.sha }}-${{ github.run_id }}-logs
+          path: ./**/*.log
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -144,17 +144,36 @@ jobs:
           make
           make check
 
-      - name: Build with CMake
+      - name: Prepare CMake build directory
         if: startsWith(matrix.build-system,'cmake')
         env:
           CC: ${{ matrix.cc }}
           CXX: ${{ matrix.cxx }}
-        run: |
-          mkdir cmake-build
-          cd cmake-build
-          cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON ${{ matrix.configure-opts }} -DCMAKE_FIND_FRAMEWORK=NEVER
-          cmake --build .
-          ctest -V
+        run: mkdir cmake-build
+
+      - name: CMake generator
+        if: startsWith(matrix.build-system,'cmake')
+        env:
+          CC: ${{ matrix.cc }}
+          CXX: ${{ matrix.cxx }}
+        working-directory: cmake-build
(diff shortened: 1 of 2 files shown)


rillian:
@xnorpx Would you mind taking a look at this? I tried to add a Windows built job. It fails (missing ogg package) but [reports as success](https://github.com/xiph/flac/pull/246/checks?check_run_id=2985100521)!

rillian:
The windows runners use powershell by default. Perhaps that doesn't react the same way to multi-line commands as the unix shells? If I split the cmake generator, build, and test commands into separate steps it [marks the build as failing](https://github.com/rillian/flac/runs/2987020427?check_suite_focus=true) as I expected.

xnorpx:
@rillian yeah not sure about the multi line, I checked my actions and those are single line and split up between, configure, build and test

rillian:
Ok, thanks. Looks like that will do as a work around. I'll try to come up with something not too ugly.

rillian:
@xnorpx I split the multi-line cmake step, which seemed to address detection the build failure. Any idea why the [windows job doesn't find any tests](https://github.com/xiph/flac/pull/246/checks?check_run_id=3034205343) to run?

xnorpx:
@rillian let me check if I can get some time this weekend and I can try to build it locally.

rillian:
Thanks, that would be great!

xnorpx:
Couple of things it looks like:

- There is a if not unix then early return, so no tests is added.
- The add_test comman uses .sh
- You need to specify config when running ctest, "ctest -C Debug"

image

xnorpx:
Not sure what the best way to do it, I would probably do a if window then add all tests regularly, as line 1. Not so elegant, but you increase test coverage and don't mess with what is currently there.

rillian:
Thanks! That's a reasonable suggestion. The API tests, are executables of necessity, so we can at least run those.

The gitlab windows runners apparently have an `sh` available. It would be nice if we could figure out how to use that for the ci runs. Othwerwise we'll eventually have to re-write the stream tests in C. :/

I ran into the `ctest -C Debug` issue [with the ogg unit tests](https://github.com/xiph/ogg/commit/1708663af915d38508cf2baeb24157c327d278d5) as well. Is this related to `CMAKE_BUILD_TYPE` at all? It doesn't seem to be necessary that it match.

xnorpx:
here is the bug for ctest on windows; https://gitlab.kitware.com/cmake/cmake/-/issues/20283

.github/workflows/action.yml
```diff
@@ -138,14 +144,40 @@ jobs:
...
-          ctest -V
+        run: mkdir cmake-build
+
+      - name: CMake generator
+        if: startsWith(matrix.build-system,'cmake')
+        env:
+          CC: ${{ matrix.cc }}
+          CXX: ${{ matrix.cxx }}
+        working-directory: cmake-build
+        run: cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON ${{ matrix.configure-opts }} -DCMAKE_FIND_FRAMEWORK=NEVER
```
xnorpx, line 160: there is a missmatch here for BUILD_TYPE and the test type, not sure if it is matter and cmake on windows will default to debug. Otherwise LGTM

rillian:
I tried switcing the `-C Release` but the tests failed. It has to be `-C Debug` otherwise `$TARGET_FILE` expands to e.g.
```
D:/a/flac/flac/cmake-build/src/test_libFLAC/Release/test_libFLAC.exe
```
which isn't where the build put the executable, despite `-DCMAKE_BUILD_TYPE=Release`.

xnorpx:
For multiconfigs such as vs you need to specify at build time. "cmake --build . --config Release"

rillian:
Ok, so it's separate from CMAKE_BUILD_TYPE, and Debug is the default. Thanks for clarifying.

Issue #237, opened 2021-04-07
MD5 signature mismatch when total samples > (2^36)


ali1234:
To reproduce:

    head -c 68719476737 /dev/zero | tee >(md5sum > big.md5) | flac --sign=signed --sample-rate=48000 --channels=1 --endian=little --bps=8 -f - -o big.flac

    metaflac --list big.flac

Result:
```
METADATA block #0
  type: 0 (STREAMINFO)
  is last: false
  length: 34
  minimum blocksize: 4096 samples
  maximum blocksize: 4096 samples
  minimum framesize: 10 bytes
  maximum framesize: 15 bytes
  sample_rate: 48000 Hz
  channels: 1
  bits-per-sample: 8
  total samples: 1
  MD5 signature: 42867245cf5773e2ff5c2a25b39b9346
METADATA block #1
  type: 4 (VORBIS_COMMENT)
  is last: false
  length: 40
  vendor string: reference libFLAC 1.3.3 20190804
  comments: 0
METADATA block #2
  type: 1 (PADDING)
  is last: true
  length: 8192
```

Total samples is 1, ie 68719476737 truncated to 36 bits. However, the md5sum is correct. Hex editing the file to set total samples to 0 will allow the file to be decoded correctly.

Expected result: Encoder should detect overflow of total samples and write zero into the metadata.

PR #247, opened 2021-07-09
Fix overflow of streaminfo total samples (github issue 237)

ktmf01:
When writing a FLAC file with over 2^36 samples, the total samples entry in streaminfo will overflow and wrap around. Setting it to 0 (which means unknown number of samples) makes sure the decoder doesn't rely on this number. All tools should be already familiar with handling this situation, as it also occurs when seeking backswards to write the number of samples to the streaminfo block after encoding is not possible.

This should fix issue #237

### DIFF
--- a/src/libFLAC/stream_encoder_framing.c
+++ b/src/libFLAC/stream_encoder_framing.c
@@ -96,9 +96,13 @@ FLAC__bool FLAC__add_metadata_block(const FLAC__StreamMetadata *metadata, FLAC__
 			FLAC__ASSERT(metadata->data.stream_info.bits_per_sample <= (1u << FLAC__STREAM_METADATA_STREAMINFO_BITS_PER_SAMPLE_LEN));
 			if(!FLAC__bitwriter_write_raw_uint32(bw, metadata->data.stream_info.bits_per_sample-1, FLAC__STREAM_METADATA_STREAMINFO_BITS_PER_SAMPLE_LEN))
 				return false;
-			FLAC__ASSERT(metadata->data.stream_info.total_samples < (FLAC__U64L(1) << FLAC__STREAM_METADATA_STREAMINFO_TOTAL_SAMPLES_LEN));
-			if(!FLAC__bitwriter_write_raw_uint64(bw, metadata->data.stream_info.total_samples, FLAC__STREAM_METADATA_STREAMINFO_TOTAL_SAMPLES_LEN))
-				return false;
+			if(metadata->data.stream_info.total_samples > (FLAC__U64L(1) << FLAC__STREAM_METADATA_STREAMINFO_TOTAL_SAMPLES_LEN)){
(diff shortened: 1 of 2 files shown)


src/libFLAC/stream_encoder_framing.c
```diff
@@ -96,9 +96,13 @@ FLAC__bool FLAC__add_metadata_block(const FLAC__StreamMetadata *metadata, FLAC__
 			FLAC__ASSERT(metadata->data.stream_info.bits_per_sample <= (1u << FLAC__STREAM_METADATA_STREAMINFO_BITS_PER_SAMPLE_LEN));
 			if(!FLAC__bitwriter_write_raw_uint32(bw, metadata->data.stream_info.bits_per_sample-1, FLAC__STREAM_METADATA_STREAMINFO_BITS_PER_SAMPLE_LEN))
 				return false;
-			FLAC__ASSERT(metadata->data.stream_info.total_samples < (FLAC__U64L(1) << FLAC__STREAM_METADATA_STREAMINFO_TOTAL_SAMPLES_LEN));
-			if(!FLAC__bitwriter_write_raw_uint64(bw, metadata->data.stream_info.total_samples, FLAC__STREAM_METADATA_STREAMINFO_TOTAL_SAMPLES_LEN))
-				return false;
+			if(metadata->data.stream_info.total_samples > (FLAC__U64L(1) << FLAC__STREAM_METADATA_STREAMINFO_TOTAL_SAMPLES_LEN)){
```
rillian, line 99: Shouldn't this be `>=' ?

rillian (changes requested):
Thanks for looking at this. I suppose it's hard to write a test for this because of the encode length?

src/libFLAC/stream_encoder_framing.c
ktmf01: From an overflow point-of-view, yes, this should be >=, and I didn't notice that. Now I think of it however, functionally, it should make no difference, because `metadata->data.stream_info.total_samples = (FLAC__U64L(1) << FLAC__STREAM_METADATA_STREAMINFO_TOTAL_SAMPLES_LEN)` overflows to 0, which is the value that we are setting anyway.

I could fix this and force-push a different commit. However, I could also leave it as-is, and check for it in a test (as you suggested). If something changes in the overflow behaviour, which is something undesirable by itself, this could be picked up by the test.

Please tell me which behaviour would be preferred, I'm not really sure.

Concerning writing a test, I will add a commit to this branch today to add this to test_flac.sh. Running the test takes quite a while and uses 54MiB of disk space, but it is doable. I will place it under FLAC__TEST_LEVEL=2 to not stress the CI too much.

ktmf01:
```suggestion
			if(metadata->data.stream_info.total_samples >= (FLAC__U64L(1) << FLAC__STREAM_METADATA_STREAMINFO_TOTAL_SAMPLES_LEN)){
```

ValZapod:
Please merge this already.

PR #228, opened 2020-09-16
build on GCC/MinGW-w64 using cmake

asanoic:
because some MinGW-w64 build may not have stack-protector or FORTIFY enabled.

using `cmake [PATH] -DWITH_OGG=OFF` build successfully after this change.

PR #248, opened 2021-07-12
CMake: Make FORTIFY_SOURCE optional

NotTsunami:
* Also add in checking for libssp for MinGW


NotTsunami:
@sezero @asanoic What do you guys think about this as an alternative to #228? This should catch the case 228 is trying to solve and allow FORTIFY_SOURCE to be optional, similar to Opus.

There are many ways to go about this, I chose the most forward-facing way (It's not the most elegant), but I'm open to other solutions/alternatives, hence why I opened this as a draft PR.

sezero:
I never like fortify options being shoved down my throat, so this should be better

NotTsunami:
> I never like fortify options being shoved down my throat, so this should be better

With it enabled by default (when supported), it's actually [now](https://github.com/xiph/flac/runs/3049290387#step:8:18) being used in the GitHub Actions cmake builds when [previously](https://github.com/xiph/flac/runs/3041644240#step:8:18) it wasn't. In general though, this should also fix the edge case for MinGW toolchains that aren't built with libssp (caused by #167). I'm open to having it disabled by default as well, but I would like to see it enabled by default.

I'll try to spin up a new image with different mingw toolchains with and without libssp support in a little bit to make sure the patch works as expected. I'll shift it out of draft status once I (or anyone else) confirm it actually fixes that build failure and has no regressions.

ktmf01:
I just tested this, cmake fails here with `Unknown CMake command "check_library_exists"`. Adding `include(CheckLibraryExists)` to CMakeLists.txt fixes that.

However, building still fails. I think libssp has to be explicitly added to the linker flags?

NotTsunami:
> I just tested this, cmake fails here with `Unknown CMake command "check_library_exists"`. Adding `include(CheckLibraryExists)` to CMakeLists.txt fixes that.
> 
> However, building still fails. I think libssp has to be explicitly added to the linker flags? Because, despite passing the test HAVE_LIBSSP, when linking `flac` I still get `undefined reference to '__stack_chk_fail'`
> 
> edit: something like this?
> 
> ```
>   if(NOT HAVE_LIBSSP)
>     message(WARNING "Could not find libssp in MinGW, stack protection and/or FORTIFY_SOURCE are unavailable")
>   else()
> 	link_libraries(ssp)
>   endif()
> ```
> 
> With this, I still get some errors, but they do not seem to be caused by anything related to stack protection or FORTIFY_SOURCE. Still, this is as far as I've ever gotten with CMake on MinGW

Yes, those two lines were missing. If you'd like, you can continue work on the PR. I don't have a lot of time right now, so it'll be a while before I would be able to work on it again

PR #250, opened 2021-07-14
action.yml: Use config Debug in all cmake steps

c72578:
Use `-DCMAKE_BUILD_TYPE=Debug` and specify `--config Debug` explicitly for
the build.

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -157,7 +157,7 @@ jobs:
           CC: ${{ matrix.cc }}
           CXX: ${{ matrix.cxx }}
         working-directory: cmake-build
-        run: cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON ${{ matrix.configure-opts }} -DCMAKE_FIND_FRAMEWORK=NEVER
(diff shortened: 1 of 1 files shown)


rillian:
Thanks for making this consistent. I wasn't sure how to do that.

It looks like this sets `NDEBUG` and therefore turns off assertions. I'd like to have test coverage with those on. What about making it consistently build in debug mode for CI? Or add another matrix dimension and build both Debug and Release, although we already have a long test latency.

c72578:
Thanks for the feedback. I will update the PR and switch to `Debug` in all cmake steps.

c72578:
OK, the PR has been updated to use `Debug`.

c72578:
Info concerning the failed CI build (`macos-latest-clang-cmake-64-bit-words`) [[1]](https://github.com/xiph/flac/pull/250/checks?check_run_id=3194504262):

From log file `build (macos-latest-clang-cmake-64-bit-words)/9_CMake test.txt`:
```console
2021-07-29T17:57:06.1577150Z The following tests FAILED:
2021-07-29T17:57:06.1580180Z 	  8 - streams (Timeout)
```
[1] https://github.com/xiph/flac/pull/250/checks?check_run_id=3194504262

ktmf01:
In #267 I actually set the config to release instead of debug. The problem I had with debug was that the output is so tremendously verbose, the CI fails because the logs grow to several GiB in size.

Can this PR be closed, or am I missing something?

c72578:
Yes, this PR can be closed.

PR #252, opened 2021-08-13
Remove special treatment of partition_order == 0

ktmf01:
in set_partitioned_rice, there was a special treatment in case partition_order == 0. In this case, a rice parameter was passed as an argument. However, in case of an LPC block, this rice parameter was calculated from the LPC error, which is based on the samples after windowing. Quite often, this resulted in a subframe with two partition with the same rice parameter. Also, in case of partial_tukey and punchout_tukey windows, the suggested rice parameter could be quite sub-optimal, because the calculated error would be in no sane relation to the actual error, as these two windows 'mute' quite a bit of the signal. This can even go so far as to cause problems such as described in pull request #251 

This patch removes the special treatment of partition_order == 0 and the passing of a rice_parameter as an argument. This fixes the root cause of the issue described in #251 and https://hydrogenaud.io/index.php?topic=121349, whereas PR #251 fixes an overflow caused by this issue.

Testing reveals there is no noteworthy compression gain or measurable slowdown is associated with this patch:
[rice-without-windowing.pdf](https://github.com/xiph/flac/files/6981520/rice-without-windowing.pdf)

### DIFF
--- a/src/libFLAC/stream_encoder.c
+++ b/src/libFLAC/stream_encoder.c
@@ -182,7 +182,6 @@ static uint32_t evaluate_fixed_subframe_(
 	uint32_t blocksize,
 	uint32_t subframe_bps,
 	uint32_t order,
-	uint32_t rice_parameter,
 	uint32_t rice_parameter_limit,
 	uint32_t min_partition_order,
 	uint32_t max_partition_order,
@@ -204,7 +203,6 @@ static uint32_t evaluate_lpc_subframe_(
 	uint32_t subframe_bps,
 	uint32_t order,
 	uint32_t qlp_coeff_precision,
-	uint32_t rice_parameter,
 	uint32_t rice_parameter_limit,
 	uint32_t min_partition_order,
 	uint32_t max_partition_order,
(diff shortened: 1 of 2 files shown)


src/libFLAC/stream_encoder.c
```diff
@@ -4166,46 +4135,99 @@ FLAC__bool set_partitioned_rice_(
...
 #endif
 
-	FLAC__ASSERT(suggested_rice_parameter < FLAC__ENTROPY_CODING_METHOD_PARTITIONED_RICE2_ESCAPE_PARAMETER);
 	FLAC__ASSERT(rice_parameter_limit <= FLAC__ENTROPY_CODING_METHOD_PARTITIONED_RICE2_ESCAPE_PARAMETER);
 
 	FLAC__format_entropy_coding_method_partitioned_rice_contents_ensure_size(partitioned_rice_contents, flac_max(6u, partition_order));
 	parameters = partitioned_rice_contents->parameters;
 	raw_bits = partitioned_rice_contents->raw_bits;
 
-	if(partition_order == 0) {
```
ktmf01, line 4182: This part of the diff is a little hard to follow. In essence, the if(partition_order == 0) block is removed, and the else block that followed it has had its indentation removed

PR #255, opened 2021-09-24
Add SSSE3 intrinsic routine for fixed_compute_best_predictor

ktmf01:
A new SSSE3 intrinsic routine is added to compute the best fixed predictor. The existing routine is kept for IA32 systems, the new routine is only for x86_64 systems as these have 8 extra vector registers, which are needed for the new routine.

According to my testing, the new routine is about 3 times as fast as the existing routine. This means an overall speedup for preset -2 of about 30% and for preset -0 of about 15%

@lvqcl As you wrote quite a few routines with intrinsics, would you mind taking a look at this?

### DIFF
--- a/src/libFLAC/fixed_intrin_sse2.c
+++ b/src/libFLAC/fixed_intrin_sse2.c
@@ -192,69 +192,122 @@ FLAC__SSE_TARGET("sse2")
 uint32_t FLAC__fixed_compute_best_predictor_wide_intrin_sse2(const FLAC__int32 data[], uint32_t data_len, float residual_bits_per_sample[FLAC__MAX_FIXED_ORDER + 1])
 {
 	FLAC__uint64 total_error_0, total_error_1, total_error_2, total_error_3, total_error_4;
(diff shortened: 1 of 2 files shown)


ktmf01:
[Here are the results](https://github.com/xiph/flac/files/7647220/very.large.set.of.tracks-fixedsse2.pdf) of comparing the current SSSE3 intrinsics with the proposed SSE2 and SSSE3 intrinsics. From top right to bottom left are preset -0, -1, -2, -3, -4 and -5. Tested on a Kaby Lake-R processor, which is reasonably recent.

For 64-bit, using SSSE3 gives almost no gain, for 32-bit there is still a small gain. Perhaps the SSSE3 code should be removed because of the small gain, in an effort to keep the code base reasonably sized.

ktmf01:
For the _wide functions, gains weren't as large, but still present. [Results on a 64-bit machine](https://github.com/xiph/flac/files/7751557/very.large.set.of.tracks.24bit.pdf)

PR #259, opened 2021-10-14
libFlac: Exit at EOS in verify mode

NeelkamalSemwal:
When verify mode is enabled, once decoder flags end of stream, encode processing is considered complete


rillian:
Rebased and merged upstream as e1575e4a7c5157cbf4e4a16dbd39b74f7174c7be. Thanks again for submitting the fix.

Issue #254, opened 2021-09-22
Ogg chaining support?


jprjr:
One feature of Ogg is so-called "chaining" - essentially concatenating multiple Ogg files together, and the player program attempts to play them as a continuous stream.

It seems like libFLAC does not support it - if I use `flac --ogg` to encapsulate the FLAC streams in Ogg files, then concatenate them together, most programs and tools just ignore everything after the first file. Example, assuming I have a folder of WAV files:

```bash
flag --ogg *.wav
cat *.oga > chained.oga
flac -d chained.oga
# the result "chained.wav" only contain's the first WAV file's audio
```

Most programs seem to support Vorbis chaining and Opus chaining, Ogg FLAC chaining would be a great addition. Are there any plans to support it?

ziplantil:
I second this. Fundamentally I think it's just about checking the OGG stream serial numbers in `FLAC__ogg_decoder_aspect_read_callback_wrapper` in [ogg_decoder_aspect.c](https://github.com/xiph/flac/blob/master/src/libFLAC/ogg_decoder_aspect.c) and handling that situation appropriately. It seems that as things are right now, libFLAC enters this mode where it doesn't report an error and will happily keep reading the stream, but will never call the write (nor really any other) callbacks, which is hardly an ideal behavior.

ktmf01:
I've never encountered a chained Ogg FLAC file, so could you perhaps tell something on where they're used?

Do the chains in these chained Ogg FLAC files usually have the same parameters such as bit-depth, samplerate and channels? For example, in TV streams, codecs sometimes change from stereo to multichannel from program to program. If this is something chained Ogg FLAC files are used for (fusing FLAC files with a different number of channels for example), extra care should be taken to test libFLAC with this.

ziplantil:
The use I have tested my PR with in practice has been to play back an Ogg FLAC Icecast stream i.e. online radio. In this circumstance, the PCM audio parameters have stayed constant as the radio decoder always decodes into the same PCM format (with constant bit depth, sample rate and channel count) before encoding it as Ogg FLAC. However, it is conceivable that these parameters could change from one Ogg stream to another.

The use for online streams has also precluded any need to support seeking which could indeed pose an issue if seeking over STREAMINFO blocks.

jprjr:
[...]

PR #261, opened 2021-10-30
Ogg chaining support

ziplantil:
Related: #254

This commit, possibly set of commits in the future, aims to add support for Ogg chaining (concatenation) and handling chained streams when decoding Ogg FLAC streams. Before these changes, attempts to decode chained streams could result in one of two things:
* If the Ogg stream serial numbers are the same for the previous and next stream, the decoder would think it went out of sync
* If they are different, the decoder will simply ignore any pages belonging to the latter; on streams with no definite end, such as online streams, this would cause the decoder to read forever

Two new API calls are added (due to this I envision having to update the version number): `FLAC__stream_decoder_set_ogg_allow_chaining` and `FLAC__stream_decoder_get_ogg_allow_chaining` (and the C++ equivalents using the same naming convention as existing calls). The new feature is only enabled if the user explicitly requests it; this preserves backwards compatibility. Specifically, a user must do `FLAC__stream_decoder_set_ogg_allow_chaining(decoder, true)` to enable this functionality.

When enabled, the Ogg aspect decoder will try to check if the end-of-stream flag is set for a page. If this is the case, a new Ogg flag is set that tells the FLAC decoder to reset itself. It will then proceed to do so; the reset is only partial and will simply cause the FLAC decoder to assume that we begun a new decode process. This will cause metadata, including STREAMINFO, to be re-read. 

The user could detect stream changes by listening to STREAMINFO metadata updates.

The intended use case of supporting Ogg chaining is, among others, to support playback of Ogg FLAC online radio streams that use chaining similar to how Ogg Vorbis streams would be implemented.

These changes do not have tests right now, since I am not sure what the best approach would be to test a feature like this. Likewise, the new feature exists only in libFLAC(++) for now; the `flac` program currently does not use this functionality.


ziplantil:
Draftifying for now due to potential issues such as seeking, as discussed in #254

ktmf01:
Perhaps we can take a look at libopusfile for inspiration about seeking? Seeing these lines of code, it seems they've already figured it out: https://github.com/xiph/opusfile/blob/b23e611ffbed22886dfc1640abfcfd5bc6677b72/src/opusfile.c#L31-L59

ktmf01:
I've been pondering this issue for a while. I've come up with a few decisions that need to be made

- What to do with metadata. After all, that is one reason chained ogg is used. If we say the decoder emits metadata for each link, the decoder must be aware of this possibility, and the 'allow chaining' functions are really necessary. At first I thought it would be better for libFLAC to support this transparently, but this will break users as they do not expect metadata after the first audio frame.
- It might also be possible to decode all audio for all users, but only emit new metadata when enabled by the user. However, this might be weird as all audio will be played but part of the metadata is hidden.
- Should libFLAC emit metadata after a seek crossing a chain link? That way the user of libFLAC can update displayed metadata. Or should the user be expected to process the metadata of all links before starting playback? That could result in large memory usage, especially when album art is involved.

philippe44:
I gave that another try here https://github.com/xiph/flac/pull/667. I think that this PR does not work for the reason I mention at the beginning of my PR: there is no direct control between pulling the last page from the Ogg stream and when that page will reach the Flac decoder. I might be wrong, but otherwise, although it does not crash, you lose Ogg synchronization and miss some last flac frames. I've tested my PR with hours of ogf webradio. 

For the seeking question, it candidly seems to me like a lot of work for now with a dubious use case. Obviously the decision is to the maintainers, but I'd rather have a first step with a disclaimer that seeking is not supported in chained files, knowing that the most important use case if live streams where seeking does not make sense.

ktmf01:
Implemented by #735

PR #263, opened 2021-11-30
cmake: Fix building tests on Windows

CookiePLMonster:
This PR fixes a regression introduced in 10797d58962dc13df5d4772322790754165a3c93. When building tests with either `BUILD_CXXLIBS` and/or `BUILD_PROGRAMS` disabled, a Windows-only branch in CMake script attempted to add tests for these components regardless, and failed.


ktmf01:
I accidentally duplicated this in #267

PR #265, opened 2021-12-21
Fix Makefile.lite: add mkdir for missing dirs and fPIC for aarch64

ktmf01:
Commit f771c64 broke Makefile.lite by removing the only files in the objs directory trees, so git 'forgot' about them. This commit adds a mkdir command to regenerate them on using Makefile.lite, and adds -fPIC to aarch64 so it builds.


ktmf01:
It is probably time Makefile.lite is removed. It isn't part of CI, and having four different build systems is quite high-maintenance

PR #266, opened 2021-12-23
Fix 4 decoder bugs found by oss-fuzz

ktmf01:
Of these 3 commits, the first and third commit fix 2 bugs found by oss-fuzz, and the second is a fix similar to the first. The oss-fuzz  bugs are referenced in the commit messages.


ktmf01:
Commit 1fd178e also fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=27391

rillian (approved):

rillian:
Rebased and merged as of 79e462fb61c860d4bf69ad00d3c9561fe427d2f6. Thanks for fixing!

PR #267, opened 2021-12-31
[CMake] Fixes for building with MinGW and MSVC

ktmf01:
These 4 commits fix various things with building with CMake for MinGW and MSVC. For MinGW, two commits enable building with a stack protector. This subsumes #248 and obsoletes #228. For MSVC, the commits add compiler options and make the IDE output prettier, with naming consistent with the directory structure, and a folder structure in the IDE for all components. Also, a libogg directory tree compiled with MSVC is now also searched for at the same directory level as flac. So, when building FLAC with MSVC, first building libogg with CMake configuration and then building FLAC (with libogg and flac directories being subdirectories of the same directory) should work without manual intervention.

[Performance testing was done for MSVC](https://github.com/xiph/flac/files/7796237/very.large.set.of.tracks.-.MSVC.pdf), comparing the current CMake solution with the current sln files and the proposed CMake building process.

These changes made encoding a bit faster and decoding a tiny bit slower, comparing MSVC 2022 builds with current CMake files and with proposed CMake files. With the proposed changes, building with CMake creates binaries as fast as with the current sln files. For comparison, MinGW still creates binaries that are a lot faster.


ktmf01:
Duplicate of #248 and duplicate of #228

CMakeLists.txt
```diff
@@ -92,7 +115,11 @@ add_compile_options(
     $<$<AND:$<COMPILE_LANGUAGE:CXX>,$<BOOL:${HAVE_WEFFCXX_FLAG}>>:-Weffc++>
     $<$<AND:$<COMPILE_LANGUAGE:C>,$<BOOL:${HAVE_DECL_AFTER_STMT_FLAG}>>:-Wdeclaration-after-statement>)
 
-if(HAVE_STACK_PROTECTOR_FLAG)
+if(HAVE_LIBSSP)
```
ktmf01, line 118:
```suggestion
if(HAVE_LIBSSP) # Implies WITH_FORTIFY_SOURCE
```

PR #269, opened 2022-01-04
Fix windows_unicode_filenames kludge

ktmf01:
Commit b917d456 cleaned up MinGW DLLs, which had lots of exported functions that don't belong there. However, the functions in windows_unicode_filenames where kept, the fix marked as a kludge, because flac, metaflac and lots of testing relied on them.

windows_unicode_filenames.c is moved from src/libFLAC/ to src/share like the .h file with the same name, which was already in
include/share. flac and metaflac binaries now use these files directly instead of through libFLAC. While this might grow binary size (it actually shrinks mine), it provides a much cleaner DLL interface.

### DIFF
--- a/include/share/windows_unicode_filenames.h
+++ b/include/share/windows_unicode_filenames.h
@@ -39,24 +39,21 @@
 #include <sys/utime.h>
 #include "FLAC/ordinals.h"
 
-/***** FIXME: KLUDGE: export these syms for flac.exe, metaflac.exe, etc. *****/
-#include "FLAC/export.h"
-
 #ifdef __cplusplus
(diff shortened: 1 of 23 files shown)


ktmf01:
TODO: fix vcproj and vcxproj files

sezero:
Make sure that your solution satisfies the scenarios noted in #122:
https://github.com/xiph/flac/pull/122#issuecomment-518209217

ktmf01:
Thanks @sezero, I made the same mistake here as you did in #122. I did test this, but I was clearly not thorough enough. I will close this for now, I see now that this needs a lot more work.

PR #271, opened 2022-01-05
Series of patches and fixes for autotools building

ktmf01:
Autoconf 2.70 and newer versions trigger warnings about obsolete macro's and revealed a problem with variable-length arrays being used in C++ code (which is apparently not according to spec). Also, two fixed specific to MinGW are included. See commit descriptions for details

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -531,7 +531,7 @@ AH_VERBATIM([FLAC_API_EXPORTS],
 #endif
 #endif])
 
-if test x$enable_shared != "xyes" ; then
+if test x$enable_static = "xyes" ; then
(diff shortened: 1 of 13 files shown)


configure.ac
```diff
@@ -529,6 +529,8 @@ AH_VERBATIM([FLAC_API_EXPORTS],
 #else
 # define FLAC_API_EXPORTS
 #endif
+#elif defined(_WIN32)
+# define FLAC__NO_DLL
```
ktmf01, line 539: Note to self: I vaguely remember this didn't work properly in some configuration because FLAC__NO_DLL was already set. Need to check this.

PR #272, opened 2022-01-08
Simplify and merge win_utf8 files and remove from DLL interface

ktmf01:
libFLAC DLLs were exposing windows_unicode_filename.h functions because flac and metaflac needed to set  flac_internal_set_utf8_filenames. Files windows_unicode_filename.{c/h} and win_utf8_io.[c/h] are merged, and all non-utf8 parts are removed.

With this commit, the libFLAC DLL interface is the same as the libFLAC interface of shared libraries on other platforms.

The effect of this commit on the ability of the flac command line utility to parse non-ASCII filenames has been tested with MSVC 2022 with its own sln/vcxproj-files based build system and with CMake, as well as with MinGW with the addition of patches from either #267 for CMake and #271 for Autotools. All of the builds mentioned before were with shared libs (DLLs).


ktmf01:
@sezero I think this is a much better patch than #269, I've tested it with a few non-ASCII scripts and emoji's with MSVC and MinGW. Thanks for helping me out.

sezero:
Well, I'm all OK with removing those abominations from exports, but it's
surely not my call. Do hope that @lvqcl (and/or @erikd) doesn't find some
breakage with it.

lvqcl:
>
> On Windows, filename must be a UTF-8 encoded filename


IIUC that's an API change.

If some old Windows program calls `FLAC__stream_encoder_init_file()` ...

`FLAC_API FLAC__StreamEncoderInitStatus FLAC__stream_encoder_init_file(FLAC__StreamEncoder *encoder, const char *filename, FLAC__StreamEncoderProgressCallback progress_callback, void *client_data)`

... this old program expects `filename` to be an ANSI-encoded string, not UTF-8 encoded.


Maybe it's OK to do it though? Non-Unicode programs are mostly dead, so any decent Windows program probably uses `FLAC__stream_decoder_init_stream()` instead of `FLAC__stream_encoder_init_file()` .

ktmf01:
> IIUC that's an API change.

Yes, definitely. As #124 was already an API change requiring a library version bump, seems sane to combine the two in one release. I'll add a commit to this PR changing the version numbers.

> ... this old program expects `filename` to be an ANSI-encoded string, not UTF-8 encoded.
> 
> Maybe it's OK to do it though? Non-Unicode programs are mostly dead, so any decent Windows program probably uses `FLAC__stream_decoder_init_stream()` instead of `FLAC__stream_encoder_init_file()`

Fortunately most Windows programs pack their own libFLAC DLL or link statically, AFAIK. Perhaps some users will be disappointed that plugging in a newer libFLAC DLL won't work correctly. Perhaps I should check with some popular programs.

At least cross-platform unicode-aware programs can now use this function properly again.

vrubleg:
Windows 10 supports UTF-8 as a process code page. Details:
https://docs.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page
Also, C runtime supports UTF-8 now:
https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/setlocale-wsetlocale?view=msvc-160#utf-8-support
Just informing. Maybe it could be utilized.

ktmf01:
This really needs an universal solution that also works with older versions of Windows, so moving away from using code pages seems like a better idea.

ktmf01:
Just force-pushed to rebase these changes on top of current master, which has UTF-8 checks added to the test suite. I want to do some more testing with this and merge soon.

PR #277, opened 2022-01-26
Do not treat frame with CRC16 mismatch as OK

### LINKED ISSUE #274: Bad frame data is used in processing
https://github.com/xiph/flac/blob/b358381a102a2c1c153ee4cf95dfc04af62faa1a/src/libFLAC/stream_decoder.c#L2135-L2146

These lines seem to be executed even when the frame CRC is invalid. Seeing the comment on line 2137 ('proper frame'), that doesn't seem to be intentional.

### LINKED ISSUE #276: seek errors
I have problems with a file in Ardour ( https://tracker.ardour.org/view.php?id=8863 ) and think I tracked it down to reproduce in xiph flac on git master, using a small hack to trigger a special position in the presumably correctly written test program:
```diff
--- a/src/test_seeking/main.c
+++ b/src/test_seeking/main.c
@@ -328,6 +328,10 @@ static FLAC__bool seek_barrage(FLAC__bool is_ogg, const char *filename, FLAC__of
        for (i = 0; !stop_signal_ && (count == 0 || i < count); i++) {
                FLAC__uint64 pos;
 
+               if (1) {
+                       pos = count;
+                       i = count;
+               } else
                /* for the first 10, seek to the first 10 samples */
                if (n >= 10 && i < 10) {
                        pos = i;
```

Testing a special case with a 160 MB flac file generated by Soundcraft Ui24R, currently available on 
https://wetransfer.com/downloads/1eadb1aa46ba40b0e301f745ee4b782120220120202728/5db0efc9546315e640f764b310cca39c20220120202728/dea7f9 :
```
$ src/test_seeking/test_seeking 12\ ORGAN\ R.flac 90059776

+++ seek test: FLAC__StreamDecoder (native FLAC, read_mode=0)

CPU info (x86-64):
  CMOV ....... Y
  MMX ........ Y
  SSE ........ Y
  SSE2 ....... Y
  SSE3 ....... Y
  SSSE3 ...... Y
  SSE41 ...... Y
  SSE42 ...... Y
  AVX ........ Y
  FMA ........ n
  AVX2 ....... n
  AVX OS sup . Y
stream decoder state is FLAC__STREAM_DECODER_END_OF_STREAM
file's total_samples is 123578720
Begin seek barrage, count=90059776
#90059776:seek(90059776)... FAILED, FLAC__stream_decoder_seek_absolute() FAILED, state = 6 (FLAC__STREAM_DECODER_SEEK_ERROR)
```
but for positions a bit before or after, it works fine:
```
$ src/test_seeking/test_seeking 12\ ORGAN\ R.flac 90059770

+++ seek test: FLAC__StreamDecoder (native FLAC, read_mode=0)

CPU info (x86-64):
  CMOV ....... Y
  MMX ........ Y
  SSE ........ Y
  SSE2 ....... Y
  SSE3 ....... Y
  SSSE3 ...... Y
  SSE41 ...... Y
  SSE42 ...... Y
  AVX ........ Y
  FMA ........ n
  AVX2 ....... n
  AVX OS sup . Y
stream decoder state is FLAC__STREAM_DECODER_END_OF_STREAM
file's total_samples is 123578720
Begin seek barrage, count=90059770
#90059770:seek(90059770)... frame@90059770(134)... OK

PASSED!
...
```
and same for
```
$ src/test_seeking/test_seeking 12\ ORGAN\ R.flac 90059779
```

I guess this file happens to hit some bad corner case in the complex bisection code in `seek_to_absolute_sample_`?

This is on Fedora 35.

ktmf01:
This commit fixes what seems to be an (very) old mistake in the code where a frame that doesn't pass CRC check in the frame footer is used anyway. This fixes #276 and fixes #274

### DIFF
--- a/src/libFLAC/stream_decoder.c
+++ b/src/libFLAC/stream_decoder.c
@@ -2123,13 +2123,10 @@ FLAC__bool read_frame_(FLAC__StreamDecoder *decoder, FLAC__bool *got_a_frame, FL
 		}
 	}
 	else {
-		/* Bad frame, emit error and zero the output signal */
+		/* Frame CRC doesn't match, emit error and return */
 		send_error_to_client_(decoder, FLAC__STREAM_DECODER_ERROR_STATUS_FRAME_CRC_MISMATCH);
-		if(do_full_decode) {
-			for(channel = 0; channel < decoder->private_->frame.header.channels; channel++) {
-				memset(decoder->private_->output[channel], 0, sizeof(FLAC__int32) * decoder->private_->frame.header.blocksize);
-			}
-		}
+		decoder->protected_->state = FLAC__STREAM_DECODER_SEARCH_FOR_FRAME_SYNC;
+		return true;
 	}
 
 	*got_a_frame = true;


src/libFLAC/stream_decoder.c
```diff
@@ -2123,13 +2123,15 @@ FLAC__bool read_frame_(FLAC__StreamDecoder *decoder, FLAC__bool *got_a_frame, FL
 		}
 	}
 	else {
-		/* Bad frame, emit error and zero the output signal */
+		/* Bad frame, emit error, zero the output signal, return */
 		send_error_to_client_(decoder, FLAC__STREAM_DECODER_ERROR_STATUS_FRAME_CRC_MISMATCH);
```
kiilerix, line 2127: Can you clarify (perhaps in commit message) why this should be kept? Some of the other early exits do not seem to do that.

src/libFLAC/stream_decoder.c
```diff
@@ -2123,13 +2123,15 @@ FLAC__bool read_frame_(FLAC__StreamDecoder *decoder, FLAC__bool *got_a_frame, FL
...
 		if(do_full_decode) {
 			for(channel = 0; channel < decoder->private_->frame.header.channels; channel++) {
 				memset(decoder->private_->output[channel], 0, sizeof(FLAC__int32) * decoder->private_->frame.header.blocksize);
 			}
 		}
+		decoder->protected_->state = FLAC__STREAM_DECODER_SEARCH_FOR_FRAME_SYNC;
+		return true;
```
kiilerix, line 2134: Can you clarify (perhaps in commit message) why it should return true? Some of the other early exits return false. It doesn't seem obvious that the existing code is consistent and correct, so it is non-trivial to argue that this is correct.

ktmf01:
@kiilerix You're right, this is quite sloppy. The truth is that I wasn't sure whether the memset really had no effect. Today I found another reason to take a good look at this bit of code, someone pointed out possible improvement of error handling: https://hydrogenaud.io/index.php?topic=122094.msg1007717#msg1007717

I'll come back to this and set this PR to draft for the time being

kiilerix:
@ktmf01 I very much appreciate and respect that you take my feedback as constructive feedback - as it was intended.

From a quick look at the hydrogenaud discussion, it seems like another problem, and an example where Postel's law isn't helpful. If some input is malformed, it really is helpful to reject it as early and strictly and clearly as possible. That helps understanding and fixing the root cause.

For the "seek to random position and avoid sync'ing incorrectly" it is essential to know that we are right in rejecting even the slightest error as a reliable indication that we didn't find a well-formed frame header. If we try to be forgiving and resilient om what we accept, there is no way we with any certainty can do "random search" reliably.


If I understand you correctly, I agree with your initial point that the flac decoder really should reject frames with incorrect CRC. Even/especially if that goes against Postel's law, and even if that might seem to have negative consequences in some cases. That shouldn't prevent us from doing the right thing.

(Also, it seems like "doing the right thing" happens to fix my problem. I like that.)

ktmf01:
> From a quick look at the hydrogenaud discussion, it seems like another problem, and an example where Postel's law isn't helpful. If some input is malformed, it really is helpful to reject it as early and strictly and clearly as possible. That helps understanding and fixing the root cause.

I was specifically referring to this (emphasis mine)

>Good but with a minus or two: FLAC. Like WavPack it isn't straightforward for a novice user to get out the position of the corruption. (But VUPlayer's audiotester.exe does, unlike for WavPack.)
But flac.exe -d - F drops a block here, **shortening the audio**. So also does ffmpeg. (FLACCL and Flake refuse to decode.)
Also flac.exe -d -F  outputs a .wav that foo_verifier gives "Warning: Indicated RIFF size exceeds actual file size, file appears to be truncated

The code that you were referring to (the code that did nothing but I left in anyway) is meant to keep the audio the same length, but replace the corrupt section with silence. This has nothing to do with seeking of course, but if I file two PRs touching the same bit of code and they are merged in, lets say, 6 months from now, I won't remember which one should take precedence. That's why I want to fix these two problems in one PR. 

In all cases of corruption during normal decoding, be it the frame header or the frame body, silence should be added to the output to keep in the same length, if possible. In current code, this is only done when the whole frame is parsed correctly but CRC doesn't match.

kiilerix:
Yes, agreed: Before, frames that had bad CRC (but didn't trigger any other checks) would be replaced by silence.

I think it is a very bad idea to silently add silence. That is *very* lossy, and the exact opposite of the loss-less nature of flac. In many cases it would be much preferred to fail loud and clear so the problem could be fixed at the root cause. That is especially the case for my use case of doing audio production. Silent skipping might be convenient for *some* casual playback use cases, but that sloppiness is something an application can build on top of the 100% reliable flac library.

But with the change proposed here, the code will return without invoking `write_audio_frame_to_client_` with the callback. I thus don't see any point in keeping the zeroing. Keeping it will add tech debt.

ktmf01:
The proposed change still creates loud and clear warning messages. It is up to the libFLAC client to do something with this. the `flac` command line utility for example, stops decoding and throws a clear error. Only when invoked with -F `flac` returns a file with errors. With the PR as it is now, a certain set of samples is simply removed, whereas with the proposed change, the length will remain correct (so syncing with whatever else will stay correct as well) as long as it is only a single frame with a problem and the problematic part is silent. This seems more graceful behaviour.

Once more: what to do in case of an error is not up to libFLAC, but the program using it. libFLAC should be as flexible as possible catering to the needs of a non-realtime command line program (like `flac`) and to a decoder application in for example broadcasting or archival. In case of `flac`, complete failure is preferred unless overriden, but in for example broadcasting graceful failure (like adding silence) is a must. Both can be achieved as error handling is left to the program using libFLAC.

kiilerix:
But with the change proposed here, when/how should the program using libFLAC see the cleared output buffer when we don't invoke the write callback?

ktmf01:
I'm still working on the code, with proposed change I meant the idea behind it.

ktmf01:
@kiilerix I decided to split the two things after all, the patch in #283 has become quite complicated, while this one is very simple.

kiilerix (approved):

retokromer (approved):

ktmf01:
This PR is no longer applicable. Apparently I already took care of this in #283

PR #282, opened 2022-02-02
Replace seek error with handling when hitting EOF while reading frame

ktmf01:
This fixes the issue mentioned by @wader in https://github.com/xiph/flac/issues/276#issuecomment-1026603582 and for which the file can be found here: https://trac.ffmpeg.org/ticket/9621

The issue resolved here caused a seek error when seeking to a sample near the end of the stream, when the last frame is truncated or a fake header causes libFLAC to try to read beyond the end of the stream.


wader:
What would happen if there was more valid frames after frame 26 so that reading the fake/false frame would not hit end of stream? i can probably cut out a bit more from the original file with this issue if you want a test case.

ktmf01:
Such a file would presumably not need this patch to seek fine. It might need #276 though.

wader:
Ah yes sorry I should have been clearer, was afraid that the test file was misleading. I think ffmpeg fails on footer crc before any eof error.

PR #283, opened 2022-02-05
Rework error handling

ktmf01:
This commit reworks the code decoding a frame, to add silence when frames are missing and output silence when something other than the frame header seems corrupted. This originates from issue #274 and #276 and PR #277

Because this PR changes the code around quite a bit this probably needs to be thoroughly fuzzed. ~That's why it's probably best not included in the next release.~ It shouldn't be included shortly before doing a release.


ktmf01:
The last commit adds a check to detect files such as https://sourceforge.net/p/flac/bugs/468/ and not output clipping when it happens.

PR #284, opened 2022-02-06
Wrap find_package Ogg with a "found" condition

martinRenou:
This is needed in the case of using `add_subdirectory` with libflac in a build tree containing Ogg


martinRenou:
Replaced by https://github.com/xiph/flac/pull/285

PR #287, opened 2022-02-09
Add CIFuzz to workflows

ktmf01:
This should enable 'continuous integration fuzzing' or CIFuzz. For details see https://google.github.io/oss-fuzz/getting-started/continuous-integration/


rillian:
Thanks for setting this up. Merged the first patch as e8143ab5277ec3c4f857187a1d1d2e773f0a7c92.

Should we be defining `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` somewhere?

PR #288, opened 2022-02-09
Update changelog and change version numbers

ktmf01:
Two things to note:

- I've added a depreciation notice to the README and changelog about building with MSVC with the solution files instead of CMake. This might be controversial
- This will intentionally not pass CI because the release dates don't match. This has to be changed to the date of the actual release, but now it is documented where these dates need to be changed


rillian:
Thanks for writing a changelog. I merged this into the 1.3.x backport branch instead. Once that's ready we can copy the changelog back onto the main branch and the flac-website repo.

PR #292, opened 2022-02-21
Fix uclibc build

ffontaine:
Commit 4fbb6d4f2ecf2a96c17ea9880108409f852c08a9 reverted commit 44036c9a9b45d03373fe90e9c112852bfc054c51 and so broke again the build on uclibc

So put back the `sys/auxv.h` check as well as a `getauxval` check in a way that doesn't break iOS

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/src/libFLAC/cpu.c
+++ b/src/libFLAC/cpu.c
@@ -53,10 +53,8 @@
 #define dfprintf(file, format, ...)
 #endif
 
-#if defined FLAC__CPU_PPC
-#if defined(__linux__) || defined(__FreeBSD__)
+#if defined(HAVE_SYS_AUXV_H)
 #include <sys/auxv.h>
-#endif
 #endif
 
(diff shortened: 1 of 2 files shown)


ktmf01:
LGTM. Note on merging: It touches the same code as #293 though (but PRs try to solve different problems) so whichever is merged last will need some attention

src/libFLAC/cpu.c
```diff
@@ -247,7 +245,7 @@ ppc_cpu_info (FLAC__CPUInfo *info)
 #define PPC_FEATURE2_ARCH_2_07		0x80000000
 #endif
 
-#ifdef __linux__
+#if defined (__linux__) && defined(HAVE_GETAUXVAL)
```
ktmf01, line 248: Shouldn't the first elif that follows this line also get a `&& defined(HAVE_GETAUXVAL)`? Like so:
```
#elif defined(__FreeBSD__) && defined(HAVE_GETAUXVAL)
```

ffontaine: I updated the PR as requested

PR #293, opened 2022-02-24
Fix hwcaps type for FreeBSD/powerpc*

pkubaj:
According to elf_aux_info() manpage, AT_HWCAP2 is of u_long type.
Feature detection works anyway in this case, but u_long is more correct.
While here, also remove FreeBSD 11 support, it's EOL.

### DIFF
--- a/src/libFLAC/cpu.c
+++ b/src/libFLAC/cpu.c
@@ -54,7 +54,7 @@
 #endif
 
 #if defined FLAC__CPU_PPC
-#if defined(__linux__) || (defined(__FreeBSD__) && (__FreeBSD__ >= 12))
+#if defined(__linux__) || defined(__FreeBSD__)
 #include <sys/auxv.h>
 #endif
 #endif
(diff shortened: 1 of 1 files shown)


petterreinholdtsen:
[pkubaj]
> While here, also remove FreeBSD 11 support, it&#39;s EOL.

Even if it is declared end of life from the FreeBSD developers side, do
we really need to make life even harder for the poor people forced to
use FreeBSD 11 or earlier versions by removing the trivial code to
support FreeBSD < 12?

-- 
Happy hacking
Petter Reinholdtsen

pkubaj:
People on FreeBSD 11 probably can build this anyway by disabling SIMD. FreeBSD 11 doesn't support modern POWER, so it shouldn't really be a problem.

FreeBSD 12.0-RELEASE is the first release with proper support for POWER8 and POWER9, with 13.0-RELEASE having huge improvements.

As such, supporting POWER8 and POWER9 SIMD on FreeBSD 11 doesn't really make sense.

ktmf01:
> People on FreeBSD 11 probably can build this anyway by disabling SIMD.

If a clear error message with instructions on how to fix it would appear, this is indeed not a problem. However, I don't think that is the case here. Leaving a broken build behind doesn't seem a good idea to me, even on system which no longer receive updates.

I have no experience with FreeBSD, nor with POWER, so I cannot check this myself: Does FLAC build properly on FreeBSD 11 on POWER8/9 without this patch, and will it build properly after applying this patch? Are extra options for configure necessary?

pkubaj:
I'm not sure whether it builds, I'm actually not sure whether it even builds on 12. I wouldn't be surprised if not. FreeBSD 12 on POWER uses GCC 4.2 (the reason being newer GCC licensed under GPLv3) and 13 switched to LLVM. I urged anyone using FreeBSD on POWER to switch to 13.
12 is officially supported, but no one from FreeBSD/POWER team does any fixes on 12. It's basically considered as a waste of time, with 13 having newer ABI, new toolchain etc.

However, if that part about dropping support for 11 is the problem, I may just revert it.

ktmf01:
I just took another look at the code and as I understand it now it works as follows

- Without this patch, FreeBSD 11 on PPC will throw an error during compilation: `Unsupported platform! Please add support for reading ppc hwcaps.`
- With this patch, FreeBSD on PPC will throw an error during compilation, something akin to `fatal error: 'sys/auxv.h' file not found`

The second error (so, with your patch as it is now) will probably give more results in a search engine pointing a potential user in the right direction. So I've changed my mind, probably best if you keep this PR as it is now.

Looks good to me

PR #294, opened 2022-02-28
Move initialization of state up in stream_encoder_new

ktmf01:
In stream_encoder_new, set_defaults_ was called before setting encoder->protected_->state to uninitialized. However, this state
was being accessed by set_compression_level in set_defaults_. As this is undefined behaviour, move the initialization of encoder->protected_->state up to before calling set_default_

### DIFF
--- a/src/libFLAC/stream_encoder.c
+++ b/src/libFLAC/stream_encoder.c
@@ -536,6 +536,8 @@ FLAC_API FLAC__StreamEncoder *FLAC__stream_encoder_new(void)
 
 	encoder->private_->file = 0;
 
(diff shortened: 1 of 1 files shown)


ktmf01:
This was also the cause of `set_compression_level` in `set_defaults_` not doing anything, so this PR changes libFLAC behavior when initialized with no specific compression level set. Without this patch, compression level defaults to 0, with this change it defaults to 5 (as was intended)

ktmf01:
CIFuzz fails because of #273, but this change improves coverage of fuzzer_encoder, causing it to find even more bugs caused which would be solved by #273

PR #305, opened 2022-03-22
add strings.h

neheb:
Needed for strncasecmp under macOS.

Signed-off-by: Rosen Penev <rosenp@gmail.com>

### DIFF
--- a/src/libFLAC/metadata_object.c
+++ b/src/libFLAC/metadata_object.c
@@ -36,6 +36,9 @@
 
 #include <stdlib.h>
 #include <string.h>
+#ifndef _MSC_VER
+#include <strings.h>
+#endif
 
 #include "private/metadata.h"
 #include "private/memory.h"


ktmf01:
Could you perhaps explain a little more why this change is necessary? The fact it doesn't come up in CI worries me. The explanation that is needed under MacOS is in contradiction with CI building properly under MacOS?

neheb:
I think the reason is that it implicitly gets included because of some cmake wizardry: https://github.com/xiph/flac/blob/master/cmake/UseSystemExtensions.cmake#L50

ktmf01:
Besides CMake, CI also build with autotools on MacOS, so that doesn't explain it.

neheb:
No idea. Maybe strings.h gets included somehow internally. Anyway, it's a POSIX header.

neheb:
Yeah, -D_DARWIN_C_SOURCE includes it.

sezero:
This should depend on `HAVE_STRINGS_H` instead. The autotools-generated config.h
has that because of `AC_HEADER_STDC`.  cmake'ry, on the other hand, needs adding
a specific check for it with config.cmake.h.in updated.

In addition, the header should be included in all sources where `str[n]casecmp`
are used either explicitly or as `FLAC__STR[N]CASECMP`.

PR #309, opened 2022-04-01
Add missing langinfo check for CMake

### LINKED ISSUE #308: Building with CMake never defines HAVE_LANGINFO_CODESET
Buliding with CMake + Ninja never adds `-DHAVE_LANGINFO_CODESET` to compiler options, so locale detection is not complete in such situation.

Notably, ArchLinux now builds flac with CMake + Ninja (https://github.com/archlinux/svntogit-packages/blob/packages/flac/trunk/PKGBUILD), which causes some locale issue when tagging FLAC files. See: https://github.com/svend/cuetools/issues/38#issuecomment-1082859037

NotTsunami:
Closes #308

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,6 +79,14 @@ check_function_exists(fseeko HAVE_FSEEKO)
 
 check_c_source_compiles("int main() { return __builtin_bswap16 (0) ; }" HAVE_BSWAP16)
 check_c_source_compiles("int main() { return __builtin_bswap32 (0) ; }" HAVE_BSWAP32)
+check_c_source_compiles("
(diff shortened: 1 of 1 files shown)


NotTsunami:
Need to fix this check, not ready for merge

NotTsunami:
The test is now passing for Ubuntu. Untested on Arch, but it should fix the issue observed in #308.

NotTsunami:
Small note: for some reason CI is failing on macOS right now, but it seems to be an issue with homebrew and unrelated to the PR

ktmf01:
Yes, I noticed. I'll try triggering a rerun in an hour, if that doesn't work tomorrow

Issue #303, opened 2022-03-10
cc1: error: unrecognized command line option "-fstack-protector-strong"


barracuda156:
`FLAC` fails to build on Leopard for `ppc` since the build forces `-fstack-protector-strong` flag despite configure established that compiler does not support it: 
```
libtool: compile:  /usr/bin/gcc-4.2 -DHAVE_CONFIG_H -I. -I../.. -I../.. -I./include -I../../include -I/opt/local/include -D_FORTIFY_SOURCE=2 -DFLAC__USE_VISIBILITY_ATTR -faltivec -I/opt/local/include -pipe -Os -std=gnu89 -arch ppc -Wall -Wextra -Wstrict-prototypes -Wmissing-prototypes -Waggregate-return -Wcast-align -Wnested-externs -Wshadow -Wundef -Wmissing-declarations -Winline -Wdeclaration-after-statement -fvisibility=hidden -fgnu89-inline -msse2 -fstack-protector-strong -MT cpu.lo -MD -MP -MF .deps/cpu.Tpo -c cpu.c  -fno-common -DPIC -o .libs/cpu.o
cc1: error: unrecognized command line option "-fstack-protector-strong"
cc1: error: unrecognized command line option "-fstack-protector-strong"
cc1: error: unrecognized command line option "-fstack-protector-strong"
cc1: error: unrecognized command line option "-fstack-protector-strong"
make[4]: *** [cpu.lo] Error 1
make[4]: *** Waiting for unfinished jobs....
make[4]: *** [bitwriter.lo] Error 1
make[4]: *** [bitreader.lo] Error 1
make[4]: *** [bitmath.lo] Error 1
make[4]: Leaving directory `/opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_audio_flac/flac/work/flac-1.3.4/src/libFLAC'
make[3]: *** [all-recursive] Error 1
make[3]: Leaving directory `/opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_audio_flac/flac/work/flac-1.3.4/src/libFLAC'
make[2]: *** [all-recursive] Error 1
make[2]: Leaving directory `/opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_audio_flac/flac/work/flac-1.3.4/src'
make[1]: *** [all-recursive] Error 1
[...]

PR #310, opened 2022-04-01
Fix stack protector detection for C++

NotTsunami:
Closes https://github.com/xiph/flac/issues/303


ktmf01:
Thanks for rebasing after #271

PR #314, opened 2022-04-18
remove completely unused headers

neheb:
none of these checks are used by the code.

Signed-off-by: Rosen Penev <rosenp@gmail.com>

### DIFF
--- a/cmake/UseSystemExtensions.cmake
+++ b/cmake/UseSystemExtensions.cmake
@@ -16,37 +16,18 @@ endif()
 check_c_source_compiles("
     #define __EXTENSIONS__ 1
     #include <stdio.h>
-    #ifdef HAVE_SYS_TYPES_H
-    # include <sys/types.h>
-    #endif
-    #ifdef HAVE_SYS_STAT_H
-    # include <sys/stat.h>
-    #endif
     #ifdef STDC_HEADERS
     # include <stdlib.h>
     # include <stddef.h>
-    #else
-    # ifdef HAVE_STDLIB_H
-    #  include <stdlib.h>
-    # endif
     #endif
     #ifdef HAVE_STRING_H
-    # if !defined STDC_HEADERS && defined HAVE_MEMORY_H
-    #  include <memory.h>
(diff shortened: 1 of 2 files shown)


NotTsunami:
This patch is incorrect, or at the very least, the changes to `UseSystemExtensions.cmake` are incorrect. I looked into this and the __EXTENSIONS__ macro, when used in autoconf, "is defined only if the headers included by AC_INCLUDES_DEFAULT (see [Default Includes](https://www.gnu.org/software/autoconf/manual/autoconf-2.70/html_node/Default-Includes.html)) work correctly with it defined". The headers it checks against are to test if __EXTENSIONS__ should be defined for Solaris. This wouldn't break Solaris building by any means, but it would cause a bigger rift between autotools and CMake builds, and potentially create a scenario where __EXTENSIONS__ is erroneously defined (not that it would have effect out of Solaris, to my knowledge, but it would still be bad practice).

As for the second portion, I'm unsure of the validity of those headers. I regenerated the config using `autoheader` and those headers are still included, so I'm inclined to believe they should be left there, but I can't comment for certain.

NotTsunami:
This patch is incorrect, or at the very least, the changes to `UseSystemExtensions.cmake` are incorrect. I looked into this and the __EXTENSIONS__ macro, when used in autoconf, "is defined only if the headers included by AC_INCLUDES_DEFAULT (see [Default Includes](https://www.gnu.org/software/autoconf/manual/autoconf-2.70/html_node/Default-Includes.html)) work correctly with it defined". The headers it checks against are to test if __EXTENSIONS__ should be defined for Solaris. This wouldn't break Solaris building by any means, but it would cause a bigger rift between autotools and CMake builds, and potentially create a scenario where __EXTENSIONS__ is erroneously defined (not that it would have effect out of Solaris, to my knowledge, but it would still be bad practice).

As for the second portion, I'm unsure of the validity of those headers. I regenerated the config using `autoheader` and those headers are still included, so I'm inclined to believe they should be left there, but I can't comment for certain.

PR #325, opened 2022-04-21
Enable building ogg together with flac

ktmf01:
This reverts the part of commit 2013738 that searched for a local ogg build

With this commit, a user only has place the ogg sources in as a subdirectory of flac with the name ogg (for example by `cd flac; git clone https://github.com/xiph/ogg`) and CMake will pick up and include it. The resulting build is seamless


NotTsunami:
Should we add /ogg to the .gitignore too then? I noticed when building with CMake in Visual Studio, the working tree gets really dirty, and was thinking about sending in a patch to update the .gitignore.

PR #326, opened 2022-04-21
Update .gitignore for VS artifacts

NotTsunami:
When building with CMake on Visual Studio, the working tree still gets pretty dirty with some Visual Studio artifacts. This commit adds in some extra Visual Studio-specific excludes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,8 +42,6 @@ m4/lt~obsolete.m4
 man/manpage.links
 man/manpage.refs
 missing
-objs/debug
-objs/release
 src/flac/flac
 src/libFLAC++/flac++.pc
 src/libFLAC/flac.pc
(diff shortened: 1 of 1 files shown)


ktmf01:
Wouldn't ignoring `*.dir/` be better? Currently this PR doesn't change the output of my `git status` to be any cleaner.

Still, we can't ignore `*.vcxproj` and `*.vcxproj.filters` yet. Perhaps it is wise to encourage out-of-tree builds for MSVC because of this? When I build out of tree (with `mkdir cmakebuild; cmake ..` or selecting such a directory in cmake-gui) I get no additions to `git status`, not even when the build directory is a subdirectory of the tree.

NotTsunami:
> Wouldn't ignoring `*.dir/` be better? Currently this PR doesn't change the output of my `git status` to be any cleaner.
> 
> Still, we can't ignore `*.vcxproj` and `*.vcxproj.filters` yet. Perhaps it is wise to encourage out-of-tree builds for MSVC because of this? When I build out of tree (with `mkdir cmakebuild; cmake ..` or selecting such a directory in cmake-gui) I get no additions to `git status`, not even when the build directory is a subdirectory of the tree.

Naturally. The .vcxproj* files aren't touched intentionally. I had a lot of .pdb, .tlog, .log and .obj files on my last working tree, perhaps I opened it via the solution or accidentally used MSVC instead of CMake.

PR #335, opened 2022-05-02
Fix ARM Neon dispatch functions handling

RonenGvili:
Latest changes committed to cpu.h and cpu.c overwrote the dispatch logic for the Neon optimized functions, leaving them unused.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,6 +78,7 @@ microbench/benchmark_residual
 /ogg/
 oss-fuzz/fuzzer_decoder
 oss-fuzz/fuzzer_encoder
+oss-fuzz/fuzzer_encoder_v2
 
 /*[Bb]uild*/
 /out/
--- a/config.cmake.h.in
+++ b/config.cmake.h.in
@@ -42,6 +42,7 @@
 
 /* Set to 1 if <arm_neon.h> is available. */
 #cmakedefine01 FLAC__HAS_NEONINTRIN
+#cmakedefine01 FLAC__HAS_A64NEONINTRIN
 
 /* define if building for Darwin / MacOS X */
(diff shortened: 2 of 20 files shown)


RonenGvili:
HI @ktmf01 . Regarding the changes in cpu.h and cpu.c - notice that the only way to dispatch the 2 Neon function: FLAC__lpc_compute_residual_from_qlp_coefficients_intrin_neon() and FLAC__lpc_compute_residual_from_qlp_coefficients_wide_intrin_neon() in run time is in case of use_asm is true. We do not need arm_cpu_info() function (like the ppc_cpu_info and x86_cpu_info) as it always true. Without this flag using ARM device will dispatch the scalar c code for the lpc_compute_residuals instead of the Neon functions.

ktmf01:
> notice that the only way to dispatch the 2 Neon function [..] in run time is in case of use_asm is true. 

I'll change that as well. Adding dead code doesn't seem like a good idea to me.

ktmf01:
I have made a new PR #336 with the change as I propose it, as this PR is currently not mergeable. Do you agree with that approach?

RonenGvili:
> I have made a new PR #336 with the change as I propose it, as this PR is currently not mergeable. Do you agree with that approach?

I agree.

ktmf01:
Closing this because of the merge of #336 @RonenGvili Thanks for spotting this problem. I only properly tested with autotools, not with cmake. Though, next time you send in a PR, it should be just the commits you want to merge, without a large sum of related (but already merged) commits. If you don't know how to do that, you could search the internet for `git rebase`. Rebasing is letting git apply your changes on top of another branch (usually the master or main branch). You can also 'squash' several commits into a single one with this command.

PR #337, opened 2022-05-04
Speedup of fixed_compute_best_predictor functions

ktmf01:
The code of fixed_compute_best_predictor was presumably once optimized, but it appears a more readable form is now much faster: measurements on a Raspberry Pi 4 (64-bit ARMv8) indicate this code runs almost twice as fast. One could conjecture this is because the code was optimized before auto-vectorization of code by compilers was feasible, and the code as it was cannot (currently) be properly optimized.

### DIFF
--- a/src/libFLAC/fixed.c
+++ b/src/libFLAC/fixed.c
@@ -219,21 +219,39 @@ uint32_t FLAC__fixed_compute_best_predictor(const FLAC__int32 data[], uint32_t d
 uint32_t FLAC__fixed_compute_best_predictor(const FLAC__int32 data[], uint32_t data_len, FLAC__fixedpoint residual_bits_per_sample[FLAC__MAX_FIXED_ORDER+1])
 #endif
 {
+	FLAC__uint32 total_error_0 = 0, total_error_1 = 0, total_error_2 = 0, total_error_3 = 0, total_error_4 = 0;
+	uint32_t order;
+#if 0
+	/* This code has been around a long time, and was written when compilers weren't able
+	 * to vectorize code. These days, compilers are better in optimizing the next block
+	 * which is also much more readable
+	 */
 	FLAC__int32 last_error_0 = data[-1];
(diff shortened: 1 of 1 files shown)


ktmf01:
Results of testing compression level -2, CDDA on ARM64
GCC 10.2.1: 3% faster
Clang 11: 2% slower

Results of testing compression level -2, 24-bit on ARM64
GCC 10.2.1: 19% faster
Clang 11: 2% faster

Results of testing compression level -2, CDDA on x86_64 (Intel Coffee Lake-S), asm optimizations disabled
GCC 9.4.0: 13% faster
Clang 10: 2% slower

Results of testing compression level -2, 24-bit on x86_64 (Intel Coffee Lake-S), asm optimizations disabled
GCC 9.4.0: 25% faster
Clang 10: 4% slower

These results show Clang disagrees with GCC on whether the code with or without this patch is easier to optimize. I will test tomorrow on another machine x86_64 with more recent GCC and Clang versions

ktmf01:
Results of using GCC 11.2.0 and Clang 14 on an MSYS2 environment running on a Intel Kaby Lake-R CPU.

Presets tested are -0 through -5 from top right to bottom left. flac-fixed-simple-\*.exe is with this patch, flac-fixed-complicated-\*.exe is without it.
[simpler_fixed_compute_best_predictor.pdf](https://github.com/xiph/flac/files/8630641/simpler_fixed_compute_best_predictor.pdf)
[simpler_fixed_compute_best_predictor_wide.pdf](https://github.com/xiph/flac/files/8630640/simpler_fixed_compute_best_predictor_wide.pdf)

From this, it seems GCC performs much better with this patch, while clang performed better than GCC on the code without this patch, and performance slightly worsens with this patch.

Still, seeing the improvement with GCC is significantly larger than the decrease in performance with Clang, and because the code readability is much improved, I'm inclined to merge this PR.

PR #339, opened 2022-05-05
Rename FLAC__get_decoder_client_data to FLAC__stream_decoder_get_client_data

ktmf01:
Also the function definition was moved to its 'siblings'. This way, the function fits better in the API with respect to naming. This implements the change [suggested here](https://github.com/xiph/flac/pull/124#issuecomment-519453916)

@sezero  and @ChristopD do you agree with this? It is yet another change to the name, but as the first change has been kept out of 1.3.4, I think this should be fixed while we still can.


sezero:
I agree and have no objections.

ChristopD:
This is ok for me too.
We was surprised to find this function with initial name in the current release. Our renaming task is still in standby, waiting for next update.
Thank you.

ktmf01:
@ChristopD The change was left out of 1.3.4 because a release with security fixes was long overdue, and this change is one which changes the ABI/API in a breaking way. As there are a few more API changes coming up, we chose to group them into a next release.

Issue #137, opened 2019-09-10
should use libogg.lib instead of libogg_static.lib


fuchanghao:
In ogg 1.3.4 it's output-lib changed to libogg.lib not libogg_static.lib.

flac should change the lib name too.

erikd:
I don't use the CMake build system.  Please feel free to provide a PR to fix this.

jbaker6953:
The problem with that is that flac.exe 1.3.4 is then built dependent on ogg.dll rather than statically. I came here to report this very issue.

jbaker6953:
As a side note: the libogg source builds libogg.dll, but flac project looks for ogg.dll.

fuchanghao:
> 
> 
> As a side note: the libogg source builds libogg.dll, but flac project looks for ogg.dll.

No, it needs libogg_static.lib in the objs\release\lib path.

But from ogg 1.3.4 the name of library called libogg.lib not libogg_static.lib.

jbaker6953:
What I mean is that if you build flac with libogg.lib flac.exe won't run. It throws an error about missing 'ogg.dll'.

zvezdochiot:
@jbaker6953 say> What I mean is that if you build flac with libogg.lib flac.exe won't run. It throws an error about missing 'ogg.dll'.

What is this static? This is a dynamic link, only a curve.

NotTsunami:
Pinging @krokoziabla as he's the one who most recently worked on the CMake files

ktmf01:
CMake had quite an overhaul with e0b62a6, especially with respect to ogg. I haven't had problems with this in a while, neither has CI. Can I close this issue?

PR #353, opened 2022-05-18
Remove MSVC and Makefile.lite build system

### LINKED ISSUE #100: Remove Visual Studio project files
Once the CMake build system has been merged (#97) it should be possible to remove the VS project files. Posting this idea here so that people can comment.

The VS project files will not be done until *after* the next release, which will contain both the CMake build system *and* the existing VS project files.

ktmf01:
As discussed on the flac-dev mailinglist here: http://lists.xiph.org/pipermail/flac-dev/2022-May/006537.html Closes #100 The removal of the MSVC build system has been announced in the changelog and README of the last release (1.3.4)

Also a few non related make distcheck fixes are added.


ktmf01:
This also fixes #137 rather indirectly

PR #361, opened 2022-05-24
Add 32 bit encoding and decoding capabilities

ktmf01:
This is still a work in progress. Decoding and encoding capabilities will probably be merged separately.

### DIFF
--- a/include/FLAC/format.h
+++ b/include/FLAC/format.h
@@ -282,14 +282,24 @@ extern FLAC_API const char * const FLAC__SubframeTypeString[];
 /** CONSTANT subframe.  (c.f. <A HREF="../format.html#subframe_constant">format specification</A>)
  */
 typedef struct {
-	FLAC__int32 value; /**< The constant signal value. */
+	FLAC__int64 value; /**< The constant signal value. */
 } FLAC__Subframe_Constant;
+
+/** An enumeration of the possible verbatim subframe data types. */
+typedef enum {
+	FLAC__VERBATIM_SUBFRAME_DATA_TYPE_INT32, /**< verbatim subframe has 32-bit int */
+	FLAC__VERBATIM_SUBFRAME_DATA_TYPE_INT64 /**< verbatim subframe has 64-bit int */
+} FLAC__VerbatimSubframeDataType;
 
 
 /** VERBATIM subframe.  (c.f. <A HREF="../format.html#subframe_verbatim">format specification</A>)
  */
 typedef struct {
-	const FLAC__int32 *data; /**< A pointer to verbatim signal. */
+	union {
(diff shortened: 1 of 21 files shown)


include/FLAC/format.h
```diff
@@ -282,14 +282,24 @@ extern FLAC_API const char * const FLAC__SubframeTypeString[];
...
+	FLAC__VERBATIM_SUBFRAME_DATA_TYPE_INT64 /**< verbatim subframe has 64-bit int */
+} FLAC__VerbatimSubframeDataType;
+
 
 /** VERBATIM subframe.  (c.f. <A HREF="../format.html#subframe_verbatim">format specification</A>)
  */
 typedef struct {
-	const FLAC__int32 *data; /**< A pointer to verbatim signal. */
+	union {
+		const FLAC__int32 *int32; /**< A FLAC__int32 pointer to verbatim signal. */
```
sezero, line 299: Bad choice of member name: many C sources out there define `int32` and/or `int64` types for themselves.

sezero: `data32` and `data64` maybe???

include/FLAC/ordinals.h
```diff
@@ -82,4 +82,20 @@ typedef FLAC__uint8 FLAC__byte;
...
+ * 64 bit signed int or 32 bit signed int */
+
+typedef enum {
+        FLAC__FLEX_INT_POINTER_DATA_TYPE_INT32,
+        FLAC__FLEX_INT_POINTER_DATA_TYPE_INT64
+} FLAC__FlexIntPointerDataType;
+
+typedef struct {
+        union {
+                const FLAC__int32 *int32;
```
sezero, line 95: same as above

include/FLAC/format.h
ktmf01: Thanks for the feedback. Do you think the general idea (having a union in a struct with a member indicating which union member is being used) uses the right concept for the job here?

sezero: A simple `void*` pointer cast to correct type where it is used would have done the job.

ktmf01: Thanks for the suggestion. I'll try and see whether that makes the code cleaner or not.

ktmf01:
I just rebased because I moved the changes for CI to the right commits. Also, I moved from FlexIntPointer to simply casting to void*, as suggested

ktmf01:
@ValZapod would you perhaps be willing to test this PR? You requested this functionality in #245

ValZapod:
Well, I would not be a be to say anything past the fact: oh, yes, it is lossless in all 32 bits, not 24 bits in 32 bits! And to compare wav in beyond compare or ffmpeg md5...

PR #368, opened 2022-05-30
Update CI build descriptions

rillian:
- Add a `make distcheck` job to github actions, to better catch regressions.
- Drop the removed `Makefile.lite` build from the gitlab ci jobs.


.github/workflows/distcheck.yml
```diff
@@ -0,0 +1,35 @@
...
+      - name: Install Linux dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libtool-bin libogg-dev doxygen libxml2-utils w3c-sgml-lib
+
+      - name: Build with Autotools
+        run: |
+          ./autogen.sh
+          ./configure
+          make
```
ktmf01, line 25:
```suggestion
```
I think this can be removed, now everything is build twice. While it doesn't take long, it makes the logs unnecessarily long

ktmf01, line 25:
```suggestion
```
I think this can be removed, now everything is build twice. While it doesn't take long, it makes the logs unnecessarily long

.github/workflows/distcheck.yml
```diff
@@ -0,0 +1,35 @@
...
+
+      - name: Install Linux dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libtool-bin libogg-dev doxygen libxml2-utils w3c-sgml-lib
+
+      - name: Build with Autotools
+        run: |
+          ./autogen.sh
+          ./configure
```
ktmf01, line 24:
```suggestion
          ./configure --disable-thorough-tests
```
The test suite is already run in all it's facets, so I think running a slimmed down one here should be fine. After all, we mostly want to know whether it builds and runs, I guess?

ktmf01, line 24:
```suggestion
          ./configure --disable-thorough-tests
```
The test suite is already run in all it's facets, so I think running a slimmed down one here should be fine. After all, we mostly want to know whether it builds and runs, I guess?

ktmf01:
Thanks for fixing this, I remember now that at some point I noticed the gitlab CI ran tests for Makefile.lite, but I forgot to remove these.

While we're at the subject, could you take a look at whether the ci directory is still necessary? I think it was at some point used for some Xiph CI at some point, but I think it is no longer in use.

.github/workflows/distcheck.yml
rillian: I don't think this will have any effect. `make distcheck` invokes configure again; the outer build options aren't propagated.

There is an `(AM_)DISTCHECK_CONFIGURE_FLAGS` you can set if you want to do this.

rillian:
> could you take a look at whether the ci directory is still necessary? I think it was at some point used for some Xiph CI at some point, but I think it is no longer in use.

Yes, it looks like `ci/flac-autotool.sh` isn't used any more and can be removed. Likewise with `.travis.yml`.

.github/workflows/distcheck.yml
ktmf01: Ok, probably best to leave it like this then

ktmf01:
Actually, I still use .travis.yml myself. Once in a while I manually trigger a CI run on my fork of flac. The reason for this is that Travis offers CI on PowerPC and ARM, which Github doesn't offer. See https://app.travis-ci.com/github/ktmf01/flac/builds (errors are mostly time-outs because the PPC runners tend to be busy quite often)

rillian:
> Actually, I still use .travis.yml myself. Once in a while I manually trigger a CI run on my fork of flac. The reason for this is that Travis offers CI on PowerPC and ARM, which Github doesn't offer. See https://app.travis-ci.com/github/ktmf01/flac/builds (errors are mostly time-outs because the PPC runners tend to be busy quite often)

Aha, I hadn't see that part of the setup. I've restored the travis config.

PR #380, opened 2022-06-23
Add fuzzer_metadata and fix found bugs

ktmf01:
@rillian Could you say something about the last commit of this list, where -1 is added to an enum? It is added because [this bit of API documentation](https://xiph.org/flac/api/group__flac__metadata__level0.html#gab9f69e48c5a33cacb924d13986bfb852) says

> type | The desired picture type. Use -1 to mean "any type".

But this leads to undefined behavior as -1 is not in the enum, so I added it. I'm not sure this is the best solution to the problem however, maybe you have an idea?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,6 +78,7 @@ microbench/benchmark_residual
 /ogg/
 oss-fuzz/fuzzer_decoder
 oss-fuzz/fuzzer_seek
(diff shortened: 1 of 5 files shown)


ktmf01:
I'm closing this as most of it has already been merged

PR #387, opened 2022-07-05
Add FMA intrinsics for autocorrelation calculation

ktmf01:
This adds FMA intrinsics to the encoder. Results for a Intel Kaby Lake-R are below: results of FLAC 1.3.4, current git (with slowdown caused by the change of calculating autocorrelation with double precision) and current git with this PR, for 16-bit and 24-bit inputs.

[set of short tracks.pdf](https://github.com/xiph/flac/files/9043568/set.of.short.tracks.pdf)
[set of short tracks 24bit.pdf](https://github.com/xiph/flac/files/9043567/set.of.short.tracks.24bit.pdf)


retokromer (approved):

src/libFLAC/stream_encoder.c
```diff
@@ -1026,6 +1029,15 @@ static FLAC__StreamEncoderInitStatus init_stream_internal_(
 			encoder->private_->local_lpc_compute_residual_from_qlp_coefficients_64bit = FLAC__lpc_compute_residual_from_qlp_coefficients_wide_intrin_avx2;
 		}
 #    endif
+#    ifdef FLAC__FMA_SUPPORTED
```
ktmf01, line 1032:
```
if(encoder->private_->cpuinfo.x86.fma) {
```
is missing here

Issue #369, opened 2022-05-31
[REQ] Please include both FLAC binaries & libFLAC.dll's with the next release of FLAC


Barough:
Please include both FLAC binaries & libFLAC.dll's (x86/x64) with the next release of FLAC here on the Github page.

ktmf01:
This is a duplicate of #113 

Just because I'm curious, could you please tell what you use libFLAC.dll for?

Barough:
With CUETools for example.

Issue #113, opened 2019-07-29
Provide compiled releases & include DLL files


Eagle3386:
The official [download page](https://xiph.org/flac/download.html) clearly states for the Windows package:
> **command-line tools only**, the file flac-X.Y.Z-win.zip contains both 32 and 64 bit **binaries**

I'd therefore like to request usage of GitHub's free (!) power to setup the complete build chain in order to release builds directly from/to here.
This way, developers like me don't need to setup a full C(++) environment just to build the FLAC lib they're then using in there (non-) C(++) IDE.

Eagle3386:
@erikd Any news regarding an updated build chain providing release straight off of GitHub? It'd really help other developers..

a-vrma:
> Any news regarding an updated build chain providing release straight off of GitHub? It'd really help other developers.

I'm looking into uploading artifacts from the current GitHub actions workflow and adding a Windows job. It's up to @erikd if he wants to host those on xiph.org. In the meantime you can get builds from <https://www.rarewares.org/lossless.php>

Eagle3386:
@erikd / @a-vrma How are things going so far?

PR #399, opened 2022-07-16
Warn user for incompatibility with MSVC Windows XP targetting

### LINKED ISSUE #72: FLAC 1.3.2 appears to require an SSE3 processor (Intel x86)
I have compiled the latest FLAC sources for x86 (32 bit) and they appear to require an SSE3 processor.  Specifically, I have a report from a customer with a XEON Prestonia processor that he cannot read his existing FLAC files (at all) and this processor supports SSE2 only.

I have replicated his problem here with a very old machine (which has no SSE support at all) but I have not investigated why this is happening.  It doesn't crash - it's just that FLAC__stream_decoder_xxx functions do nothing and return FALSE.  This is all happening in a release build.  I have not checked whether debug builds are similarly affected.  I am building with Visual Studio 2017.

Is this deliberate?  I really hope not.  The version I have been using up to now (1.3.1) works fine on *all* processors, even those without any form of SSE, and I value that.  I have therefore, slightly reluctantly, reverted to that.

I did see in the change log that 1.3.2 requires SSE2 but I'm not actually sure that's true.  Certainly the code as originally written has CPU detection logic and a set of dispatch functions so I don't see why this is necessary.  I also ran a quick benchmark on my Core i3 and 1.3.1 and 1.3.2 perform identically.  Overall, this all feels like a step backwards,

ktmf01:
This closes #72, as it provides an error referencing the issue for anyone targeting XP, instead of silently compiling a library that doesn't function correctly.

### DIFF
--- a/include/share/compat.h
+++ b/include/share/compat.h
@@ -138,6 +138,36 @@
 #    define PRId64 "I64d"
 #    define PRIx64 "I64x"
 #  endif
+#  ifdef _USING_V110_SDK71_
+#    error "libFLAC is not compatible with V110_SDK71. See comments in include/share/compat.h for details"
+/*
+ *************************************************************************************
+ * V110_SDK71, in MSVC 2017 also known as v141_xp, is a platform toolset that is supposed
+ * to target Windows XP. It turns out however that certain functions provided silently fail
+ * on Windows XP only, which makes debugging challenging. This problem has been reported
+ * to Microsoft, but there hasn't been a fix for years. See
+ * https://web.archive.org/web/20170327195018/https://connect.microsoft.com/VisualStudio/feedback/details/1557168/wstat64-returns-1-on-xp-always
+ *
+ * It is known that this problem affects the functions _wstat64 (used by flac_stat i.e.
+ * stat64_utf8) and _fstat64 (i.e. flac_fstat) and therefore affects both libFLAC in
+ * several places as well as the flac and metaflac command line tools
+ *
+ * As the extent of this problem is unknown and Microsoft seems unwilling to fix it,
+ * users of libFLAC are encouraged to use MinGW instead of Visual Studio when explicitly
+ * targeting Windows XP. When Visual Studio is used with this problematic toolset,
+ * remove the #error above to enable compilation, and be sure to check whether your
+ * application works properly on Windows XP. A possible workaround stub for fixing
+ * flac_fstat can be found below
+ *************************************************************************************
+static int flac_fstat(int fd, struct __stat64 *buffer) {
+	FLAC__int64 filelength = _filelengthi64 (fd);
+	if(filelength < 0)
+		return -1;
+	buffer->st_size = filelength;
+	return 0;
+}
+*/
+#  endif
 #endif /* defined _MSC_VER */
 
 #ifdef _WIN32
@@ -175,26 +205,7 @@
 
 #ifdef _WIN32
 #define flac_stat_s __stat64 /* stat struct */
-#if defined(_USING_V110_SDK71_) && !defined(V110_SDK71_HACK)
-#  error "libFLAC is not compatible with V110_SDK71. See github.com/xiph/flac/issues/72#issuecomment-934777777 for details"
-#else
-#if defined(_USING_V110_SDK71_)
-/* This code is when V110_SDK71_ is defined */
-/* WATCHOUT: Make sure to test any compile using this thoroughly
- * on a machine running XP! This hack only fixes libFLAC, not the
- * command line tools. It is recommended to use MinGW instead to
- * target Windows XP */
-static int flac_fstat(int fd, struct __stat64 *buffer) {
-	FLAC__int64 filelength = _filelengthi64 (fd);
-	if(filelength < 0)
-		return -1;
-	buffer->st_size = filelength;
-	return 0;
-}
-#else
 #define flac_fstat _fstat64
-#endif
-#endif
 #else
 #define flac_stat_s stat /* stat struct */
 #define flac_fstat fstat
--- a/include/share/compat.h
+++ b/include/share/compat.h
@@ -138,6 +138,9 @@
 #    define PRId64 "I64d"
 #    define PRIx64 "I64x"
 #  endif
+#  ifdef _USING_V110_SDK71_
+#    error "libFLAC is not compatible with V110_SDK71. See github.com/xiph/flac/issues/72#issuecomment-934777777 for details"
+#  endif
 #endif /* defined _MSC_VER */
 
 #ifdef _WIN32
--- a/include/share/compat.h
+++ b/include/share/compat.h
@@ -138,14 +138,14 @@
 #    define PRId64 "I64d"
 #    define PRIx64 "I64x"
 #  endif
-#  ifdef _USING_V110_SDK71_
-#    error "libFLAC is not compatible with V110_SDK71. See comments in include/share/compat.h for details"
+#  if defined(_USING_V110_SDK71_) && !defined(_DLL)
+#    pragma message("WARNING: This compile will NOT FUNCTION PROPERLY on Windows XP. See comments in include/share/compat.h for details")
 /*
  *************************************************************************************
  * V110_SDK71, in MSVC 2017 also known as v141_xp, is a platform toolset that is supposed
  * to target Windows XP. It turns out however that certain functions provided silently fail
- * on Windows XP only, which makes debugging challenging. This problem has been reported
- * to Microsoft, but there hasn't been a fix for years. See
+ * on Windows XP only, which makes debugging challenging. This only occurs when building with
+ * /MT. This problem has been reported to Microsoft, but there hasn't been a fix for years. See
  * https://web.archive.org/web/20170327195018/https://connect.microsoft.com/VisualStudio/feedback/details/1557168/wstat64-returns-1-on-xp-always
  *
  * It is known that this problem affects the functions _wstat64 (used by flac_stat i.e.
@@ -153,19 +153,11 @@
  * several places as well as the flac and metaflac command line tools
  *
  * As the extent of this problem is unknown and Microsoft seems unwilling to fix it,
- * users of libFLAC are encouraged to use MinGW instead of Visual Studio when explicitly
- * targeting Windows XP. When Visual Studio is used with this problematic toolset,
- * remove the #error above to enable compilation, and be sure to check whether your
- * application works properly on Windows XP. A possible workaround stub for fixing
- * flac_fstat can be found below
+ * users of libFLAC building with Visual Studio are encouraged to not use the /MT compile
+ * switch when explicitly targeting Windows XP. When use of /MT is deemed necessary with
+ * this toolset, be sure to check whether your application works properly on Windows XP.
+ * It is also possible to build for Windows XP with MinGW instead.
  *************************************************************************************
-static int flac_fstat(int fd, struct __stat64 *buffer) {
-	FLAC__int64 filelength = _filelengthi64 (fd);
-	if(filelength < 0)
-		return -1;
-	buffer->st_size = filelength;
-	return 0;
-}
 */
 #  endif
 #endif /* defined _MSC_VER */
--- a/include/share/compat.h
+++ b/include/share/compat.h
@@ -140,6 +140,7 @@
 #  endif
 #  if defined(_USING_V110_SDK71_) && !defined(_DLL)
 #    pragma message("WARNING: This compile will NOT FUNCTION PROPERLY on Windows XP. See comments in include/share/compat.h for details")
+#define FLAC__USE_FILELENGTHI64
 /*
  *************************************************************************************
  * V110_SDK71, in MSVC 2017 also known as v141_xp, is a platform toolset that is supposed
--- a/include/share/compat.h
+++ b/include/share/compat.h
@@ -138,9 +138,6 @@
 #    define PRId64 "I64d"
 #    define PRIx64 "I64x"
 #  endif
-#  ifdef _USING_V110_SDK71_
-#    error "libFLAC is not compatible with V110_SDK71. See github.com/xiph/flac/issues/72#issuecomment-934777777 for details"
-#  endif
 #endif /* defined _MSC_VER */
 
 #ifdef _WIN32
@@ -178,7 +175,26 @@
 
 #ifdef _WIN32
 #define flac_stat_s __stat64 /* stat struct */
+#if defined(_USING_V110_SDK71_) && !defined(V110_SDK71_HACK)
+#  error "libFLAC is not compatible with V110_SDK71. See github.com/xiph/flac/issues/72#issuecomment-934777777 for details"
+#else
+#if defined(_USING_V110_SDK71_)
+/* This code is when V110_SDK71_ is defined */
+/* WATCHOUT: Make sure to test any compile using this thoroughly
+ * on a machine running XP! This hack only fixes libFLAC, not the
+ * command line tools. It is recommended to use MinGW instead to
+ * target Windows XP */
+static int flac_fstat(int fd, struct __stat64 *buffer) {
+	FLAC__int64 filelength = _filelengthi64 (fd);
+	if(filelength < 0)
+		return -1;
+	buffer->st_size = filelength;
+	return 0;
+}
+#else
 #define flac_fstat _fstat64
+#endif
+#endif
 #else
 #define flac_stat_s stat /* stat struct */
 #define flac_fstat fstat
--- a/src/libFLAC/stream_decoder.c
+++ b/src/libFLAC/stream_decoder.c
@@ -3645,7 +3645,13 @@ FLAC__StreamDecoderLengthStatus file_length_callback_(const FLAC__StreamDecoder 
 
 	if(decoder->private_->file == stdin)
 		return FLAC__STREAM_DECODER_LENGTH_STATUS_UNSUPPORTED;
-	else if(flac_fstat(fileno(decoder->private_->file), &filestats) != 0)
+
+#ifndef FLAC__USE_FILELENGTHI64
+	if(flac_fstat(fileno(decoder->private_->file), &filestats) != 0)
+#else
+	filestats.st_size = _filelengthi64(fileno(decoder->private_->file));
+	if(filestats.st_size < 0)
+#endif
 		return FLAC__STREAM_DECODER_LENGTH_STATUS_ERROR;
 	else {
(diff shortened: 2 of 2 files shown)


ktmf01:
@sudo-rm-rf-oops: [Here](https://github.com/xiph/flac/issues/72#issuecomment-934777777) you suggested poisoning builds with Visual Studio for Windows XP.
> you probably want to somehow 'poison' flac_fstat in Windows builds to stop people from using it by mistake. If you care about XP, that is.

Do you think this solution is adequate? It breaks building and refers to that post for instruction on how to fix it.

sudo-rm-rf-oops:
Hi Martijn,

Hmmm.  Could you not apply the fix I suggested to stream_decoder.c?  In the code I'm currently using, `flac_fstat` isn't used anywhere else.  I do still support XP and I'd rather not have to edit the sources if I don't have to.

As for poisoning `flac_fstat`, maybe share/compat.h could read something like:
```
#ifdef _WIN32
#define flac_stat_s NOT_SUPPORTED
#define flac_fstat NOT_SUPPORTED
#else
#define flac_stat_s stat /* stat struct */
#define flac_fstat fstat
#endif
```
and then people will know the score.

Alternatively, if you're now using `flac_fstat` more widely, and you're using it just to get the length of the file (can't think what else you would want it for), then the obvious solution would be to factor that out into (say) a `flac_getfilesize` function and localise the `_WIN32` hacks in there.  That might be a wise investment in any case.

But if all this is a pain, just to support (in all likelihood) one old-fashioned guy, then fine, I'll go along with what you're suggesting.

ktmf01:
Well, there are a couple of things that make me feel this is the best solution:
1. Microsoft themselves is unwilling to fix this
1. MinGW creates builds that work just fine with WinXP
1. Starting with the next release, I intend to supply official DLLs (issue #369 and #113) build with MinGW which will work fine on WinXP
1. The problem also seems to affect flac_stat, not just flac_fstat
1. I don't have a testcase to test this with

Especially the lack of a testcase makes this difficult to implement, and especially the lack of a fix from Microsoft makes me hesitant to do so.

sudo-rm-rf-oops:
OK, I'll do a few tests and get back to you.

sudo-rm-rf-oops:
Hello again.  I have now refreshed my memory and you're right, MS aren't going to fix this problem.  And `flac_stat` is indeed also affected (I tested it) although that never gets called for my use-case,  so perhaps you could tell me what it's actually used for (I mean, I can see it in the source, but when I try to put a breakpoint at the various call sites I find that it's all dead code.  And I do read and write metadata).

So how about a compromise, something like this:
```
#  if defined (_USING_V110_SDK71_) && !defined (MSVC_XP_HACK)
#    error "libFLAC is not compatible with VS 2015 or later when building for XP.  To work around this, define MSVC_XP_HACK in your build environment but please be aware that this is not officially supported and you do so at your own risk."
#  endif
```
and then use `#ifdef MSVC_XP_HACK` in stream_decoder.c to enable my original fix.

Best of both worlds?  What do you think?

ktmf01:
@sudo-rm-rf-oops I still don't feel this is the right thing to do, but I've added in a hack, a bit different then you requested though. Could you check whether this works for you?

flac_stat is used in the `flac` command line program. For that, I've added the comment that this hack is only for libFLAC.

sudo-rm-rf-oops:
Hi Martijn,

That won't work unfortunately because, at that point in the code, various declarations needed by that function are not available yet.  Also, by placing the body of the function in a header file, you're going (potentially) to wind up with multiple copies.

So the function needs to be _declared_ there and the function body placed somewhere else.  `windows_unicode_filenames.c` might be a suitable place.  It also shouldn't be declared `static`.

Also, the error message could be more helpful, maybe something like:
```
#  error "libFLAC is not compatible with V110_SDK71. To make it compile, define V110_SDK71_HACK and see the comments in windows_unicode_filenames.c"
```
I don't think there's any point referencing my original bug report - that is history with these changes and you can close it.

Some other points:

  - I would start out by zeroing the buffer in `flac_fstat` so that the caller is not faced with potentially uninitialised data members
  - I would comment its limitations (i.e. it only returns the file size)

But I did test the code, after making the necessary changes, and it does work.  And thanks for keeping me in the loop on this, I do appreciate it.

ktmf01:
> That won't work unfortunately because, at that point in the code, various declarations needed by that function are not available yet. Also, by placing the body of the function in a header file, you're going (potentially) to wind up with multiple copies.

It compiles just fine here. Where do you experience problems?

> So the function needs to be _declared_ there and the function body placed somewhere else. `windows_unicode_filenames.c` might be a suitable place. It also shouldn't be declared `static`.

I really like to keep specific compatibility stuff like this in compat.h. Why do you feel a static function in a header is insufficient? Could you elaborate?

sudo-rm-rf-oops:
I shoehorned your changes into 1.3.2 for expediency.  That presumably explains why I got different results to you.

But putting the body of that function into `compat.h` is a really bad idea.  `compat.h` should be free of any dependencies - it just #defines any platform-specific stuff, basically - and as such you can #include it anywhere without having to worry about whether it relies on any other header files being included first.  You are now breaking that model, and I don't think you should.  It might work today, but it might break tomorrow.

Also, there is a clear precedent for putting Windows-specific code into `windows_unicode_filenames.c`.  That is, in fact, it's raison d'etre.  You might prefer putting `flac_fstat` into some other .c file - that would be fine - but not in `compat.h` please.

And please don't forget the issue of duplicate function bodies, when doing things your way.  The compiler / linker probably strips them, but I wouldn't want to rely on that when it can easily be avoided.  Furthermore, clang (and I believe gcc) will issue a warning for a `static` function that is defined in a translation unit but never used, and I'm sure people would rather not have to deal with the noise.

ktmf01:
> But putting the body of that function into `compat.h` is a really bad idea. `compat.h` should be free of any dependencies - it just #defines any platform-specific stuff, basically - and as such you can #include it anywhere without having to worry about whether it relies on any other header files being included first. You are now breaking that model, and I don't think you should. It might work today, but it might break tomorrow.

This hack is bound to break anyway, because it is not tested by CI and probably used by very few users. That's why a manual preprocessor define is needed to activate this code, and why there is a warning to check the result on WinXP.

I agree these dependencies shouldn't be there, but I see this as a stub for a hack that a dev should work out and test. I really don't want code so specifically to a certain compiler + sdk combination anywhere else.

> Also, there is a clear precedent for putting Windows-specific code into `windows_unicode_filenames.c`. That is, in fact, it's raison d'etre. You might prefer putting `flac_fstat` into some other .c file - that would be fine - but not in `compat.h` please.

That file is for unicode filenames on windows, for all compilers. This is something specific in my view not to Windows, but to a certain SDK used by Visual Studio of which the runtime component on Windows XP is broken. It also has nothing to do with unicode, nor with filenames.

> And please don't forget the issue of duplicate function bodies, when doing things your way. The compiler / linker probably strips them, but I wouldn't want to rely on that when it can easily be avoided. Furthermore, clang (and I believe gcc) will issue a warning for a `static` function that is defined in a translation unit but never used, and I'm sure people would rather not have to deal with the noise.

gcc nor clang will ever see this code, because this is a visual studio specific hack.

sudo-rm-rf-oops:
Sorry, still don't like it.  Developers generally avoid implementing function bodies in header files (except for functions they want the compiler to inline) and `compat.h` is absolutely the worst place to do it, for reasons already given .  If you really don't want to put it in `windows_unicode_filenames.c` (and I can see why) then let's just create a new home for it (and any future Windows-specific code can go in there).

> gcc nor clang will ever see this code, because this is a visual studio specific hack.

Good point!  Sorry about that.

ktmf01:
I can't think of a solution that would be acceptable for us both, so I'll revert this PR back to just adding a conditional `#error` (so no code), but with better explanation as to why FLAC is not compatible with this SDK.

ktmf01:
> Well. Frankly, I think that's the worst possible option.

I think the worst possible option is doing nothing, letting people compile something that they think works on Windows XP, but doesn't.

> I really don't understand the objection to adding `vs_xp_hacks.c` or whatever to the repository

Similarly I don't understand the objection to using a `static` function that is only used once anyway.

I'm trying to clean up, and I rather not have some file with just one hack for just one SDK because Microsoft doesn't fix this for whatever reason. See [here](https://web.archive.org/web/20170327195018/https://connect.microsoft.com/VisualStudio/feedback/details/1557168/wstat64-returns-1-on-xp-always), [here](https://social.msdn.microsoft.com/Forums/vstudio/en-US/52b0c797-6fa7-4933-8bb2-fe90e8764e27/visual-c-2015-express-stat-not-working-on-windows-xp?forum=vcgeneral) and [here](https://developercommunity.visualstudio.com/t/-stat64-windows-xp/320354), this problem has been known about for at least 7 years but as of yet no fix.

Furthermore, I don't know what else might be broken. Therefore I think it a good thing to encourage developers that want Windows XP compatibility to **either**
1. Use platform toolset v141_xp with extra care and be well aware that proper functionality on recent Windows might not translate to Windows XP, and weird, unexpected behaviour might occur on Windows XP only, **or**
2. Use the MinGW-compiled DLL that will be supplied from the next release onward or use MinGW themselves to compile libFLAC

Therefore, I'd like to encourage all devs targetting Windows XP to seriously consider option 2, or be prepared to do some extra testing.

sudo-rm-rf-oops:
I really do think you're overthinking this Martijn.  As I understand it, your concern is that you cannot include this in your automated testing, but that's fine: just put a disclaimer in `vs_xp_hacks.c` and move on.  And I can assure you that (apart from the command line tool), nothing else is broken.

As for the rest, I've tried my best to explain why `compat.h` is the wrong place for this.  If you can't see it, please talk to a colleague.  Also, I don't want to use your DLL (because I use an unusual set of build options) and I very much don't want to use MinGW.

ktmf01:
> I really do think you're overthinking this Martijn. As I understand it, your concern is that you cannot include this in your automated testing, but that's fine: just put a disclaimer in `vs_xp_hacks.c` and move on. And I can assure you that (apart from the command line tool), nothing else is broken.

Do you use the metadata interfaces? They depend on _wstat64 and it seems to me they are probably broken too.

> As for the rest, I've tried my best to explain why `compat.h` is the wrong place for this. If you can't see it, please talk to a colleague, get a fresh perspective. Also, I don't want to use your DLL (because I use an unusual set of build options) and I very much don't want to use MinGW.

I think you're misunderstanding how the FLAC project is run. I have no colleagues, I work on this in my own time, for free, voluntarily.

Also, I know of a few projects where having function definitions in header files are commonplace. See for example FFmpeg [here](https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/put_golomb.h), [here](https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/put_bits.h), [here](https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/lpc.h) and [here](https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/wavpack.h). Another example that is less comparable to FLAC is the linux kernel, for example [here](https://github.com/torvalds/linux/blob/master/include/net/ipv6.h)

> Edit: One thing I do agree with, actually, is that it makes sense for you to **advise** people to use your DLL or MinGW if that's what you're routinely testing. I have no argument with that. But, it wouldn't be my choice and it shouldn't be forced on people.

I don't feel like I'm forcing anyone to not use Visual Studio. In fact, the FLAC project went to great length recently to improve the MSVC user experience by moving to CMake, enabling configuration of the build, which wasn't possible previously. The only thing is that I'm adding an `#error` that must be manually removed by anyone wanting to compile libFLAC with Visual Studio with the problematic toolset. You could call that forcing in a sense that I'm forcing people to take a step extra (removing that `#error`) but I feel that is necessary, because just adding some remark here or there is going to be missed by many.

include/share/compat.h
```diff
@@ -138,6 +138,36 @@
 #    define PRId64 "I64d"
 #    define PRIx64 "I64x"
 #  endif
+#  ifdef _USING_V110_SDK71_
+#    error "libFLAC is not compatible with V110_SDK71. See comments in include/share/compat.h for details"
```
ktmf01, line 142:
```suggestion
#    pragma message("WARNING: This compile will NOT FUNCTION PROPERLY on Windows XP. See comments in include/share/compat.h for details")
```
Apparently this is also possible

sudo-rm-rf-oops:
Yes it is, but that's not the whole story.  People like me (if there are any people like me) will also have to edit `stream_decoder.c` for every new release, and that's what I'm objecting to.  Maybe just check in my suggested change (probably using a more suitable `#ifdef`).

I'll reply to your other comment later if that suggestion won't fly with you, but just to say I do use the metadata interfaces in `libflac` and they work fine (and they most definitely do not call `_wstat64` - I checked), so I don't know what's going on there.  If you want to know what interfaces I use in my code, just ask.

ktmf01:
I've further narrowed down this problem to using the `/MT` switch. I assume you use this in your project?

In other words: when the resulting compile depends on Microsoft Visual C++ 2015 Redistributable, everything seems to work fine.

sudo-rm-rf-oops:
> I've further narrowed down this problem to using the /MT switch. I assume you use this in your project?

I do (and have good reasons to)

ktmf01:
I've checked all other uses of _fstat64 and _wstat64 and it seems in all places where they are used there is some kind of fallback or use is optional, except for stream_decoder.c. In most places, those functions are used to copy file creation/modification dates on encoding or changing metadata. It is also used to display compression ratio when encoding with `flac`, but this issue is purely cosmetic.

As said, I don't know how to test this, so I must simply assume the suggested fix is correct (which I don't like obviously).

Still, I'll keep the warning around, that seems appropriate.

sudo-rm-rf-oops:
OK, great.  Thanks very much for doing this Martijn, I appreciate it.  I don't mind the warning.

I should be able to test `stream_decoder.c` for you.  If you can let me know how / when I can download a complete source tree I'll give it a go.  Also, if you'd like to fill me in on why you can't test this (maybe you don't have an installation of XP available or something), maybe I can make some suggestions.

ktmf01:
The top of this page says:
> [ktmf01](https://github.com/ktmf01) wants to merge 5 commits into [xiph:master](https://github.com/xiph/flac) from [ktmf01:winxp-sdk](https://github.com/ktmf01/flac/tree/winxp-sdk)

You can find the source tree at [ktmf01:winxp-sdk](https://github.com/ktmf01/flac/tree/winxp-sdk). Be aware that building requires configuration with CMake though.

Anyway, considering testing: the problem wasn't the lack of access to a Windows XP system (I still have one running in a VM) but rather the lack of a clear test case: I didn't know which program used that part of the API. Using `flac` or `metaflac` didn't result in any problems, but `flac --skip` doesn't work because that depends on _stat64, which is broken too.

I found a testcase though, with a small patch (similar to this one) for test_libFLAC.exe, that can be run. It fails without this patch and passes with it. These lines needed to be patched for the tests to work:

https://github.com/xiph/flac/blob/05ff5692660fb9c7907a1fa62b7def88e34a17be/src/test_libs_common/file_utils_flac.c#L145-L152

I can now say that with this patch, libFLAC works on Windows XP compiled with said specific configuration, but `flac` and `metaflac` have some broken/missing features.

sudo-rm-rf-oops:
OK thanks (sorry, I'm a bit of a GitHub n00b), I'll take a look.  I'll see if I can get it to build with [a modification of] my existing VS 2022 project file, make sure there aren't going to be any nasty surprises when the next formal release goes out.  How close are you to a making a release, would you say?

As for your automated tests, that sounds like a useful patch to roll into the official sources.  Seems to me, from my 20,000 foot view, that if that file wasn't testing whether the call `flac_stat` succeeded, it should do.

Are you worried about the slightly broken state of `flac` and `metaflac`?  I'm not, because I don't use them, and maybe nobody wants to build them from source with VS + `/MT` and run them on XP anyway.  Maybe you just need to find a sensible place to tell people what the state of the play is.

sudo-rm-rf-oops:
Hello again Martijn,

I have now built and tested the source (libFLAC only), and all seems well bar one minor issue which I'll come to in a moment.  My test conditions:

  - build in VS 2022 (with VS 2017 compiler, perforce) and v141_xp toolset
  - /MT
  - tested on Windows 11 and XP, no problems found

My XP machine is an ancient Pentium II (deliberately) and, because of my particular build settings, libFLAC still works correctly, despite the limited instruction set supported by that CPU.

So, the only problem I encountered when building my app (not libFLAC itself) was this line in `include/FLAC/format.h`:
```
typedef struct {
...
} FLAC__StreamMetadata;
```

This used to read:
```
typedef struct FLAC__StreamMetadata {
...
} FLAC__StreamMetadata;
```

and that has the advantage that you can forward declare `struct FLAC__StreamMetadata;` (and not include `format.h`) in compilation units that only need to pass pointers around.  So it would be nice if that could be reinstated so that I don't have to change my code (and I imagine I'm not the only one who would appreciate this).

Please let me know if you'd like me to open a new issue for this and I'll be happy to do so.  None of us like 'mission creep'.

ktmf01:
> How close are you to a making a release, would you say?

I hope a month or two, depends on what issues might still come up.

> As for your automated tests, that sounds like a useful patch to roll into the official sources. Seems to me, from my 20,000 foot view, that if that file wasn't testing whether `flac_stat` works, it should do.

The problem was the other way round actually: without the patch, the test would always fail, so I couldn't see the difference with and without this PR. With the mentioned patch, flac_stat was not called and thus the test would only fail without this PR. So, I'll leave it in, it did its job.

> Maybe you just need to find a sensible place to tell people what the state of the play is.

I think the warning pointing to the explanation in include/share/compat.h should be enough. The warning should only show up with MSVC, with the problematic toolset, when build with `/MT`.

> I have now built and tested the source (libFLAC only), and all seems well bar one minor issue

Thanks for testing.

> This used to read:
> 
> ```
> typedef struct FLAC__StreamMetadata {
> ...
> } FLAC__StreamMetadata;
> ```

Well, according to git blame it was never like that. See https://github.com/ktmf01/flac/blame/master/include/FLAC/format.h#L537 It says that is has been `typedef struct {` since the start of the CVS (which was migrated to git later) 22 years ago.

> Please let me know if you'd like me to open a new issue for this and I'll be happy to do so. None of us like 'mission creep'.

Yes, please open a new issue. That way, this PR and #72 can be closed. Otherwise this might be forgotten about.

sudo-rm-rf-oops:
> Well, according to git blame it was never like that.

Oh, OK.  I must have tweaked my local copy then, sorry about that.

> Yes, please open a new issue.

OK, will do.  Please go ahead and close whatever you want to close, no need to wait for me.

PR #410, opened 2022-08-04
Remove HAVE_GETAUXVAL check for FreeBSD

pkubaj:
FreeBSD doesn't have getauxval().

### DIFF
--- a/src/libFLAC/cpu.c
+++ b/src/libFLAC/cpu.c
@@ -251,7 +251,7 @@ ppc_cpu_info (FLAC__CPUInfo *info)
 	} else if (getauxval(AT_HWCAP2) & PPC_FEATURE2_ARCH_2_07) {
 		info->ppc.arch_2_07 = true;
 	}
-#elif defined(__FreeBSD__) && defined(HAVE_GETAUXVAL)
+#elif defined(__FreeBSD__) && defined(HAVE_SYS_AUXV_H)
 	unsigned long hwcaps;
 	elf_aux_info(AT_HWCAP2, &hwcaps, sizeof(hwcaps));
 	if (hwcaps & PPC_FEATURE2_ARCH_3_00) {


src/libFLAC/cpu.c
```diff
@@ -251,7 +251,7 @@ ppc_cpu_info (FLAC__CPUInfo *info)
 	} else if (getauxval(AT_HWCAP2) & PPC_FEATURE2_ARCH_2_07) {
 		info->ppc.arch_2_07 = true;
 	}
-#elif defined(__FreeBSD__) && defined(HAVE_GETAUXVAL)
+#elif defined(__FreeBSD__)
```
ktmf01, line 254: Shouldn't this be the following?

```suggestion
#elif defined(__FreeBSD__) && defined(HAVE_SYS_AUXV_H)
```

pkubaj: Corrected.
However, elf_aux_info() is available since FreeBSD 12.0. Branches before 12 are already unsupported and frankly, so much happened in the powerpc* land in FreeBSD 13.0 that no one should use anything before 13.0.

ktmf01: The thing is, that HAVE_SYS_AUXV_H check was added for uclibc here: https://github.com/xiph/flac/pull/292 I made the mistake to suggest to check for HAVE_GETAUXVAL, I see now that HAVE_SYS_AUXV_H must be checked.

I really don't know anything about uclibc nor FreeBSD, it seems uclibc is only compatible with linux, but perhaps other libc's have something similar, omitting sys/auxv.h on FreeBSD as well? In that case, this check would make sense even on FreeBSD 12 and later?

pkubaj: I don't think there is any other libc available on FreeBSD for powerpc*. There actually are GNU/kFreeBSD distros (FreeBSD kernel with GNU userland), but since they use GNU, they probably use glibc (and all of them are only for amd64 / i386 anyway). I don't think anyone has ever attempted to run uclibc on FreeBSD and the probability of that happening on powerpc* is even lower.

I think you can just assume that FreeBSD 12.0 and newer have sys/auxv.h.

ktmf01: Do you think we can leave it in though, just in case?

pkubaj: Sure you can, it doesn't hurt.

PR #414, opened 2022-08-13
add option metaflac --show-all-tags

harridu:
Signed-off-by: Harald Dunkel <harri@afaics.de>
I would like to suggest to add an option **--show-all-tags** to metaflac, listing all tags found in the header. The functionality is already in the code. This patch adds just a command line option to use it. 

Sample:
```
% metaflac --show-all-tags Kraftwerk/Autobahn/01_Autobahn.flac 
ARTIST=Kraftwerk
TRACKNUMBER=01
ALBUM=Autobahn
TITLE=Autobahn
GENRE=Electronic
DATE=1974
```


ktmf01:
Is there any difference to `metaflac --export-tags-to=-` or was that option unknown to you?

harridu:
I heard about this before, but it is not intuitive to use and hard to find in the man page. There are options --remove-tag and --remove-all-tags, but next to --show-tag there is no --show-all-tags.

ktmf01:
I'd rather not duplicate functionality. I see two ways to do this.
1. drop most of this patch, and simply add text to the man page for option `--show-tag` suggesting to use `--export-tags-to=-` to show all tags
2. minimize this patch by making `--show-all-tags` an alias for `--export-tags-to=-`, implementing this alias in options.c

ktmf01:
@harridu Are you still planning to do anything with this PR?

ktmf01:
I added a commit simplifying this PR.

Note to self: I still need to add a test for `--export-tags-to=-` and `--show-all-tags`, but that would conflict with #504, so I should probably wait with that until that one is merged.

PR #431, opened 2022-09-09
Trivial fixes in cmake files for 1.4.0

kloczek:
- install cmake module in default cmake search path
- Fix man pages lvl 1 install path


ktmf01:
Thanks for reporting and sending in a fix. It took me a while to figure out what happens here.

For the first one: CMake was installing the man pages in the wrong directory, right? The autotools build tools do this by themselves, but for CMake, it must be specified. Is hard-coding man1 a cross-platform solution?

For the second one, I don't know anything about these cmake modules, so can't comment really. Could you perhaps explain?

kloczek:
> For the first one: CMake was installing the man pages in the wrong directory, right?

Correct.

> Thanks for reporting and sending in a fix. It took me a while to figure out what happens here.

> For the first one: CMake was installing the man pages in the wrong directory, right? The autotools build tools do this by themselves, but for CMake, it must be specified. Is hard-coding man1 a cross-platform solution?

>For the second one, I don't know anything about these cmake modules, so can't comment really. Could you perhaps explain?

If Those .cmake files will be installed in that direcrory to use those modules all what needs to be done in some other project CMakeLists.txt would be just `include(<module_name>)`. Without that it will be bneccessry to pass additional directory in which `flac` ckame modules are installed.

ktmf01:
> If Those .cmake files will be installed in that direcrory to use those modules all what needs to be done in some other project CMakeLists.txt would be just `include(<module_name>)`. Without that it will be bneccessry to pass additional directory in which `flac` ckame modules are installed.

Thanks. What I still don't understand: apparently these files were being installed in the wrong directory. Could it be the original implementer had different intentions, or was it plain wrong? I can't find anything saying that `${CMAKE_INSTALL_LIBDIR}/cmake/` is the CMake module directory [as mentioned here](https://cmake.org/cmake/help/latest/command/include.html).

kloczek:
To be able use `include(<module_name>)` in the `${CMAKE_INSTALL_LIBDIR}/cmake/<module_name>` needs to be installed file like `<module_name>-config.cmake` which in this case is fulfilled so all what was necessaty to trim was base directory inm which all those files needs to be installed.

ktmf01:
I've added another related fix to this PR, because otherwise it would conflict, as it touches the same bit of code. I hope you don't mind.

kloczek:
Just suggestion: pandoc is written in ocaml.
I would suggest to have source of the man page in xml/docbook and use xsltproc to generate roff output.
xml/docbook can ba easyly wrapped to make translations using gettext.

ktmf01:
The source of the man page used to be in docbook, but I converted to markdown so they are easily human readable. I'm happy with the output of pandoc as it is now, so I don't see a reason to move back to docbook.

Issue #433, opened 2022-09-10
Wrong printf format specifier in stream_decoder.c


manxorist:
Building 1.4.0 with VS2022, our CI found the following warning:
```
src\libFLAC\stream_decoder.c(2201,22): warning C4477: 'fprintf' : format string '%lu' requires an argument of type 'unsigned long', but variadic argument 1 has type 'FLAC__uint64'
```

<https://github.com/xiph/flac/blob/2ab4677750138640ba3f5099f6158ede3d011ad5/src/libFLAC/stream_decoder.c#L2201>
should be
```
fprintf(stderr, "Bitreader was %" PRIu64 " bytes short\n", current_decode_position - decoder->private_->last_seen_framesync);
```

PR #434, opened 2022-09-10
Fix wrong format string specifier.

manxorist:
Fixes https://github.com/xiph/flac/issues/433


ktmf01:
Thanks for reporting and sending in a fix!

Issue #437, opened 2022-09-12
Unit tests using CMake fails on FreeBSD 13


diizzyy:
Tests libFLAC, libFLAC++ and flac fails
This is also seems incorrect as amd64/x86-64 is little endian, ```++ libFLAC unit test: endswap (big endian host)``` and GNU Autotools gets this right.

The specific test that libFLAC units fails passes using GNU Autotools

I've attached a log of tests (I've disabled streams as it takes a long time to run)
[flac-ctest-log.zip](https://github.com/xiph/flac/files/9545983/flac-ctest-log.zip)

OS: FreeBSD 13.1-STABLE (amd64)
FLAC: 1.4.0 Release
CMake: 3.23.3

ktmf01:
Could you also supply a log of the CMake configuration? Furthermore, what happens if you build for release instead of debug? There is an issue with running the tests on a debug buil, see #436

diizzyy:
Hi,

Build type is Release,

Arguments passed to CMake:

[...]

PR #442, opened 2022-09-13
[CMake] Prepend compiler flags, fix Clang compiler warnings

ktmf01:
Because of issue https://github.com/xiph/flac/issues/437 compiler flags are prepended instead of appended, so the user can override them without altering the CMakeLists.txt.

Also, Clang doesn't support per-function optimize options, so in CMake the fma file gets that option per-file. This is not supported by automake.


diizzyy:
Looks good to me and works fine on my testbox! :-)

PR #451, opened 2022-09-15
stream_decoder.c: move sys/stat.h include after sys/types.h.

sezero:
It is common practice that sys/types.h is usually included before
sys/stat.h, see e.g. stat/fstat man page.  Without this, some old
systems might emit warnings.  (I remember running into it quite a
while ago, can't really remember which environment though..)


ktmf01:
Huh, this code had been untouched for 15 years. Must be really old systems I guess?

PR #456, opened 2022-09-19
CMake disable TestBigEndian use with Emscripten

robUx4:
The test always fails:

```
CMake Error at /usr/share/cmake-3.18/Modules/TestBigEndian.cmake:50 (message):
  no suitable type found
Call Stack (most recent call first):
  CMakeLists.txt:71 (test_big_endian)
```

But it's OK it only supports little-endian:
https://emscripten.org/docs/porting/guidelines/portability_guidelines.html#code-that-cannot-be-compiled

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -105,7 +105,9 @@ include(CheckCXXSourceCompiles)
 include(CheckLibraryExists)
 include(GNUInstallDirs)
 include(UseSystemExtensions)
-include(TestBigEndian)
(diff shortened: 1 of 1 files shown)


robUx4:
It may not be needed after all. emscripten comes with a `emcmake` tool to set the proper environment when calling cmake, that will fix the endian test and a few other things. We don't use it in VLC, hence the issue, but we're working on it https://code.videolan.org/videolan/vlc/-/merge_requests/2576.

PR #457, opened 2022-09-19
Add Universal Windows Platform support

robUx4:
We have been maintaining similar patches in VLC for a long time. It would be nice to have them merged directly here.


ktmf01:
It would seem to me this break compilation when WINAPI_FAMILY_PARTITION is not defined. Is there a way to detect that WINAPI_FAMILY_PARTITION cannot be used? Than we can default to the current behaviour.

Something like `#if !HAVE_WINAPI_FAMILY || WINAPI_FAMILY_PARTITION` or similar

ktmf01:
This addition broke compilation on original MinGW (not MinGW-w64). I've added a commit that should fix this. Please let me know what you think.

robUx4:
The proper way would be to include `<winapifamily.h>` it will do all the work for us. Nowdays it's automatically included with `<windows.h>` that why the CI is happy.

There might be some Windows SDK and mingw-w64 or mingw versions where it's not the case. It's likely they don't have `winapifamily.h` or that noone builds UWP with these anyway. I checked the latest win32api of mingw and it doesn't have this header. So your patch should be enough to cover the old SDKs and won't affect the new ones.

PR #461, opened 2022-09-22
gitlab ci: Install pandoc dep

rillian:
This is needed to generate the manpages for the `distcheck` target.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -10,7 +10,7 @@ autotools:
     - apt-get update &&
       apt-get install -y libogg-dev
         libtool-bin gettext zip
-        doxygen graphviz
+        doxygen graphviz pandoc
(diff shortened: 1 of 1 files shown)


ktmf01:
Ah, yes, had to do that for github as well. I keep forgetting about the gitlab CI, sorry about that.

ktmf01:
I added `-DINSTALL_MANPAGES=OFF` for the CMake items in GitHub CI. This is also fine, I don't have a preference

ktmf01 (approved):

rillian:
> I added `-DINSTALL_MANPAGES=OFF` for the CMake items in GitHub CI. This is also fine, I don't have a preference

Thanks! That's a reasonable idea, but I mostly wanted to turn the ci report green, so I'm ok with this for now.

PR #462, opened 2022-09-22
Remove XMMS plugin

ktmf01:
XMMS hasn't had a release in almost 15 years now.


rnsanchez:
> XMMS hasn't had a release in almost 15 years now.

Hi!  Indeed.
But I have been using XMMS daily, for more years than that now (I'd say since 2000).  If this is not a maintenance burden towards flac, could you please consider reverting?  On the bright side, XMMS (1.x at least) is as solid as anything will get.

Thank you for your time.

ktmf01:
Please further explain why you feel this is necessary. If the software itself is not being updated, what is wrong with using an old FLAC plugin?

rnsanchez:
Good point.

Well I've noticed this after a regular system upgrade, which updated flac from 1.3.4 to 1.4.2; that's where I've noticed something unexpected with XMMS.  While XMMS itself is not being maintained for that many years, many of the components that XMMS uses, are.  Libraries, X.org (XFree86 back then).

I would guess that, as long as libxmms-flac builds fine and doesn't experience a major API breakage, it could be acceptable to be kept around.  Though XMMS won't change its use of flac in any way, it still directly and indirectly benefits from any fixes and improvements on flac's side, as long as they don't reflect on the APIs used by the plugin.  If anything, a testament of time and engineering: "we're really stable API-wise, XMMS plugin still works fine, you can hear for yourself".

But on the other hand, I understand that this might be unwanted for the project, maybe it is not a trivial component, tech debt, etc, and the userbase on XMMS' side is minuscule (myself and who else).

ktmf01:
The problem I had with this code is that I didn't know that it even worked. Almost 5000 lines of code with no tests, no fuzzing and no regular builds from it as far as I knew.

rnsanchez:
For the offline parts (regular files on disk), considering normal usage, they indeed work and have worked for longer than I remember (I don't think I ever had an issue with libxmms-flac).  As for the network components in the plugin, I don't know.  While I have occasionally used network streams in XMMS, I'm sure those were not flac streams.

But I can understand if you see this as an undesired liability.

ktmf01:
One other problem here is that I feel this plugin doesn't belong in this codebase. I feel it should actually be considered a separate project really. There are hundreds of FLAC players, but no player other than xmms has code specific to it in the FLAC codebase.

If someone wants to keep this alive, I think it is best to grab the source code as it was before the removal (from one of the many past releases of FLAC) and maintain it in a separate repository.

rnsanchez:
I must agree with your reasoning.  libxmms-flac could move on, to an external repository of its own.

Let us acknowlegde XMMS plugin as being present in the repository since its first commit,  bb7f6b99d02e15e0ff98ff1d9d037098dc5dc4e9.

Thank you again for your time, and of course for considering the revert (which is now clearly unnecessary).  :-)

PR #467, opened 2022-09-25
Speed up set_partitioned_rice_

ktmf01:
set_partitioned_rice_ used some code to quickly calculate log2, but there is already log2-calculating code elsewhere.

Results of performance testing are below. Tests were done on a Raspbery Pi 4 and a Xeon E-2224G, results are percent of speedup. While this code is faster on Aarch64, it isn't on x86_64. I need to do some more research on why this is happening.

Source - preset | Aarch64 | x86_64
:---------------|:--------|:-------
16-bit -5       |  < 1%   |  -0.5%
16-bit -8       |   3%    |   -4%
16-bit -8ep     |   9%    |   -20%
24-bit -5       |   1%    |   0%
24-bit -8       |   1%    |   -3%
24-bit -8ep     |   5%    |   -11%


ktmf01:
@lvqcl: you might find this interesting, the code you contributed in commit 64f34e6 is faster than using a single CPU instruction (bit scan reverse, bsr). I'm baffled that apparently this single instruction is so slow it is outperformed. Perhaps branch prediction plays a role here? Usually the result of this bit of code doesn't change very rapidly. In which properly predicted branching is faster than the bsr instruction?

lvqcl:
I have no idea actually... but:

What compiler did you use? What asm code does it generate?

ktmf01:
x86_64 tests were done with GCC 11.2 (shipped with Ubuntu 22.04), Aarch64 tests with GCC 10.2 (shipped with Debian Bullseye), so nothing out of the ordinary.

lvqcl:
¯\\\_(ツ)\_/¯

ktmf01:
Just found out: the problem isn't that `bsr` is slow, it is the integer division. Didn't know that was slow. Should be easy to work around that though.

ktmf01:
New results. x86_64 is faster, but aarch64 is slower.

Source - preset | Aarch64 | x86_64
:---------------|:--------|:-------
16-bit -5       |  +0.6%  |  +0.4%
16-bit -8       |  +2.1%  |  +2.6%
16-bit -8ep     |  +3.5%  |  +5.5%
24-bit -5       |  +0%    |  +0%
24-bit -8       |  +1.3%  |  +1.9%
24-bit -8ep     |  +5.8%  |  +4.2%

ktmf01:
Note to self: one problem with this code is that it does not take into account that the first partition has a size different.

ktmf01:
I was under the impression the code presented here would only work fast on 64-bit archs (as the code was differentiated for 32-bit and 64-bit archs) but it seems many use cases in fact get faster. I did some comparing 32-bit compiles on a Xeon E2224G machine.

For 16-bit PCM, setting -8 speeds up by 0.5%, setting -8ep **slows down by 1.2%** and setting -8epr8 speeds up by 5.0%
For 24-bit PCM, setting -8 speeds up by 1.0%, setting -8ep speeds up by 2.4% and setting -8epr8 speeds up by 8.0%.

So, I'm dropping the 32-bit specific code, as it was pretty hard to read.

PR #469, opened 2022-09-26
Assume Clang supports x86 intrinsics up to FMA

ktmf01:
This mirrors behaviour for GCC

### DIFF
--- a/src/libFLAC/include/private/cpu.h
+++ b/src/libFLAC/include/private/cpu.h
@@ -82,28 +82,14 @@
 #elif defined __clang__ && __has_attribute(__target__) /* clang */
   #define FLAC__SSE_TARGET(x) __attribute__ ((__target__ (x)))
   #define FLAC__FAST_MATH_TARGET(x) __attribute__ ((__target__ (x)))
-  #if __has_builtin(__builtin_ia32_maxps)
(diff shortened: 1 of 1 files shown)


sezero:
If this is about http://lists.xiph.org/pipermail/flac-dev/2022-September/006588.html
the specific issue seems to be with `__builtin_ia32_pabs*`:
https://lists.freebsd.org/archives/freebsd-current/2022-September/002633.html
https://reviews.llvm.org/D117791
https://github.com/llvm/llvm-project/commit/e5147f82e1cb

Why not just add additional builtin checks for the replacement one, instead?

ktmf01:
For the following reasons
- None of the other compilers check for availability of functions
- All x86 ISA extensions up to FMA have been supported in LLVM/Clang since 3.1, released 10 years ago
- `__has_attribute(__target__)` has only been true for Clang since Clang 3.7, after support all x86 ISA extensions up to FMA was introduced

Unless there is something weird going on and Clang has selectively left out support for these in certain releases, or support can be left out of a compile, I don't see a reason to add the checks back in, only for them to bitrot again at some point.

sezero:
I see, OK then.

ktmf01:
Tested: https://lists.xiph.org/pipermail/flac-dev/2022-September/006596.html

PR #472, opened 2022-10-06
Reset last_frame_is_set on flush

### LINKED ISSUE #471: Change in stream decoder behavior in 1.4.0
1793632ee6988deb933ff7551fee92134622b558 changes how libFLAC behaves when it encounters missing frames. 

Suppose an application implemented seek by flushing the decoder and then passing frames from the new (post-seek) position to the decoder immediately. In this case, because the libFLAC internal seek function isn't being used, libFLAC will (after the linked commit) treat the frames in between the starting position and seek position as missing and insert silence.

Gstreamer implements seeking in [exactly this way](https://gitlab.freedesktop.org/gstreamer/gstreamer/-/issues/1475). (This unfortunately can't easily be changed due to how their code is modularized. They have seeking implemented on the file level for some formats like Matroska, as opposed to seeking the FLAC stream itself.)

This may or may not be a regression in 1.4.0 depending on whether you treat applications doing seeking without using a callback as legitimate. I believe the gstreamer folks plan to work around this in the next point release, so in this case no action would be needed to fix libFLAC. However, I think this should probably be documented as a [breaking change](https://www.xiph.org/flac/api/group__porting__1__3__4__to__1__4__0.html) for other libFLAC users.

ktmf01:
This should fix #471

### DIFF
--- a/src/libFLAC/stream_decoder.c
+++ b/src/libFLAC/stream_decoder.c
@@ -894,6 +894,7 @@ FLAC_API FLAC__bool FLAC__stream_decoder_flush(FLAC__StreamDecoder *decoder)
 	decoder->private_->samples_decoded = 0;
 	decoder->private_->do_md5_checking = false;
 	decoder->private_->last_seen_framesync = 0;
+	decoder->private_->last_frame_is_set = false;
 
 #if FLAC__HAS_OGG
 	if(decoder->private_->is_ogg)


ktmf01:
@afontenot Did I understand correctly you already verified this fix?

afontenot:
@ktmf01 Yes, this change fixes the issue for me.

tp-m:
Thanks! Do you have plans for a bug fix release in the near future? :)

ktmf01:
Depends on what you call near future. In FLAC terms, yes, near future. Between 1.2.1 and 1.3.4 (6 releases) were 15 years, I plan to release this in less then a month from now. I would like to get a few more bugfixes and perhaps a few performance improvements in.

tp-m:
Ok, thanks. I'm asking because this will affect FLAC playback with all GStreamer-based music players on standard Linux desktops that ship the new 1.4.x version, so I guess we'll just point distros to this MR then.

PR #474, opened 2022-10-07
do not use CPPFLAGS as global overring criteria

philippe44:
When one wants to add some compiler/preprocessor flags, as soon as any of CFLAGS/CXXFLAGS and CPPFAGS are user-defined, the configure.ac ignores all built-in definitions. 

This is unfortunate as it means that no "addition" can be made, user needs to redefine the full CFLAGS/CPPFLAGS. 

Now, CPPFLAGS are not defined internally in configure.ac, so this PR proposes to not include a definition of CPPFLAGS in overriding criteria, so that user can keep built-in compile flags and *add* its own preprocessor options. This is particularly useful for cross compiling where the CC/CXX compilers need some specific flags (e.g. setting the architecture) that are not in configure.ac and putting them in CPPFLAGS then voids flags like -O3

Of course, it changes behavior for users that used CPPFLAGS previously to overwrite all CFLAGS/CXXFLAGS and I realize that might breach some compatibility, but it seems more logical to proceed this way.


ktmf01:
What is the way most projects deal with this? I see ogg prepends its flags to CFLAGS. Perhaps flac should do the same?

philippe44:
I can't say I'm an authority on that topic :smile:. It seems to me that most projects prepend/append their mandatory non-negotiable flags to CFLAGS/CXXFLAGS and then have optional FLAGS as well that can be overwritten when user sets the environment variables/configure command line. But they let you play with CPPFLAGS, and that seems to me the best rules

1- set what you really don't want to be removed in "core" CFLAGS/CXXFLAGS
2- set "optional" flags maybe like -O3 in CFLAGS/CXXFLAGS. Such "optional" flags will be ignored if user sets them in environment/configure
3-don't use CPPFLAGS for your own configuration and don't use them to ignore your "optional" FLAGS values, instead just add CPPFLAGS to the build command.

Feels like it provides best flexibility

ktmf01:
I would prefer the approach of simply prepending flags. Flags like `-O3 -funroll-loops` can simply be overriden when the user supplies `CFLAGS='-O2'` when prepending, because GCC and Clang always take the last of such flags to have precedence.

This is also what automake does with `AM_CFLAGS`, so they are overriden by `CFLAGS` when necessary.

Anyway, I'll take a look at what other similar projects do. I'll have to be documented in the changelog, obviously.

philippe44:
That's obviously your call, but I liked the idea of having flags like `-03 -funroll-loops` set automatically by the configurator and when I want to just add flags (e.g. `-march=xxx` for cross-compiling), I use CPPFLAGS (it's for C and C++ anyway). Otherwise, when I want to just add my architecture flag, I would have to mimic what is set by the configuration `-03 -funroll-loops`  *and* add mine.

ktmf01:
@philippe44 What do you think of #480 instead of this one?

philippe44:
It works for me as well. but I'm not sure how to unset -funroll-loops in gcc. I think it also implies a series of other optimizations and I don't know either what happens if -funroll-loops is unset. I don't think -O2 implies resetting these.

ktmf01:
You can use `-fno-funroll-loops`. According to docs it implies `-frerun-cse-after-loop` which is part of -O2 already.

PR #480, opened 2022-10-12
Prepend to CFLAGS instead of only setting when blank

ktmf01:
I prefer this over #474.


sezero (approved):
This looks good to me.

ktmf01:
Thanks!

PR #483, opened 2022-10-16
fix errors: 'for' loop initial declarations are only allowed in C99 mode

sezero:
Fixes build if c99 mode isn't enabled.

### DIFF
--- a/src/flac/decode.c
+++ b/src/flac/decode.c
@@ -334,11 +334,12 @@ FLAC__bool DecoderSession_init_decoder(DecoderSession *decoder_session, const ch
 	if (decoder_session->replaygain.spec.apply || !decoder_session->channel_map_none)
 		FLAC__stream_decoder_set_metadata_respond(decoder_session->decoder, FLAC__METADATA_TYPE_VORBIS_COMMENT);
 
-	if(!decoder_session->analysis_mode && !decoder_session->test_only && decoder_session->foreign_metadata == NULL)
+	if(!decoder_session->analysis_mode && !decoder_session->test_only && decoder_session->foreign_metadata == NULL) {
 		/* Warn user if foreign metadata is found */
-		for(uint32_t i = 0; i < FLAC__FOREIGN_METADATA_NUMBER_OF_RECOGNIZED_APPLICATION_IDS; i++)
+		uint32_t i;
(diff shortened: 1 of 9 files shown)


ktmf01:
I'd rather use in-loop declarations, makes debugging stuff easier quite often. For what compilers/platforms is this necessary?

sezero:
Well, that depends: I am on an old CentOS 6.10 setup (yes I know that it's EOL), it comes with gcc4.4 by default which actually _can_support in-loop declarations _if_ c99 mode is enabled, and unlike newish gcc, gcc4.4 (for e.g.) doesn't enable c99 mode by default. This is where the autoconf version used to generate the configure script comes into play: only autoconf 2.70 and newer enable c99 or c11 or whatever newest standart the compiler supports behind the user's back (a behavior which I actually do not like), so there is the problem.

ktmf01:
Okay, so a choice has to be made here to either push this PR or force C99 behaviour. Correct?

sezero:
> Okay, so a choice has to be made here to either push this PR or force C99 behaviour. Correct?

Yes

ktmf01:
Any I assume you tested that with this PR, flac is otherwise C89 compatible?

I'm unsure what to do here. Obviously C99 support is pretty standard nowadays for targets running any OS, but it might not be for embedded applications. However, flac has had a few bugs in the past where a loop variable was used outside of the loop, which is easy to catch (as in: won't compile) when that variable has the loop as its scope.

sezero:
> Any I assume you tested that with this PR, flac is otherwise C89 compatible?

Yes -- with a subset of c99 available anway, such as stdint.h. Tested with e.g. down to gcc-3.4.6

> 
> I'm unsure what to do here. Obviously C99 support is pretty standard nowadays for targets running any OS, but it might not be for embedded applications. However, flac has had a few bugs in the past where a loop variable was used outside of the loop, which is easy to catch (as in: won't compile) when that variable has the loop as its scope.

Your call, obviously.

sezero:
> > Any I assume you tested that with this PR, flac is otherwise C89 compatible?
> 
> Yes -- with a subset of c99 available anway, such as stdint.h. Tested with e.g. down to gcc-3.4.6

... And gcc older than 3.4 choke on `-Wextra`: If I change it to `-W` instead, even gcc2.95
seems to succeed building the source tree.

PR #484, opened 2022-10-16
fix --disable-stack-smash-protection mingw issues, add --disable-fortify-source

sezero:
The `AC_SEARCH_LIBS` checks for ssp leaves `-lssp` in `LIBS` even if one
specifies `--disable-stack-smash-protection` on the configure cmdline. As
a result, dll generation would fail if mingw toolchain is built as static
only, e.g. https://debbugs.gnu.org/cgi/bugreport.cgi?bug=45738

This patch adds `-lssp` only if stack smash protection is enabled.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -342,14 +342,16 @@ AC_ARG_ENABLE([stack-smash-protection],
         [enable_stack_smash_protection=yes],[enable_stack_smash_protection=no])])
 
 AC_ARG_ENABLE([fortify-source],
-    [AS_HELP_STRING([--disable-fortify-source],[Disable __FORTIFY_SOURCE buffer overflow protection])],,
+    [AS_HELP_STRING([--disable-fortify-source],[Disable _FORTIFY_SOURCE buffer overflow protection])],,
     [AS_IF([test "$ac_cv_c_compiler_gnu" = "yes"],
         [enable_fortify_source=yes],[enable_fortify_source=no])])
 
 case "$host" in
 	*mingw*)
-		if test "$enable_fortify_source" = "yes" || test "$enable_stack_smash_protection" = "yes"; then
+		if test "$enable_fortify_source" = "yes"; then
 			AC_SEARCH_LIBS(__memset_chk, ssp, , mingw_has_memset_chk=no)
+		fi
+		if test "$enable_stack_smash_protection" = "yes"; then
 			AC_SEARCH_LIBS(__stack_chk_fail, ssp, , mingw_has_stack_chk_fail=no)
 		fi
 	;;
--- a/configure.ac
+++ b/configure.ac
@@ -236,8 +236,10 @@ os_is_windows=no
 case "$host" in
 	*mingw*)
 		os_is_windows=yes
+		old_LIBS=$LIBS
 		AC_SEARCH_LIBS(__memset_chk, ssp, , mingw_has_memset_chk=no)
 		AC_SEARCH_LIBS(__stack_chk_fail, ssp, , mingw_has_stack_chk_fail=no)
+		LIBS=$old_LIBS
 		;;
 esac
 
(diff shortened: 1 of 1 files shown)


ktmf01:
The problem here is that libssp also supplies stuff for _FORTIFY_SOURCE. I can do static linking just fine on MinGW running under Windows, but I've had the same problem as you link to on Debian. I think it is some problem with that specific MinGW version that Debian has.

sezero:
For the record, I'm not running on debian :)

ktmf01:
Sorry, assumed the 'deb' had something to do with debian, my bad.

sezero:
That 'deb' in 'debbugs' is a mystery of the universe :)
OTOH, the mingw toolchain I use is my own build.

ktmf01:
Still, this works just fine at my end, and libssp is also needed for _FORTIFY_SOURCE. So as this PR is, it isn't a proper fix.

sezero:
OK, will try changing things, making it possibly like the cmake side where fortify is an option (which is not in autotools.)

ktmf01:
I still think this is a libtool bug, not a FLAC bug.

sezero:
> I still think this is a libtool bug, not a FLAC bug.

Yes, that's correct. And IMO, since you are using libtool, you should (and can) work around that.

How about something simpler like the following? (May need correction, especially wording, etc.)
If it's good, I can commit and force-push to this PR.
```diff
diff --git a/configure.ac b/configure.ac
index 5117a0b..c813954 100644
--- a/configure.ac
+++ b/configure.ac
@@ -236,8 +236,10 @@ os_is_windows=no
 case "$host" in
 	*mingw*)
 		os_is_windows=yes
+		old_LIBS=$LIBS
 		AC_SEARCH_LIBS(__memset_chk, ssp, , mingw_has_memset_chk=no)
 		AC_SEARCH_LIBS(__stack_chk_fail, ssp, , mingw_has_stack_chk_fail=no)
+		LIBS=$old_LIBS
 		;;
 esac
 
@@ -343,6 +345,19 @@ AC_ARG_ENABLE([stack-smash-protection],
     [AS_IF([test "$ac_cv_c_compiler_gnu" = "yes" && test "x$mingw_has_stack_chk_fail" != "xno"],
         [enable_stack_smash_protection=yes],[enable_stack_smash_protection=no])])
 
+AC_ARG_ENABLE([fortify-source],
+    [AS_HELP_STRING([--disable-fortify-source],[Disable __FORTIFY_SOURCE buffer overflow protection])],,
+    [AS_IF([test "$ac_cv_c_compiler_gnu" = "yes" && test "x$mingw_has_memset_chk" != "xno"],
+        [enable_fortify_source=yes],[enable_fortify_source=no])])
+
+case "$host" in
+	*mingw*)
+		if test enable_fortify_source=yes || enable_stack_smash_protection=yes; then
+			LIBS="$LIBS -lssp"
+			fi
+	;;
+esac
+
 AC_ARG_ENABLE(64-bit-words,
 	AS_HELP_STRING([--enable-64-bit-words],[Set FLAC__BYTES_PER_WORD to 8 (4 is the default)]))
 if test "x$enable_64_bit_words" = xyes ; then
```

sezero:
Revised patch force-pushed: please review.

configure.ac
```diff
@@ -236,8 +236,10 @@ os_is_windows=no
 case "$host" in
 	*mingw*)
 		os_is_windows=yes
+		old_LIBS=$LIBS
 		AC_SEARCH_LIBS(__memset_chk, ssp, , mingw_has_memset_chk=no)
 		AC_SEARCH_LIBS(__stack_chk_fail, ssp, , mingw_has_stack_chk_fail=no)
```
ktmf01, line 241: It would be great if these those lines can somehow be moved downwards to the next
```
case "$host" in
	*mingw*)
```
If that can be done, this old_LIBS workaround is not necessary, and if MinGW at some point decides to move libssp to libgcc, this will not break down.

sezero: > It would be great if these those lines can somehow be moved downwards to the next

AFAICS that's hard to do, because `stack-smash-protection` and `fortify-source` options
rely on results from that case...esac being not set to no. And the later `-lssp` addition
relies on either `stack-smash-protection` or `fortify-source` option being set to yes. So
double checking seems necessary to me unless I'm missing something. If you have
something better, can you please show me?

> If that can be done, this old_LIBS workaround is not necessary, and if MinGW at some point decides to move libssp to libgcc, this will not break down.

sezero:
> What do you think of this addition?

Well, yeah. And certainly solves my specific problem.

It seems to rely on `__memset_chk` and `__stack_chk_fail` being either
both available or unavailable (as far as I can see), but this is OK for
me. Thanks.

PR #485, opened 2022-10-16
fix broken bswap.m4:

sezero:
it used to always define HAVE_BSWAP32 and HAVE_BSWAP16 even if the builtin isn't supported.

### DIFF
--- a/m4/bswap.m4
+++ b/m4/bswap.m4
@@ -39,16 +39,10 @@ AC_DEFUN([XIPH_C_BSWAP32],
 [AC_CACHE_CHECK(for bswap32 intrinsic,
   ac_cv_c_bswap32,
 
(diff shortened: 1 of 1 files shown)


sezero:
The existing version used to work prior to 1.4.0. It broke with https://github.com/xiph/flac/commit/bfe5ff9455ca62d7827e14a22fbb1522a63c4e3e which changed `#if` checks to `#ifdef`, but seems to have missed this.

ktmf01:
Thanks! Just out of curiosity, which compilers do not have these intrinsics?

sezero:
I was messing with older gcc - e.g.: gcc3.4 doesn't have them. Not what everyone would use, surely, but having the source tree correct is more important IMO.

PR #489, opened 2022-10-18
build: Fix 32-bit Visual Studio builds

fanc999:
Hi,

This attempts to fix the CMake build files so that 32-bit Visual Studio builds can be done successfully when buliding the NASM bits, namely by restricting the application of `/wd4996` and `/wd4267` when building C and C++ files only.

With blessings, thank you.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -161,8 +161,8 @@ if(HAVE_WERROR_FLAG)
 endif()
 
 add_compile_options(
-    $<$<BOOL:${MSVC}>:/wd4267>
-    $<$<BOOL:${MSVC}>:/wd4996>
+    $<$<AND:$<COMPILE_LANGUAGE:C>,$<BOOL:${MSVC}>>:/wd4267>
+    $<$<AND:$<COMPILE_LANGUAGE:CXX>,$<BOOL:${MSVC}>>:/wd4996>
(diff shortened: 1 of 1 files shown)


fanc999:
(Ah, didn't realize assembly code support was removed, but I thought we should still limit these flags to C and C++)

ktmf01:
There are no other languages in use anymore, so this doesn't serve any purpose anymore, right?

I was actually planning on removing such generator expressions, now assembler has been dropped.

fanc999:
Hi @ktmf01,

Yeah, I realized that after cloning the repo... if you prefer, I think you can just close this PR.

ktmf01:
Yeah, sorry about that. I don't like turning down a PR, especially from first-time contributor :)

Anyway, how did building go in general? The removal of the MSVC-specific build system in favor of CMake is rather new for FLAC. Did you experience any problems besides those specific to nasm?

fanc999:
Hi,

Thanks though, but sometimes plans don't catch up with changes...oh well :)

I have a few issues on ARM64/aarch64 Visual Studio builds, so I will try to put things there together--the Visual Studio builds otherwise work fine for me.

ktmf01:
That would be great. I do have ARM64 hardware to test on, but nothing running Windows/MSVC, so that is currently very much untested.

PR #521, opened 2022-11-23
libFLAC.m4, libFLAC++.m4: fix -Wstrict-prototypes

thesamesam:
Signed-off-by: Sam James <sam@gentoo.org>

### DIFF
--- a/src/libFLAC++/libFLAC++.m4
+++ b/src/libFLAC++/libFLAC++.m4
@@ -51,19 +51,16 @@ dnl
 dnl Now check if the installed libFLAC++ is sufficiently new.
 dnl
       rm -f conf.libFLAC++test
-      AC_TRY_RUN([
+      AC_RUN_IFELSE([AC_LANG_PROGRAM([[
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include <FLAC++/decoder.h>
-
-int main (void)
-{
+]],[[
   system("touch conf.libFLAC++test");
   return 0;
-}
+]])],[],[no_libFLACPP=yes],[echo $ac_n "cross compiling; assumed OK... $ac_c"])
 
-],, no_libFLACPP=yes,[echo $ac_n "cross compiling; assumed OK... $ac_c"])
        CFLAGS="$ac_save_CFLAGS"
        CXXFLAGS="$ac_save_CXXFLAGS"
        LIBS="$ac_save_LIBS"
--- a/src/libFLAC++/libFLAC++.m4
+++ b/src/libFLAC++/libFLAC++.m4
@@ -57,7 +57,7 @@ dnl
 #include <string.h>
 #include <FLAC++/decoder.h>
 
-int main ()
+int main (void)
 {
   system("touch conf.libFLAC++test");
   return 0;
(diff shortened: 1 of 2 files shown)


ktmf01:
Thanks!

`AC_TRY_RUN` is kinda obsolete though. Perhaps better to replace it with `AC_RUN_IFELSE` combined with `AC_LANG_PROGRAM`? That way this file is rid of the main function altogether. I guess issues like this one are exactly why `AC_LANG_PROGRAM` was made.

Then we get something like this

```
AC_LINK_IFELSE([AC_LANG_PROGRAM([[
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <FLAC/format.h>
]],[[
  system("touch conf.libFLACtest");
  return 0;
]])],[],[no_libFLAC=yes],[echo $ac_n "cross compiling; assumed OK... $ac_c"])
```

Any thoughts?

thesamesam:
Sure, I've not been doing it proactively because there's _tonnes_ of these (https://lwn.net/Articles/913505/ etc) (also because I don't want someone to reject the fix if they don't want churn), but I'm happy to do it when asked!

Let me give it a whack.

ktmf01:
Interesting read. This change is only a small one (in some m4 files the project itself doesn't use) but FLAC defines it's own bool, true and false, so I guess that is a more impactful change. However, there are people around here that would like to keep using pre-C99 compilers (see https://github.com/xiph/flac/pull/483) so I guess a little bit of code needs to be added that chooses between stdbool.h and the current solution. Also, FLAC__bool is used extensively at the C API interface.

I'm not familiar with the implications of such a change (especially for API/ABI compatibility), so if you have done this before, could you please advise?

ktmf01:
Looking into the bool/true/false stuff a little more, I think there's nothing API/ABI related here that needs to change. I think there is no issue with `typedef int FLAC__bool;`. This bit would need to change I think

https://github.com/xiph/flac/blob/7270930e9e269838a5f133f130a96881488ea6e8/include/FLAC/ordinals.h#L74-L83

Perhaps that whole file can be simplified, removing support for really old compilers. MSVC versions older than 2010 have been unsupported for 4 years or more.

```
#ifndef FLAC__ORDINALS_H
#define FLAC__ORDINALS_H

#include <stdint.h>
#include <stdbool.h>

typedef int8_t FLAC__int8;
typedef uint8_t FLAC__uint8;

typedef int16_t FLAC__int16;
typedef int32_t FLAC__int32;
typedef int64_t FLAC__int64;
typedef uint16_t FLAC__uint16;
typedef uint32_t FLAC__uint32;
typedef uint64_t FLAC__uint64;

typedef int FLAC__bool;
typedef FLAC__uint8 FLAC__byte;

#endif
````

Any thoughts? @sezero as you suggested #483, would this work for that particular case, or have there been GCC versions that had stdint.h but not stdbool.h? [Reading this](https://gcc.gnu.org/c99status.html) it seems stdbool.h is properly present from GCC 3 and up.

sezero:
> Any thoughts? @sezero as you suggested #483, would this work for that particular case, or have there been GCC versions that had stdint.h but not stdbool.h? [Reading this](https://gcc.gnu.org/c99status.html) it seems stdbool.h is properly present from GCC 3 and up.

That's fine by me.

ktmf01:
Opened PR #524 for the second issue

thesamesam:
Thanks, sorry, I've been away the last few days.

PR #534, opened 2023-01-07
Remove all PPC-specific code (and a mention of SPARC)

ktmf01:
As it turns out that the PPC-specific code provides no benefit with modern, autovectorizing compilers, and given that there is a lot of cruft surrounding the PPC-specific parts, remove all of it.

This follows https://lists.xiph.org/pipermail/flac-dev/2022-December/006620.html


ktmf01:
Results of CI on PPC can be found here: https://app.travis-ci.com/github/ktmf01/flac/builds/259472012

ktmf01:
I've received test results from someone with access to a POWER9 machine running Fedora. A plain compile was compared with a compile configured with `--disable-asm-optimizations`. No difference in execution speed was measured.

PR #551, opened 2023-02-21
Add fuzzer for flac command-line tool

ktmf01:
As the `flac` tool has quite some code of its own, add a fuzzer for that. Of course, this also indirectly fuzzes libFLAC, though coverage is limited compared to the existing fuzzers


ktmf01:
I'm closing this, probably better to split into two PRs

PR #556, opened 2023-02-23
Improve x86 intrinsics functions for fixed predictors

ktmf01:
Profiling revealed that this intrinsics function is about 50% slower than plain C on my Intel Xeon E-2224G.


ktmf01:
Performance testing on my Intel Kaby Lake-R (Core i5 7200U) shows the same: depending on the preset, this PR gives 5 to 10% speedup.

From top to bottom presets -0, -1, -2, -3 and -4 with MD5 calculation turned off.
[set of 24bit tracks - fixed-noasm.pdf](https://github.com/xiph/flac/files/10818379/set.of.24bit.tracks.-.fixed-noasm.pdf)

ktmf01:
Performance testing of the newly added functions on my Intel Kaby Lake-R (Core i5 7200U). From top to bottom presets -0, -1, -2, -3 and -4 with MD5 calculation turned off.

[set of 24bit tracks - fixed-newasm.pdf](https://github.com/xiph/flac/files/10832258/set.of.24bit.tracks.-.fixed-newasm.pdf)
[set of 32bit tracks - fixed-newasm.pdf](https://github.com/xiph/flac/files/10832313/set.of.32bit.tracks.-.fixed-newasm.pdf)

As I see the results, 24-bit gets a ~5% speed boost (over the speed boost of dropping the old intrinsics) and 32-bit gets a ~10% speed boost. The reason the absolute speeds for the 32-bit tracks is higher, is because the sample rates of the tracks in that test set are generally lower.

PR #559, opened 2023-03-01
Speed up format input

ktmf01:
While profiling the faster presets of libFLAC with `flac`, it seemed the function format_input was taking a disproportionate amount of time. With a few changes, this function became much faster, especially for 24-bit and 32-bit inputs.

The most extreme case I've seen was a 26x speedup for mono 32-bit PCM files. For stereo 24-bit files, this function became twice as fast.

### DIFF
--- a/src/flac/encode.c
+++ b/src/flac/encode.c
@@ -35,6 +35,7 @@
 #include "share/compat.h"
 #include "share/private.h"
 #include "share/safe_str.h"
+#include "share/endswap.h"
 #include "encode.h"
 
 #ifdef min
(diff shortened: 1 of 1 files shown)


ktmf01:
For 16-bit PCM, a tiny improvement is visible in the following graphs.
[very large set of tracks - fasterfixed.pdf](https://github.com/xiph/flac/files/10895705/very.large.set.of.tracks.-.fasterfixed.pdf)

ktmf01:
For 24-bit input, a slightly larger improvement is visible
[set of 24bit tracks - fasterfixed.pdf](https://github.com/xiph/flac/files/10896344/set.of.24bit.tracks.-.fasterfixed.pdf)

PR #564, opened 2023-03-08
Speed up bitreader

ktmf01:
This is in some way a partial revert of #347: a new function is added accelerated with the BMI2 instruction set. However, contrary to the situation before #347, there is no separate code involved: the code for reading rice symbols is simply duplicated and compiled once with and once without permitting use of BMI2 instructions.

Also, the default word size is switched to 64-bit. This gives a respectable increase in encoding speed (about 7%) and enables a 4% larger decoding speed improvement for the BMI2 specific function than with 32-bit words.


chocolate42:
Funny timing, I was about to make a thread asking how much effort has been put into optimising the bitreader? So much of a decoders time is spent reading bits it makes sense that it'd already be optimised to hell and back, but if not (maybe the initial solution got optimised but a better algorithm was never tried?) then it seems like a good target to optimise further. In the next weeks I plan on benchmarking to try and beat a libflac x86_64 build.

ktmf01:
I've tried several times, but every time it turns out to be very hard to optimize. The code is very branchy and the branches are (seemingly) hard to predict for CPUs. As you can see, the only improvements this PR makes is enabling use of a x86 specific instructions set extension, and switching to 64-bit words as the default. This last change has been 'on hold' for several years now, but I think the time is right to make the switch.

Anyway, preparing this PR I've tried a few things, like writing a less branchy alternative for the core bitreading part, but execution was slower instead of faster. I've also tried to add some compiler hints about branching, but that only lead to slower execution. In fact, even trying GCCs `-fprofile-generate` and `-fprofile-use` make this code slower instead of faster. I think this is because GCC accepts 'code bloat' by default (and this is helpful for the bitreader) but aggressively cuts that bloat when profiling information is available. So, branch prediction might get better, but the code itself gets slower nonetheless.

So, yes, this is rather difficult. Also, I know from benchmarks done by other FLAC decoder implementations, that libFLACs bitreader is reasonably fast, and that there probably isn't a lot of room for improvement.

ktmf01:
Still, there could very well be a much faster approach. See #550, a different approach for the bitwriter resulted in quite a large speedup. However, bitwriting is easier than bitreading, because on writing, all information is known beforehand. I've tried a similar approach for bitreading, didn't work.

Here's a breakdown of the decoding process as I profiled it with this PR applied. The file I used was a classical music piece compressed with -8ep
- Rice bitreading is ~38%
- Calculation of LPC predictor is ~28%
- MD5 sum calculation is ~24%
- Generic parsing and stereo decorrelation is ~6% (this is function read_frame_)
- CRC16 calculation is ~3%
- What remains is ~1%

The same piece compressed with -0 gives the following breakdown
- Rice bitreading is ~50%
- MD5 sum calculation is ~30%
- Calculation of fixed predictor is ~8%
- Generic parsing is ~6% (-0 has no stereo decorrelation)
- CRC16 calculation is ~4%
- What is remains is ~2%

So yes, bitreading is clearly the most taxing part of decoding

ktmf01:
So no low-hanging fruit it seems. I agree that it is a nice challenge, working on the bitreader, as it combines a lot of aspects of programming and the outcome is quite unpredictable.

PR #566, opened 2023-03-10
Add missing *local_fixed_compute_best_predictor_wide member

harishdm:
It seems like (*local_fixed_compute_best_predictor_wide) was unintentionally removed from FLAC__StreamEncoderPrivate in
the commit d24b45515eed253ab372435c481e01e14ea126ac. This CL adds it back.


ktmf01:
Thanks. The integer-only variant of the library is not regularly tested so these problems tend to creep in.

That said, I assume you use this integer-only variant of the library, or you wouldn't have found this problem. If it is not too much hassle, could you perhaps tell something about your use-case? As maintainer, I often make decisions based on the most common use case because I am simply unaware of use cases such as this one.

harishdm:
Thank you for taking a look at this.
Android fork of this library uses the integer only variant. https://android.googlesource.com/platform/external/flac/+/refs/heads/master

I found this when I was trying to update it to the latest commit.

PR #576, opened 2023-03-28
Retune compression levels

ktmf01:
The compression levels in FLAC have been kept the same for a long time, only preset 6, 7 and 8 got a retune 10 years ago. However, as some parts of FLAC got much faster, I felt it was time to retune the presets.

The presets as they currently are mostly add complexity as the level goes up, increasing encoding time and improving compression more or less monotonically.

The presets as this PR proposes them are more like the approach TAK (Tom's lossless Audio Kompressor) has to presets: grouping them based on decoding speed. Preset 0, 1 and 2 are the fastest decoding, presets 3, 4 and 5 are a little less fast and presets 6, 7 and 8 are the slowest (but still pretty fast!).

Within those groups, the first of the presets (0, 3 and 6) encode the fastest, the second of the presets (1, 4 and 7) compress a little better, and the last of the presets (2, 5 and 8) encode the best/slowest.

The advantage of this approach is that presets are more clearly differentiated. Currently, I find it kind of hard justifying the use of presets 0 (because 1 is almost as fast but compressed much better), 3 (same reason) and 6 (same reason).

All changes
- presets 0, 1 and 2 go from blocksize 1152 to blocksize 4096
- preset 0 switches from no stereo decorrelation to 'loose mid-side'
- preset 1 increases its max rice partition order from 3 to 5
- preset 2 increases its max rice partition order from 3 to 6, and does exhaustive predictor order search
- preset 3 switches from no stereo decorrelation to 'loose mid-side', increases its max predictor order to 8 and lowers its max rice partition order to 3
- preset 4 takes on the options from former preset 5
- preset 5 takes on the options from former preset 6
- preset 6 increases its max predictor order from 8 to 32, lowers its max rice partition order from 6 to 3 and switches apodization from subdivide_tukey(2) to tukey(5e-1)
- preset 7 increases its max predictor order from 12 to 32, lowers its max rice partition order from 6 to 5 and switches apodization from subdivide_tukey(2) to tukey(5e-1)
- preset 8 increases its max predictor order from 12 to 32. increases its max rice partition order from 6 to 8 and switches apodization from subdivide_tukey(3) to subdivide_tukey(2)
- for presets 6, 7 and 8, the max predictor order (32) is limited to 12 when encoding material with a sample rate of 48kHz or lower, to stay within subset limits

This probably seems rather confusing. Hopefully some graphs help. 

Here is data for 24-bit tracks. Presets 0 through 8 go from top left to bottom right.
[set of 24bit tracks.pdf](https://github.com/xiph/flac/files/11089432/set.of.24bit.tracks.pdf)

Compression seems to increase monotonically, however, this is not always the case. Here are the results split by track. Page 4 shows a very clear case where preset 6 gives a file larger than preset 4 and preset 7 gives a file larger than preset 5. On average, this retune improves compression however.
[set of 24bit tracks-per-track.pdf](https://github.com/xiph/flac/files/11089443/set.of.24bit.tracks-per-track.pdf)

Here is data for 16-bit tracks. 
[very large set of tracks.pdf](https://github.com/xiph/flac/files/11089490/very.large.set.of.tracks.pdf)

Also per track:
[very large set of tracks-per-track.pdf](https://github.com/xiph/flac/files/11089540/very.large.set.of.tracks-per-track.pdf)

At the time of writing this, the PR doesn't yet update the documentation.

### DIFF
--- a/src/libFLAC/deduplication/lpc_compute_and_use_predictor_above_12.c
+++ b/src/libFLAC/deduplication/lpc_compute_and_use_predictor_above_12.c
@@ -1,0 +1,136 @@
+/* This code is ordered such that each loop is at least 2/3ths unrolled */
+
+if(order <= 22) {
+	if(order <= 16) {
+		for(i = 0; i < (int)data_len; i++) {
+			sum = 0;
+			switch(order) {
+				case 16: sum += qlp_coeff[15] * (FLAC__LPC_DATA_TYPE)data[i-16]; /* Falls through. */
+				case 15: sum += qlp_coeff[14] * (FLAC__LPC_DATA_TYPE)data[i-15]; /* Falls through. */
+				case 14: sum += qlp_coeff[13] * (FLAC__LPC_DATA_TYPE)data[i-14]; /* Falls through. */
+				case 13: sum += qlp_coeff[12] * (FLAC__LPC_DATA_TYPE)data[i-13];
+						 sum += qlp_coeff[11] * (FLAC__LPC_DATA_TYPE)data[i-12];
+						 sum += qlp_coeff[10] * (FLAC__LPC_DATA_TYPE)data[i-11];
+						 sum += qlp_coeff[ 9] * (FLAC__LPC_DATA_TYPE)data[i-10];
+						 sum += qlp_coeff[ 8] * (FLAC__LPC_DATA_TYPE)data[i- 9];
+						 sum += qlp_coeff[ 7] * (FLAC__LPC_DATA_TYPE)data[i- 8];
+						 sum += qlp_coeff[ 6] * (FLAC__LPC_DATA_TYPE)data[i- 7];
+						 sum += qlp_coeff[ 5] * (FLAC__LPC_DATA_TYPE)data[i- 6];
+						 sum += qlp_coeff[ 4] * (FLAC__LPC_DATA_TYPE)data[i- 5];
+						 sum += qlp_coeff[ 3] * (FLAC__LPC_DATA_TYPE)data[i- 4];
+						 sum += qlp_coeff[ 2] * (FLAC__LPC_DATA_TYPE)data[i- 3];
+						 sum += qlp_coeff[ 1] * (FLAC__LPC_DATA_TYPE)data[i- 2];
+						 sum += qlp_coeff[ 0] * (FLAC__LPC_DATA_TYPE)data[i- 1];
+			}
+			FLAC__LPC_ACTION;
+		}
+	}
+	else { /* 16 < order <= 20 */
+		for(i = 0; i < (int)data_len; i++) {
+			sum = 0;
+			switch(order) {
+				case 22: sum += qlp_coeff[21] * (FLAC__LPC_DATA_TYPE)data[i-22]; /* Falls through. */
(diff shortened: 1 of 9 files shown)


dwillmore:
Looks good to me.  I am only concerned with compression ratio, though.  I only ever use --best and encode 16 bit files, so my concern has limited scope.

kleinesfilmroellchen:
Looks good; I'm especially liking even stronger prefixes for the highest levels since I mostly use those.

regorxxx:
As noted on HA, this PR breaks the fundamentals of FLAC having the same decoded speed for all encoding presets (at least in theory).

So I would suggest to to split this into 2 PRs, one with lineal compression improvements and other with the tuned presets touching the decoding time and presets grouping. 

The second part clearly changes one of the most known features of FLAC (official encoder) and should require a separate discussion since it would affect almost every implementation out there. (And should go along a major FLAC revision)

ktmf01:
@regorxxx I fail to see how these changes affect other FLAC implementations. Please elaborate.

regorxxx:
I'm just stating what the docs say, which in turn was reflected on wikipedia and further testing, and the way other software and users have been using flac for years. I am not saying FLAC's design should stick to the web info out there, but changing a well known design requires more thought.

https://xiph.org/flac/faq.html#general__what_is
> Why do the encoder settings have a big effect on the encoding time but not the decoding time?
> 
> It's hard to explain without going into the codec design, but to oversimplify, the encoder is looking for functions that approximate the signal. Higher settings make the encoder search more to find better approximations. The functions are themselves encoded in the FLAC file. Decoding only requires computing the one chosen function, and the complexity of the function is very stable. This is by design, to make decoding easier, and is one of the things that makes FLAC easy to implement in hardware.

https://en.wikipedia.org/wiki/FLAC#Encoding_and_decoding

> FLAC is optimized for decoding speed at the expense of encoding speed. A benchmark has shown that, while there is little variation in decoding speed as compression level increases, beyond the default compression level 5, the encoding process takes up considerably more time with little space saved compared to level 5.[[17]](https://en.wikipedia.org/wiki/FLAC#cite_note-17)

**Your changes further deviate from that statement, making the default presets decoding time even more different.** It's pretty clear what I'm saying... there is a somewhat popular assumption that decoding time should be aprox. the same no matter the preset you use. Part of this PR -further- breaks that assumption, which is baked in a lot of software/hardware and user's workflow. (like which presets are better for DAPs considering battery usage)

**And that's why I ask to split these changes into 2 PRs**. Whether the doc statements are right or not, that's another point. It's obvious the encoding parameters change the decoding time, but at least the default presets should stick to that assumption as much as possible. If you want to change that, that's fine, but then a lot of guides, docs and things must be changed too (and this PR omits that).

ktmf01:
I think you should look at these statements in comparison with other codecs. See this: http://audiograaf.nl/losslesstest/revision%205/Average%20of%20all%20CDDA%20sources.pdf

As you can see, the spread in decoding speed for WavPack is 100% (2x), for Monkey's Audio and ALS about 500% (5x), and for OptimFROG about 1000% (10x). For FLAC it is about 8% and this PR will push it to 15%, by mostly making is faster.

So, I still think that the statements you mention are valid with this PR. 15% spread is 'not a big effect' and 'little variation' to me.

PR #578, opened 2023-03-29
getopt: Fix the build with clang-16

orbea:
Clang-16 no longer allows these legacy declarations.


orbea:
Seems this was half fixed in PR https://github.com/xiph/flac/pull/455, but not all of the instances were found.

PR #581, opened 2023-04-03
Fixed compilation for Windows UWP

dmitrykos:
Fixed compilation for Windows UWP ~~by making `get_utf8_argv()` unimplemented in case of UWP target~~. 

LoadLibrary is unavailable on UWP. ~~The proposal is to leave `get_utf8_argv()` unimplemented because argv functionality is not needed for UWP anyway. Alternative solution would be to return 1 by `get_utf8_argv()` in case of UWP.~~

_[update]_ Changed implementation to actually make `get_utf8_argv()` working in UWP environment. The arguments are obtained  as per MS [example](https://learn.microsoft.com/en-us/windows/uwp/launch-resume/console-uwp).

### DIFF
--- a/src/share/win_utf8_io/win_utf8_io.c
+++ b/src/share/win_utf8_io/win_utf8_io.c
@@ -39,9 +39,12 @@
 
 #define UTF8_BUFFER_SIZE 32768
 
-#if !defined(WINAPI_FAMILY_PARTITION)
-#define WINAPI_FAMILY_PARTITION(x) x
-#define WINAPI_PARTITION_DESKTOP 1
+/* detect whether it is Windows APP (UWP) or standard Win32 envionment */
+#if defined(WINAPI_FAMILY_PARTITION) &&\
+	WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP) && !WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)
+	#define FLAC_WINDOWS_APP 1
+#else
+	#define FLAC_WINDOWS_APP 0
 #endif
 
 static int local_vsnprintf(char *str, size_t size, const char *fmt, va_list va)
@@ -106,15 +109,22 @@ static wchar_t *wchar_from_utf8(const char *str)
 /* retrieve WCHAR commandline, expand wildcards and convert everything to UTF-8 */
 int get_utf8_argv(int *argc, char ***argv)
 {
+#if !FLAC_WINDOWS_APP
 	typedef int (__cdecl *wgetmainargs_t)(int*, wchar_t***, wchar_t***, int, int*);
 	wgetmainargs_t wgetmainargs;
 	HMODULE handle;
+#endif // !FLAC_WINDOWS_APP
 	int wargc;
 	wchar_t **wargv;
 	wchar_t **wenv;
 	char **utf8argv;
 	int ret, i;
 
+#if FLAC_WINDOWS_APP
+	wargc = __argc;
+	wargv = __wargv;
+	wenv = _wenviron;
+#else // !FLAC_WINDOWS_APP
 	if ((handle = LoadLibraryW(L"msvcrt.dll")) == NULL) return 1;
 	if ((wgetmainargs = (wgetmainargs_t)GetProcAddress(handle, "__wgetmainargs")) == NULL) {
 		FreeLibrary(handle);
@@ -126,8 +136,11 @@ int get_utf8_argv(int *argc, char ***argv)
 		FreeLibrary(handle);
 		return 1;
 	}
+#endif // !FLAC_WINDOWS_APP
 	if ((utf8argv = (char **)calloc(wargc, sizeof(char*))) == NULL) {
+	#if !FLAC_WINDOWS_APP
 		FreeLibrary(handle);
+	#endif // !FLAC_WINDOWS_APP
 		return 1;
 	}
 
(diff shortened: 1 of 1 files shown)


ktmf01:
Why is argv functionality not necessary anyway?

dmitrykos:
@ktmf01 my assumption was that classic UWP app is Window-based.

Thanks to your question I double checked about Console UWP app possibility and actually it already exists (missed that fact) and described in MS docs ["Create a Universal Windows Platform console app"](https://learn.microsoft.com/en-us/windows/uwp/launch-resume/console-uwp).

Therefore I updated my PR with changes which read argv in UWP environment, therefore `get_utf8_argv()` works again.

ktmf01:
Thanks for providing this patch. It'll take some time for me to review this.

Getting an alternative for LoadLibrary has been on my mind for a while now. Not for UWP (because it has been officially deprecated for a while now) but for other platforms in which msvcrt.dll might not be available. I think maybe Windows for ARM? That would make this patch have a broader effect.

dmitrykos:
It seems to me `msvcrt.dll` is always available, on Windows for ARM too, so `LoadLibrary` approach looks fine. But, if CMake is used for project generation then it could be possible to lookup `__argc`, `__wargv`, `_wenviron` symbols in `stdlib.h` and use/link them directly, similar to the approach in my PR. 

These symbols are a part of MS C library extension - [Global variables](https://learn.microsoft.com/en-us/cpp/c-runtime-library/global-variables?view=msvc-170) and this approach can replace or compliment `LoadLibrary` implementation.

ktmf01:
Have you tested this code? It segfaults at my end

dmitrykos:
> Have you tested this code? It segfaults at my end

Yes, I tested in VS under the debugger. Have you tried to run some UWP app, do you have a stack trace?

ktmf01:
The problem is that `WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP)` is also true on normal Windows desktop apparently, and it seems __wargv is NULL, causing a segfault.

dmitrykos:
@ktmf01 you are right, there is an additional check missing. I committed the fix, checked in Win32 and UWP environments. Now works fine.

dmitrykos:
@ktmf01 I decided to squash both commits for better consistency of the proposed change.

sezero:
This breaks build when winapifamily.h is not available:
`src/libFLAC/../share/win_utf8_io/win_utf8_io.c:44:25: error: missing binary operator before token "("`

I suggest something like the following. (Also see https://github.com/libsdl-org/SDL/blob/main/include/SDL3/SDL_platform_defines.h#L144-L170)
```diff
diff --git a/src/share/win_utf8_io/win_utf8_io.c b/src/share/win_utf8_io/win_utf8_io.c
index 1892174..a98bbfb 100644
--- a/src/share/win_utf8_io/win_utf8_io.c
+++ b/src/share/win_utf8_io/win_utf8_io.c
@@ -38,15 +38,26 @@
 #include "share/win_utf8_io.h"
 
 #define UTF8_BUFFER_SIZE 32768
 
+#if defined(_MSC_VER) && defined(__has_include)
+#if __has_include(<winapifamily.h>)
+#define HAVE_WINAPIFAMILY_H 1
+#else
+#define HAVE_WINAPIFAMILY_H 0
+#endif
+#endif
+
+#if HAVE_WINAPIFAMILY_H
+#include <winapifamily.h>
 /* detect whether it is Windows APP (UWP) or standard Win32 envionment */
 #if defined(WINAPI_FAMILY_PARTITION) &&\
 	WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP) && !WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)
 	#define FLAC_WINDOWS_APP 1
 #else
 	#define FLAC_WINDOWS_APP 0
 #endif
+#endif
 
 static int local_vsnprintf(char *str, size_t size, const char *fmt, va_list va)
 {
 	int rc;
```

ktmf01:
Thanks!

Perhaps better to detect winapifamily.h in configure.ac and CMakeLists.txt instead of through __has_include, to make this consistent with the rest of the code?

sezero:
> Perhaps better to detect winapifamily.h in configure.ac and CMakeLists.txt instead of through __has_include, to make this consistent with the rest of the code?

That works too I think, since you don't have build systems not relying on config.h such as MSVC project files.

dmitrykos:
@sezero you are right. My misassumption was that compiler will not do a 2-nd check if first fails (just like during run-time) but it is not true in case of pre-processor.

So implementation shall be fixed to such:

```c
#if defined(WINAPI_FAMILY_PARTITION)
    #if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP) && !WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)
        #define FLAC_WINDOWS_APP 1
    #else
        #define FLAC_WINDOWS_APP 0
    #endif
#else
    #define FLAC_WINDOWS_APP 0
#endif
```

I will provide a fix for it later in a separate PR.

ktmf01:
So, apparently GCC's preprocessor does short-circuit evaluation, but MSVC does not? Or at least it seems it tries to parse the entire expression before evaluating?

https://stackoverflow.com/questions/24037595/is-there-logical-short-circuiting-in-the-c-preprocessor
https://developercommunity.visualstudio.com/t/1468786

Anyway, @dmitrykos  thanks, I'll await your PR.

dmitrykos:
I just created new PR #610 to fix compile error reported by @sezero.

> So, apparently GCC's preprocessor does short-circuit evaluation, but MSVC does not? Or at least it seems it (older versions) tries to parse the entire expression before evaluating?

In my tests short-circuit does not happen in case of GCC and MSVC (Microsoft Visual Studio Community 2019 Version 16.11.23).

PR #583, opened 2023-04-07
clip output instead of returning error

harishdm:
In some cases, output may be outside the allowed range due to integer arithmetic. 
In such cases, instead of returning an error and inserting silence, output is clipped to be within the allowed range.

### DIFF
--- a/src/libFLAC/stream_decoder.c
+++ b/src/libFLAC/stream_decoder.c
@@ -2122,18 +2122,14 @@ FLAC__bool read_frame_(FLAC__StreamDecoder *decoder, FLAC__bool *got_a_frame, FL
 		if(do_full_decode) {
 			/* Undo any special channel coding */
 			undo_channel_coding(decoder);
-			/* Check whether decoded data actually fits bps */
+			/* Check whether decoded data actually fits bps, if not clip it */
 			for(channel = 0; channel < decoder->private_->frame.header.channels; channel++) {
 				for(i = 0; i < decoder->private_->frame.header.blocksize; i++) {
 					int shift_bits = 32 - decoder->private_->frame.header.bits_per_sample;
-					/* Check whether shift_bits MSBs are 'empty' by shifting up and down */
-					if((decoder->private_->output[channel][i] < (INT32_MIN >> shift_bits)) ||
-					   (decoder->private_->output[channel][i] > (INT32_MAX >> shift_bits))) {
-						/* Bad frame, emit error */
-						send_error_to_client_(decoder, FLAC__STREAM_DECODER_ERROR_STATUS_FRAME_CRC_MISMATCH);
-						decoder->protected_->state = FLAC__STREAM_DECODER_SEARCH_FOR_FRAME_SYNC;
-						break;
-					}
+					if(decoder->private_->output[channel][i] < (INT32_MIN >> shift_bits))
+						decoder->private_->output[channel][i] = (INT32_MIN >> shift_bits);
+					else if (decoder->private_->output[channel][i] > (INT32_MAX >> shift_bits))
+						decoder->private_->output[channel][i] = (INT32_MAX >> shift_bits);
 				}
 			}
 		}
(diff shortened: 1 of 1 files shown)


ktmf01:
Please explain why you think this is a better solution. The current behavior is not lossless, but it clearly lets the user know by sending an error and not outputting potentially broken output. The behavior this PR introduces is silently altering data. See the [contributing guidelines](https://github.com/xiph/flac/blob/master/CONTRIBUTING.md).

> **Goals**
>
> - FLAC should be and stay an open format with an open-source reference implementation.
> - **FLAC should be lossless. This seems obvious but lossy compression seems to creep into every audio codec. This goal also means that flac should stay archival quality and be truly lossless for all input. Testing of releases should be thorough.**

harishdm:
Thank you for reviewing this.

I agree that there is no clear right or wrong way of handling this case.

In versions 1.3.4 and before of this repository, decoder was just returning that data as is which lead to a +1 being interpreted as -1. Now the current version identifies this as an error and inserts silence. 
Instead of returning an error and inserting silence, clipping would provide an output that is similar to how it was handled prior to 
1793632ee6988deb933ff7551fee92134622b558.

for one particular case being tested, output was 0x8000 in [FLAC__lpc_restore_signal](https://github.com/xiph/flac/blob/master/src/libFLAC/lpc.c#L1003), which ideally should have been 0x7fff. Because of this difference of 1, entire frame was silenced.

Also clip would produce output similar to what other decoders produce. Tried one clip where current version detects values outside the range and inserts silence whereas ffmpeg and Audacity and both were able to successfully decode the same clip.

ktmf01:
On what kind of files do you have this problem? Because as far as I know, all files that have problems with 1793632 are simply broken. All files I have seen with this condition have been created by violating one of the requirements for using libFLAC, as stated here https://xiph.org/flac/api/group__flac__stream__encoder.html#ga87b9c361292da5c5928a8fb5fda7c423

> Each sample should be a signed integer, right-justified to the resolution set by [FLAC__stream_encoder_set_bits_per_sample()](https://xiph.org/flac/api/group__flac__stream__encoder.html#ga7453fc29d7e86b499f23b1adfba98da1). For example, if the resolution is 16 bits per sample, the samples should all be in the range [-32768,32767].

If you have other files that have problems with 1793632, I'd really like to hear about them. But I don't think it is a good idea to silently fix broken files without being sure what went wrong exactly.

harishdm:
I will check if the clip that shows this problem can be shared. I also agree that the clip maybe broken.
This PR was just trying to make the behavior similar to older versions of flac and other flac decoders.

Since clipping output is not a preferred solution, would it be okay to do this under a macro and keep the macro disabled by default? Whoever wants to get the clipped output instead of silence for such cases can define that macro?

If yes, I will move this clip under a macro PREFER_CLIPPED_OUTPUT?

ktmf01:
Even with such a macro, I'd like to send an error. Maybe a new one, like FLAC__STREAM_DECODER_ERROR_STATUS_NOT_LOSSLESS

Would that cause problems?

harishdm:
Sending error along with clipping works too.
I have added a commit which sends FLAC__STREAM_DECODER_ERROR_STATUS_FRAME_CRC_MISMATCH (same as what was being sent earlier for this case) when the output is clipped.
Now that the decoder signals an error when output is clipped, I think macro isn't really necessary.

So change with these two commits w.r.t. current tip of tree is, instead of silence decoder clips the output and doesn't seek to sync frame and continues to decode. Decoder will signal an error like it was being done before.

ktmf01:
Having had some time to ponder this, I am strongly disliking the change proposed here.

FLAC is a lossless codec, and while emitting silence instead of an actual signal isn't lossless either, I think it is a better solution than 'fixing' a signal by clipping it.

Also, FLAC uses each sample as a starting point to predict the next samples. If we have a sample that is out of bounds, a lot of assumptions no longer hold for the next samples, which might cause problems. It could very well be the predictor overflows, destabilizes and the whole signal chain generates static for the rest of the block. Only muting protects from that.

So I am inclined to reject this PR.

harishdm:
Thank you for reviewing the proposed changes.
Based on the discussion in these comments, I will close this pull request now.

PR #612, opened 2023-05-24
[CI] add Universal Windows Platform build

robUx4:


ktmf01:
Thanks for trying to make this work. I hate to be blunt, but isn't UWP deprecated since about 18 months ago? I haven't had many complaints about it not working, and AFAIK the FLAC libs work fine for anyone wanting to use them in an UWP app.

If you can get it to work, that would be great, but I'm not sure it is worth a lot of trouble?

robUx4:
It does compile and should work (no extra patch needed, as soon as mingw-w64 is patched). This extra CI would be to make sure it doesn't break in the future. Not sure if not running tests would be an option, otherwise we're close.

As for being dead, it's hard to tell. If you want to publish Apps on the Xbox that's still the only way to go. Otherwise there's the GDK option (not available to everyone) and then you have other restrictions triggered by `WINAPI_FAMILY_APP` instead of `WINAPI_FAMILY_GAMES`. This MR could totally be switched to that. That mode is definitely alive at Microsoft. They even added support to vcpkg for it (semi officially).

ktmf01:
Oh, OK. It seemed the information about the state of UWP is rather vague at best.

ktmf01:
I've tried to run this locally, it seems UWP apps need to run in some kind of sandbox? Whatever I compile for UWP, it won't run apparently. That seems to be the reason the tests fail now. @dmitrykos perhaps you know how to run/test the compiled executable?

dmitrykos:
@ktmf01 I have no experience with UWP tests automation, so can't point to some ready solution unfortunately but what I know is that UWP Console app (I assume test is a console app) has to declare self as console app - `desktop4:Subsystem="console"`. Microsoft gives some hints in [Create a Universal Windows Platform console app](https://learn.microsoft.com/en-us/windows/uwp/launch-resume/console-uwp).

When I was experimenting with #581 I installed UWP Console App extension in Visual Studio 2019, then created "Hello World!" app and when started it Visual Studio deployed compiled UWP Console app to the OS, so I could see it in Start menu. Later it could be invoked in PowerShell.

robUx4:
I identified the issue. The problem is that when building for UWP, it relies on a `vcruntime140_app.dll` for some runtime calls (memcmp, strchr, strrchr, memset, memcpy, memchr, memmove). The `_app` part of the DLL name means it's only for UWP/UAP apps. Except that DLL is not in the `PATH` when running Ctest. Once you put it in the PATH, the tests work (the .exe of the tests are found in the proper location). In mingw-w64 we can get around this problem by using `api-ms-win-crt-private-l1-1-0.dll` which contains all the API's we need.

Now the question is how to put that DLL in the `PATH`, or not use it at all when using Visual Studio. I suppose it should be found in the Windows SDK to be redistributed, but it doesn't seem so. It is found in some WinSxS package in the Windows folder, but it's only meant to be used through a manifest. It also seems to be found in some path like `C:\Program Files\WindowsApps\Microsoft.VCLibs.140.00_14.0.32530.0_x64__8wekyb3d8bbwe`. But I'm not sure there's an easy way to find this path.

There's also something called [Hybrid CRT](https://github.com/microsoft/WindowsAppSDK/blob/main/docs/Coding-Guidelines/HybridCRT.md) which is supposed to simplify those cased by using a static library for regular CRT calls. I've followed their instructions but it still ends up requiring `vcruntime140_app.dll` :( Although I don't have a HybridCRT.props file on my system...

PR #634, opened 2023-07-06
Add multithreading to libFLAC

ktmf01:


ktmf01:
A lively discussion with test results, comparisons and suggestions can be found here: https://hydrogenaud.io/index.php/topic,124437.0.html

sezero:
This seems to only implement threading via pthreads even for windows, e.g. autotools/cmake check pthreads unconditionally. Is that really the intention? (I know mingw-w64 has pthread implementation, but still...)

ktmf01:
Yes. This had been extensively tested for Windows on mingw-w64. Could you explain what you think is wrong with that?

sezero:
> Could you explain what you think is wrong with that?

The extra dependency on the pthread dll whereas win32 apis could be used directly. But that's an administrative decision, so I won't fight it.

ktmf01:
I don't know what you mean with administrative. I would think to use the win32 api's directly would mean adding quite a bit of code, because those APIs can't be used in the same way pthreads can be used? That code would get little CI coverage and no fuzzing coverage, so I'd rather not do that.

Issue #637, opened 2023-07-19
incompatible pointer types in bitmath.h


manxorist:
Compiling with Clang-cl on Windows amd64 results in the following warnings:
```
src\libFLAC\include\private/bitmath.h(86,19): warning : incompatible pointer types passing 'uint32_t *' (aka 'unsigned int *') to parameter of type 'unsigned long *' [-Wincompatible-pointer-types]
20>C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\Llvm\x64\lib\clang\15.0.1\include\intrin.h(135,46): note: passing argument to parameter '_Index' here

src\libFLAC\include\private/bitmath.h(110,21): warning : incompatible pointer types passing 'uint32_t *' (aka 'unsigned int *') to parameter of type 'unsigned long *' [-Wincompatible-pointer-types]
20>C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\Llvm\x64\lib\clang\15.0.1\include\intrin.h(249,48): note: passing argument to parameter '_Index' here
```

For MSVC, these intrinsics are declared as:
```
_Success_(return!=0)
BOOLEAN
_BitScanReverse (
    _Out_ DWORD *Index,
    _In_ DWORD Mask
    );

_Success_(return!=0)
BOOLEAN
_BitScanReverse64 (
    _Out_ DWORD *Index,
    _In_ DWORD64 Mask
    );
```
(with `DWORD` being `typedef unsigned long`).

For Clang-cl, these intrinsics are declared as:
```
unsigned char _BitScanReverse(unsigned long *_Index, unsigned long _Mask);

unsigned char _BitScanReverse64(unsigned long *_Index, unsigned __int64 _Mask);
```

<https://learn.microsoft.com/en-us/cpp/intrinsics/bitscanreverse-bitscanreverse64?view=msvc-170> documents them as
```
unsigned char _BitScanReverse(
   unsigned long * Index,
   unsigned long Mask
);

unsigned char _BitScanReverse64(
   unsigned long * Index,
   unsigned __int64 Mask
);
```

I cannot test with any Intel Compiler on Windows.

[...]

PR #638, opened 2023-07-19
Use correct argument type for _BitScanReverse and _BitScanReverse64.

manxorist:
_BitScanReverse and _BitScanReverse64 are documented to take an `unsigned long *` as the first argument (see <https://learn.microsoft.com/en-us/cpp/intrinsics/bitscanreverse-bitscanreverse64?view=msvc-170>), however libFLAC used `uint32_t` which happens to be `unsigned int`.

This silences Clang warning `incompatible pointer types passing 'uint32_t *' (aka 'unsigned int *') to parameter of type 'unsigned long *' [-Wincompatible-pointer-types]`.

Fixes https://github.com/xiph/flac/issues/637

### DIFF
--- a/src/libFLAC/include/private/bitmath.h
+++ b/src/libFLAC/include/private/bitmath.h
@@ -82,7 +82,7 @@ static inline uint32_t FLAC__clz_uint32(FLAC__uint32 v)
 	return __builtin_clz(v);
 #elif defined(_MSC_VER)
 	{
-		uint32_t idx;
+		unsigned long idx;
 		_BitScanReverse(&idx, v);
 		return idx ^ 31U;
 	}
@@ -106,7 +106,7 @@ static inline uint32_t FLAC__clz_uint64(FLAC__uint64 v)
 	return __builtin_clzll(v);
 #elif (defined(__INTEL_COMPILER) || defined(_MSC_VER)) && (defined(_M_IA64) || defined(_M_X64))
 	{
-		uint32_t idx;
+		unsigned long idx;
 		_BitScanReverse64(&idx, v);
 		return idx ^ 63U;
 	}
@@ -160,7 +160,7 @@ static inline uint32_t FLAC__bitmath_ilog2(FLAC__uint32 v)
 	return _bit_scan_reverse(v);
 #elif defined(_MSC_VER)
 	{
-		uint32_t idx;
+		unsigned long idx;
 		_BitScanReverse(&idx, v);
 		return idx;
 	}
@@ -177,7 +177,7 @@ static inline uint32_t FLAC__bitmath_ilog2_wide(FLAC__uint64 v)
 /* Sorry, only supported in x64/Itanium.. and both have fast FPU which makes integer-only encoder pointless */
 #elif (defined(__INTEL_COMPILER) || defined(_MSC_VER)) && (defined(_M_IA64) || defined(_M_X64))
 	{
-		uint32_t idx;
+		unsigned long idx;
 		_BitScanReverse64(&idx, v);
 		return idx;
 	}


ktmf01:
I'll take a look at this soon

ktmf01:
This seems to conflict with this:
https://www.intel.com/content/www/us/en/docs/cpp-compiler/developer-guide-reference/2021-8/miscellaneous-intrinsics-001.html

```
unsigned char _BitScanReverse(unsigned __int32 *p, unsigned __int32 b);
```

```
unsigned char _BitScanReverse64(unsigned __int32 *p, unsigned __int64 b);
```

I don't have the Intel compiler installed to test. Maybe it is a good idea to change this so it really only applies to MSVC?

Also, I'm confused about whether this warning is actually signalling a real problem. From what I read, Microsoft considers long and int both 32-bit.

src/libFLAC/include/private/bitmath.h
```diff
@@ -106,7 +106,7 @@ static inline uint32_t FLAC__clz_uint64(FLAC__uint64 v)
 	return __builtin_clzll(v);
 #elif (defined(__INTEL_COMPILER) || defined(_MSC_VER)) && (defined(_M_IA64) || defined(_M_X64))
 	{
-		uint32_t idx;
+		unsigned long idx;
```
ktmf01, line 109:
```suggestion
#if defined(_MSC_VER)
		unsigned long idx;
#else
		uint32_t idx;
#endif
```

src/libFLAC/include/private/bitmath.h
```diff
@@ -177,7 +177,7 @@ static inline uint32_t FLAC__bitmath_ilog2_wide(FLAC__uint64 v)
 /* Sorry, only supported in x64/Itanium.. and both have fast FPU which makes integer-only encoder pointless */
 #elif (defined(__INTEL_COMPILER) || defined(_MSC_VER)) && (defined(_M_IA64) || defined(_M_X64))
 	{
-		uint32_t idx;
+		unsigned long idx;
```
ktmf01, line 180:
```suggestion
#if defined(_MSC_VER)
		unsigned long idx;
#else
		uint32_t idx;
#endif
```

ktmf01 (commented):
Maybe like this?

manxorist:
This will be just as wrong according to the documentation you cited. If Intel documents this as `unsigned __int32`, it literally needs to be type  `unsigned __int32`. `uint32_t` may or may not be the same type.

Also, the Intel Compiler (as does Clang) pretends to be all sorts of other compilers, this includes defining `_MSC_VER` on Windows (see <https://www.intel.com/content/www/us/en/docs/cpp-compiler/developer-guide-reference/2021-10/additional-predefined-macros.html>), so the guard you suggested will not work.

Note that there are a lot of different Intel Compilers in existence nowadays:
 1. icc (based on (I think) an EDG frontend in GCC compatibility mode for Linux)
 2. icl (based on (I think) an EDG frontend in MSVC compatibility mode for Windows)
 3. icx (based on LLVM/Clang)
 4. dpcpp (based on LLVM/Clang, which (I think) is a hybrid compiler for targeting CPU and GPU)
(see <https://www.intel.com/content/www/us/en/developer/articles/technical/use-predefined-macros-for-specific-code-for-intel-dpcpp-compiler-intel-cpp-compiler-intel-cpp-compiler-classic.html> and <https://www.intel.com/content/www/us/en/docs/cpp-compiler/developer-guide-reference/2021-10/overview.html>).

So, as we cannot test with any Intel Compiler on Windows, I suggest we either keep `unsigned long` for `__INTEL_COMPILER`, or change it to `unsigned __int32` for `__INTEL_COMPILER` as per documentation. In the 2 32bit cases, `__INTEL_COMPILER` is already handled differently anyway.

> Also, I'm confused about whether this warning is actually signalling a real problem. From what I read, Microsoft considers long and int both 32-bit.

`unsigned long` and `unsigned int` (and the Microsoft extension `unsigned __int32`) are all distinct types in the C Standard (whether they happen to be of the same size does not matter in lights of C). As per the C Standard, code is not allowed to alias pointers of incompatible types (except for character types). GCC and Clang are known to optimize based on that guarantee. I do not know whether Clang does in this particular situation with intrinsics, but this is not a risk any line of code should take. The compiler warns because it explicitly detected Undefined Behavior here, and the code should get fixed.

ktmf01:
Note to self, this reverts commits c6318e9, 8967db0 and f60a8cc. It looks like this was incorrectly changed in an effort to replace all use of 'unsigned' with types of explicit size. Apparently this is not correct for this specific case.

manxorist:
Yeah, it looks like the argument types to the intrinsic function should never have been changed in the first place. All other changes in the given commits in `bitmath.h` look fine to me. In that case, I would suggest to just revert to the old behaviour for Intel Compiler, which would be `unsigned long`. It was maybe missed in <https://github.com/xiph/flac/commit/c6318e9dd3f7a91f40340911bcf57bf36768910e> that these variables are used to call intrinsics.

In that case, the patch itself does not need changing. Should I update the commit message to reflect our findings?

ktmf01:
I've added a reference to this PR to the commit message, I think that should do.

PR #654, opened 2023-11-05
flac: foreign_metadata: fix -Walloc-size

thesamesam:
GCC 14 introduces a new -Walloc-size included in -Wextra which gives:
```
src/flac/foreign_metadata.c:803:33: warning: allocation of insufficient size ‘1’ for type ‘foreign_metadata_t’ with size ‘64’ [-Walloc-size]
```

The calloc prototype is:
```
void *calloc(size_t nmemb, size_t size);
```

So, just swap the number of members and size arguments to match the prototype, as we're initialising 1 struct of size `sizeof(foreign_metadata_t)`. GCC then sees we're not doing anything wrong.


thesamesam:
Thanks!

Issue #494, opened 2022-10-21
Building for Android with ANDROID_PLATFORM < 24 results in fseeko() and ftello() always return error in 32-bit build


svenoaks:
The [32-bit ABI bugs document](https://android.googlesource.com/platform/bionic/+/master/docs/32-bit-abi.md) suggests they're not available, but I didn't notice it until encoder `finish()` is always false and metadata not written, it seems like `fseeko()` is always returning error.

I'm not sure how to fix it yet, I can't raise my ANDROID_PLATFORM to 24 yet.

ktmf01:
A stupid question perhaps, but what would one gain by building libFLAC on Android? libFLAC is already part of the Android system API's.

Related to that: perhaps there is a patch somewhere for this in the Android sources? https://android.googlesource.com/platform/external/flac/

svenoaks:
> A stupid question perhaps, but what would one gain by building libFLAC on Android? libFLAC is already part of the Android system API's.
> 
> Related to that: perhaps there is a patch somewhere for this in the Android sources? https://android.googlesource.com/platform/external/flac/

As far as I can tell, libFLAC can't be used directly by app developers, you have to go through through the Media API, which is cumbersome. I'll look more, but can't see how (if?) they resolved the fseeko thing there

svenoaks:
I added this to compat.h 

```
#if defined __ANDROID__ && (defined __arm__ || defined __i386__) && __ANDROID_API__ < 24
#define fseeko fseek
#define ftello ftell
#endif
```

and it works, but I think that will fail with files larger than 2GB.

ktmf01:
I can't seem to find any changes in Android's version of FLAC related to this.

Perhaps it is interesting to try and generate a FLAC file through some existing Android app using the Media API? If such a file also lacks the metadata you reference, perhaps this isn't fixed in the Media API either?

Anyway, I can't be of much help here, sadly.

ktmf01:
I've been looking through the libFLAC that shipped with Android 6.0 (which shipped with API level 23) at https://android.googlesource.com/platform/external/flac/+/refs/tags/android-6.0.0_r26/

I can't find anything relating to this. It seems fseeko is used at several places. So, it seems fseeko wasn't made available but is available at a lower level?

Can I close this issue? You've found a fix (from the document you linked it seems the only possible fix).

svenoaks:
[...]

PR #659, opened 2023-11-29
Add the needed Android API check for fseeko and ftello.

google-yfyang:
fseeko and ftello are available on Android API 24 or later. 

See for the libc src. https://android.googlesource.com/platform/bionic/+/master/libc/include/stdio.h#211

image

### DIFF
--- a/include/share/compat.h
+++ b/include/share/compat.h
@@ -213,6 +213,10 @@
 
 #ifdef ANDROID
 #include <limits.h>
+#if __ANDROID_API__ < 24
+#define fseeko fseek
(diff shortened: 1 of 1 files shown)


google-yfyang:
@ktmf01 Hi Martijn, can you please take a look?

ktmf01:
I can't verify this. #494 is related.

robUx4:
This is not correct. It's disabling fseeko for **all** Android targets with an API older than 24. However `x64_86` and `aarch64` are perfectly fine with 64-bit fseeko.

PR #667, opened 2024-01-19
chained streams

philippe44:
This is probably not yet an acceptable PR but getting closer and I wanted to know if this goes into the right direction for you. 

Beyond the simple, dirty hack that is to reset serial number on page EoS which works but creates some sync issue for new metadata headers that are missed, this PR tries to properly handle chained stream. 

I realized that the difference of data acquisition where flac pulls and ogg pushes creates a real challenge and I wanted to minimize modification to code. What happens is that when EoS is detected in the ogg decoder, we have no idea when that packet (audio frame) will be processed by the flac decoder. When it pulls 8kB of data, it could contain 2 remaining audio frames, then the streaminfo/vorbis/.. and audio frame for the next stream, alltogether. The ogg decoder has detected EoS but for the flac decoder, we are far from being there.

So it's easy to stop the ogg decoder when it has sent all packets of the EoS page but we still cannot instruct yet the flac decoder to self-reset because we don't know how many frames are in its internal buffer, although the read callback has sent everything. The only solution I found is to block ogg decoder once it has sent the last packet of the EoS page *and* fully "simulate" a FLAC_END_OF_STREAM to the flac decoder. Then we have to wait for flac to re-authorize ogg decoder to grab data once it has detected that end of stream. It's fully hidden to client, even when they decode frame-by-frame. They will not see a return from `FLAC__stream_decoder_process_single` until we have certainty. Once re-authorized to grab data, a "real" end of stream will be generated if we are really at the end of the stream. Otherwise, the decoder will reset, grab streaminfo and all other metadata headers and start decoding the new stream.

I've now added the boilerplate code to enabme/disable it as well as a `--chaining` option in flac application. When it is disabled, I don't think a single extra line of code is executed that was not before, so I'm hopeful there is no regression.

### DIFF
--- a/src/libFLAC/stream_decoder.c
+++ b/src/libFLAC/stream_decoder.c
@@ -684,7 +684,7 @@ FLAC_API FLAC__bool FLAC__stream_decoder_set_ogg_serial_number(FLAC__StreamDecod
 #endif
 }
 
-FLAC_API FLAC__bool _FLAC__stream_decoder_set_ogg_chaining(FLAC__StreamDecoder* decoder, FLAC__bool value)
+FLAC_API FLAC__bool FLAC__stream_decoder_set_ogg_chaining(FLAC__StreamDecoder* decoder, FLAC__bool value)
 {
 	FLAC__ASSERT(0 != decoder);
 	FLAC__ASSERT(0 != decoder->protected_);
@@ -837,7 +837,7 @@ FLAC_API const char *FLAC__stream_decoder_get_resolved_state_string(const FLAC__
 	return FLAC__StreamDecoderStateString[decoder->protected_->state];
 }
 
-FLAC_API FLAC__bool _FLAC__stream_decoder_get_ogg_chaining(const FLAC__StreamDecoder* decoder)
+FLAC_API FLAC__bool FLAC__stream_decoder_get_ogg_chaining(const FLAC__StreamDecoder* decoder)
 {
 	FLAC__ASSERT(0 != decoder);
 	FLAC__ASSERT(0 != decoder->protected_);
--- a/src/libFLAC/stream_decoder.c
+++ b/src/libFLAC/stream_decoder.c
@@ -112,6 +112,7 @@ static FLAC__StreamDecoderSeekStatus file_seek_callback_(const FLAC__StreamDecod
 static FLAC__StreamDecoderTellStatus file_tell_callback_(const FLAC__StreamDecoder *decoder, FLAC__uint64 *absolute_byte_offset, void *client_data);
 static FLAC__StreamDecoderLengthStatus file_length_callback_(const FLAC__StreamDecoder *decoder, FLAC__uint64 *stream_length, void *client_data);
 static FLAC__bool file_eof_callback_(const FLAC__StreamDecoder *decoder, void *client_data);
+static void reset_decoder(FLAC__StreamDecoder* decoder);
 
 /***********************************************************************
  *
@@ -1006,16 +1007,9 @@ FLAC_API FLAC__bool FLAC__stream_decoder_process_single(FLAC__StreamDecoder *dec
 #if FLAC__HAS_OGG
 		if(decoder->private_->is_ogg &&
 		   decoder->protected_->state == FLAC__STREAM_DECODER_END_OF_STREAM &&
-		   FLAC__ogg_decoder_aspect_page_eos(&decoder->protected_->ogg_decoder_aspect, true)) {
-			decoder->protected_->state = FLAC__STREAM_DECODER_SEARCH_FOR_METADATA;
-			decoder->private_->has_stream_info = false;
-			decoder->private_->first_frame_offset = 0;
-			decoder->private_->unparseable_frame_count = 0;
-			decoder->private_->last_seen_framesync = 0;
-			decoder->private_->last_frame_is_set = false;
-		}
+		   FLAC__ogg_decoder_aspect_page_eos(&decoder->protected_->ogg_decoder_aspect, true))
+			reset_decoder(decoder);
 #endif
-
 		switch(decoder->protected_->state) {
 			case FLAC__STREAM_DECODER_SEARCH_FOR_METADATA:
 				if(!find_metadata_(decoder))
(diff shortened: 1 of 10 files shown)


philippe44:
⬆️ ⬇️ ?

ktmf01:
I am currently not able to review this and I probably won't for quite a while. I am not very familiar with the ogg code and as you say it is rather complicated.

philippe44:
Understood, no worries. What I'll do then is add the "boiler plate" code so that is complete while it's still fresh in my mind and you can review it at your leisure.

src/libFLAC/stream_decoder.c
```diff
@@ -1007,8 +1037,16 @@ FLAC_API FLAC__bool FLAC__stream_decoder_process_single(FLAC__StreamDecoder *dec
 				else
 					return true;
 			case FLAC__STREAM_DECODER_SEARCH_FOR_FRAME_SYNC:
-				if(!frame_sync_(decoder))
-					return true; /* above function sets the status for us */
+				if(!frame_sync_(decoder)) {
```
ktmf01, line 1040: Do you think there might be a way to replace this (and similar) bits of code with more code in ogg_decoder_aspect.c? The current approach 'catches' end of stream in the generic code, but I think it would be nice if the ogg_decoder_aspect code could handle that. More specifically, it would be great if as much handling as possible would be in the function `FLAC__ogg_decoder_aspect_read_callback_wrapper`

ktmf01:
Hi @philippe44, sorry it took quite a while to review this. I think the code looks nice, but it would be even nicer if as much of it as possible would be in `FLAC__ogg_decoder_aspect_read_callback_wrapper`.

Also, can you explain to me what the last few commits' back-and-forth is supposed to do?

philippe44:
No worries, I understand what it is to be a maintainer and thank you for doing it, this is hard as the recent xz issue reminded us.

The latest back and forth was a miserable failure to try to deal with some backlash I had from people using project that required (optionally) my changes and that had to deal with existing libflac for static and dynamic linking. GCC and LLVC offer ways to deal with that at run time but not MSVC and not all versions of GCC. At the end I could not find a good solution so I just added a #define, parking to see what you'd think if/when you had time to review it.

I'll review the rest of your comments soon.

fabiangreffrath:
Any chance this gets merged into the main branch soon?

ktmf01:
No. However, you can help along by explaining your use case? The main problem I currently have is that I want to implement this in a way that covers as much usage scenarios as possible, and I don't knownhow people want the API to function

fabiangreffrath:
I have been pointed here by the following Debian bug report:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=498048

philippe44:
How did you try it? I've been running it for many month now and it works for me.

whatamesss:
@philippe44 i simply cloned your git master and built it. can you confirm that this stream plays for you with metadata and continues after a song change? http://radio3.radio-calico.com:8080/calico   ..thanks!

ktmf01:
> such a brutal bug 

This is no bug, this is a missing feature, and a rather complex one too. Sure, this looks like 'just keep playing' but to me this is a feature that either should be implemented properly (with seeking, metadata handling etc.) or not at all. And such an implementation means breaking the API, which also means we can't 'just fix it' in the next release either.

> going unfixed for so long.

Apparently no-one cares enough about FLAC to take a role regularly adding new features after the creator left the project 14 years ago.

philippe44:
I think I've done most of it, except seeking which is really an order of magnitude more difficult. But in general, need it's not an easy feature, despite what if seems. At least it was not an easy addition for me to do :-)

philippe44:
I think I've done most of it except the seeking which is an order of magnitude more complex. But it was indeed not an easy update, despite what it seems and I thought at the beginning. At least it was not easy for me :-) Hence I pushed for a review as my memory of what I did fades away ...

whatamesss:
i'm not even sure how or why seeking is accomplished in a live stream, which is where i've found oggchaining 100% of the time..
seeking is completely unecessary, unlike continuous playback. many broadcasters/streamers use the format, convicing them all to change something sounds considerably more difficult than making it work at the receiving end. 
after all, arent we just talking about making it work on linux? if windows users were experiencing the same problem, i couldnt imagine so many stations adoptng the format...it just wouldnt work, such as is the case for us poor linux users.

philippe44:
Not sure I understand - it work on all platforms (my PR)

philippe44:
Are you sure that mpd uses the right library? You probably know, but you can use a static flac library linked with your application (mod, but not that I have no experience rebuilding mpd) or the shared library but you have to make sure that the newly built .so is installed or in the LD_LIBRARY of the application you are running. 

In addition, and it is coming as while I'm writing that, per @ktmf01 asked, I built this extension of the api with a requirement to explicitly enable chained stream. The reasonable requirement is to have NO change compared to existing behavior. You can either change mpd or change my flac mod so that chained are enabled by default.

whatamesss:
i'm fairly certain mpd is using the system flac...which i'm building from https://github.com/philippe44/flac.git..
#equery g mpd|grep flac  [  1]  media-libs/flac-9999  

to be clear i'm not doing any streaming, i just have a playlist of maybe 30 radio stations in mpd that i switch between.

" built this extension of the api with a requirement to explicitly enable chained stream. " -do you mean disable?
" You can either change mpd or change my flac mod so that chained are enabled by default." -how do i enable chaining in your flac mod? i'm not sure if i need to pass an extra arg here. configure shows:      Ogg/FLAC support : ........................ yes

philippe44:
> " built this extension of the api with a requirement to explicitly enable chained stream. " -do you mean disable?
> 
No, by default my PR has chained disabled to be compatible with existing applications 

> " You can either change mpd or change my flac mod so that chained are enabled by default." -how do i enable chaining in your flac mod? i'm not sure if i need to pass an extra arg here. configure shows:      Ogg/FLAC support : ........................ yes
> 
No, you need to make a api call to enable it

whatamesss:
@philippe44 i was not aware that chaining was disabled by default! i have patched your source to force-enable it ...a configure option would be preferable here, but it works! thanks so much for improving my life! i hope xiph/flac can integrate this sometime soon for all the stream listeners out there

philippe44:
Glad it worked! I know it was a pain for you but to some extend showing that it is not enabled by default demonstrates that the PR is not modifying current behavior 😃

ktmf01:
@philippe44 I took a few hours to study your approach in-depth and I think it is rather nice. However, I would like to make some changes. Normally I would make a few changes right in PR itself (for which you have given permission to do) but it seems that tree is being used by people wanting this functionality. Do you mind if I 'duplicate' this PR, squash your commits and make a few changes? Most importantly, I would like to rename all instances of 'chaining' (variables, API functions, command line options) to 'decode_chained_stream' which I think is more descriptive.

philippe44:
Yes, please do as you want.

whatamesss:
@ktmf01 i'm happy to test, please ping me when you've committed something as i'm not syncing with xiph/flac

ktmf01:
Implemented by https://github.com/xiph/flac/pull/735

PR #685, opened 2024-04-03
Fix CMake Doxygen install (match Makefile builds)

jevinskie:
The CMake install was trying to install `<gitroot>/docs/api/` instead of `<buildroot>/docs/doxytemp/`. Matches Makefile behavior.

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -22,5 +22,5 @@ add_custom_target(FLAC-doxygen ALL
         COMMENT "Generating API documentation with Doxygen"
         VERBATIM )
 
-install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/api/"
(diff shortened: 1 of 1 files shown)


ktmf01:
Diving into this, I see now that CMake Doxygen installing is broken anyway, but this patch would make installing from a tarball even more broken.

It would probably be better to have a post build hook to rename the `doxytmp/html` directory to `api`. That is what [the makefile](https://github.com/xiph/flac/blob/master/doc/Makefile.am#L26) does, so that would truely match it.

Note to self: Currently api docs are only installed when doxygen is detected, in which case they are rebuild anyway. The [return here](https://github.com/xiph/flac/blob/5f6a352921cd58360a6ace049ee26068fd5b888c/doc/CMakeLists.txt#L12) probably needs a little more sophistication.

PR #691, opened 2024-04-22
disable fseeko in android before API 24

robUx4:
It can be linked in the NDK so `check_function_exists()` detects it. But it's only supporting `_FILE_OFFSET_BITS=64`
[since Android API 24](https://android.googlesource.com/platform/bionic/+/main/docs/32-bit-abi.md).

With NDK 26 it's no longer possible to build assuming the API is always available.

Fallback to 32-bit fseek/ftell whenever `HAVE_FSEEKO` is not set.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -125,7 +125,15 @@ else()
     check_include_file("x86intrin.h" FLAC__HAS_X86INTRIN)
 endif()
 
-check_function_exists(fseeko HAVE_FSEEKO)
+
+if(ANDROID AND CMAKE_SYSTEM_VERSION VERSION_LESS 24 AND (CMAKE_SYSTEM_PROCESSOR MATCHES "i686" OR CMAKE_SYSTEM_PROCESSOR MATCHES "armv7-a"))
+    # fseeko/ftello may link, but it's not usable before Android API 24 on 32-bit Android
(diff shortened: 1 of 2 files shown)


ktmf01:
Hi @robUx4,

Thanks for the PR. However, I don't know how to check this. There is another PR, #659, that tries to do the same, but simpler. Could you perhaps review that one, as it does not seem to rely on CMake?

robUx4:
It does not "rely" on CMake because the fix in the header will work for every autotools build that also detects `fseeko()` is not available.

The CMake fix is to handle the fseeko detection for Android too.

ktmf01:
Yes, but I guess that means the autotools build system would need a fix too? I have absolutely zero knowledge on building for Android.

robUx4:
autotools uses [AC_FUNC_FSEEKO](https://github.com/xiph/flac/blob/master/configure.ac#L46) which is a common autoconf function: https://www.gnu.org/software/autoconf/manual/autoconf-2.66/html_node/Particular-Functions.html. It will define `HAVE_FSEEKO` accordingly.

I assume it's working with Android too, but we (VLC) do not have any contrib that use it (or rather we build with CMake for those who do).

ktmf01:
Thanks! I guess there's no patching for autotools necessary there indeed.

Issue #658, opened 2023-11-21
Windows x86-64 Clang compiler fails to compile because of lacking -mavx2


Rt39:
I'm compileing flac on Windows x86-64 with Clang version 17.0.1 and targeting at x86_64-pc-windows-msvc with Visual Studio 2022 installed. This is the command I used to configure:
```
cmake .. -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DWITH_OGG=OFF -DBUILD_EXAMPLES=OFF -DBUILD_TESTING=OFF -DBUILD_DOCS=OFF -DINSTALL_MANPAGES=OFF -DINSTALL_PKGCONFIG_MODULES=OFF -DCMAKE_INSTALL_PREFIX="D:\flac"
```

and I got the error
```
C:/Users/Administrator/Downloads/Compressed/flac-master/src/libFLAC/stream_encoder_intrin_avx2.c:65:5: error: use of undeclared identifier '__m256i'
   65 |                                 __m256i sum256 = _mm256_setzero_si256();
      |                                 ^
C:/Users/Administrator/Downloads/Compressed/flac-master/src/libFLAC/stream_encoder_intrin_avx2.c:70:6: error: use of undeclared identifier '__m256i'
   70 |                                         __m256i res256 = _mm256_abs_epi32(_mm256_loadu_si256((const __m256i*)(const void*)(residual+residual_sample)));
      |
```

It seems you need to add compile flag `-mavx2` to the compiler, where I add the `-DCMAKE_CXX_FLAGS="-mavx2" -DCMAKE_C_FLAGS="-mavx2"` to the command and everthing works well.

Rt39:
It seems everything go well in the cmake console output
```
D:\cs\flac-master\build>cmake .. -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DWITH_OGG=OFF -DBUILD_EXAMPLES=OFF -DBUILD_TESTING=OFF -DBUILD_DOCS=OFF -DINSTALL_MANPAGES=OFF -DINSTALL_PKGCONFIG_MODULES=OFF -DCMAKE_INSTALL_PREFIX="D:\flac" -DCMAKE_CXX_FLAGS="-mavx2" -DCMAKE_C_FLAGS="-mavx2"
[...]

PR #718, opened 2024-07-09
On WIN32, fix compile errors on non MSVC compilers

Rt39:
#658


src/libFLAC/CMakeLists.txt
```diff
@@ -21,6 +21,9 @@ if(FLAC__CPU_X86_64 OR FLAC__CPU_IA32)
     if(WITH_AVX AND MSVC)
         set_source_files_properties(fixed_intrin_avx2.c lpc_intrin_avx2.c stream_encoder_intrin_avx2.c lpc_intrin_fma.c PROPERTIES COMPILE_FLAGS /arch:AVX2)
     endif()
+    if(WITH_AVX AND WIN32 AND NOT MSVC)
+        set_source_files_properties(fixed_intrin_avx2.c lpc_intrin_avx2.c stream_encoder_intrin_avx2.c lpc_intrin_fma.c PROPERTIES COMPILE_FLAGS -mavx2)
```
ktmf01, line 25: I don't understand why this is necessary. I regularly build with Clang on Win32 through MinGW, and that works perfectly fine without this patch.

So, either the version of Clang you're using is modified in some way, or there is something with the MSVC header files happening here.

In any case, the `if()` is too broad because it also applies to MinGW.

src/share/getopt/getopt.c
```diff
@@ -221,7 +221,7 @@ static char *posixly_correct;
 /* Avoid depending on library functions or files
    whose names are inconsistent.  */
 
-#ifndef getenv
+#if !defined(getenv) && !(defined(__clang__) && defined(_MSC_VER))
```
ktmf01, line 224: I'm not too sure what this does, and why it doesn't work for your particular configuration. Maybe it can be removed completely? I think it catches some corner case that existed 30 years ago or something.

Rt39: It seems on Windows, using Clang with MSVC headers would cause `error: conflicting types for 'getenv'`

```
X:\path\to\clang.exe -DHAVE_CONFIG_H -D_DARWIN_C_SOURCE -D_FORTIFY_SOURCE=2 -D_POSIX_PTHREAD_SEMANTICS -D_TANDEM_SOURCE -D__STDC_WANT_IEC_60559_BFP_EXT__ -D__STDC_WANT_IEC_60559_DFP_EXT__ -D__STDC_WANT_IEC_60559_FUNCS_EXT__ -D__STDC_WANT_IEC_60559_TYPES_EXT__ -D__STDC_WANT_LIB_EXT2__ -D__STDC_WANT_MATH_SPEC_FUNCS__ -IX:/path/to/flac/include -IX:/path/to/flac/build -Wall -Wextra -Wstrict-prototypes -Wmissing-prototypes -Waggregate-return -Wcast-align -Wnested-externs -Wshadow -Wundef -Wmissing-declarations -Winline  -DNDEBUG -O3 -DNDEBUG -D_DLL -D_MT -Xclang --dependent-lib=msvcrt -Wdeclaration-after-statement -fstack-protector-strong -MD -MT src/share/getopt/CMakeFiles/getopt.dir/getopt.c.obj -MF src\share\getopt\CMakeFiles\getopt.dir\getopt.c.obj.d -o src/share/getopt/CMakeFiles/getopt.dir/getopt.c.obj -c X:/path/to/flac/src/share/getopt/getopt.c
X:/path/to/flac/src/share/getopt/getopt.c:225:14: error: conflicting types for 'getenv'
  225 | extern char *getenv (const char * name);
      |              ^
C:\Program Files (x86)\Windows Kits\10\Include\10.0.22621.0\ucrt\stdlib.h:1184:28: note: previous declaration is here
 1184 |     _DCRTIMP char* __cdecl getenv(
      | 
```

Which is defined as
```c
_Check_return_ _CRT_INSECURE_DEPRECATE(_dupenv_s)
_DCRTIMP char* __cdecl getenv(
    _In_z_ char const* _VarName
    );
```
in <stdlib.h>. 

I think the best way to handle this error is to remove the extern declaration when it is clang and MSVC.

src/libFLAC/CMakeLists.txt
Rt39, line 1: I think now it is only targeting when using Clang and MSVC headers, for MinGW headers contains unistd.h but MSVC ones does not. Adding cmake_policy is to suppress the warnings.

src/libFLAC/CMakeLists.txt
ktmf01: This still doesn't make any sense to me. Why would this only be necessary on Clang on Windows with MSVC headers? I'd rather understand this first. It seems like some logic is missing in https://github.com/xiph/flac/blob/master/src/libFLAC/include/private/cpu.h

ktmf01: Apparently this is a bug in ClangCL: https://github.com/llvm/llvm-project/issues/53520

PR #721, opened 2024-07-10
Remove --explain option from flac command line program

### LINKED ISSUE #699: "--explain" does not state what can be maximal allowed value for "--max-lpc-order"
It only states that: "Max LPC order; 0 => only fixed predictors. Must be <= 12 for Subset streams if sampleate is <=48kHz"
No mention of maximal value 32.

Maximal value for --qlp-coeff-precision is not stated too.

Also description for --rice-partition-order seems to be a bit incorrect: "<...> the default is -r 0 <...>".
But "the default" is -r 5 actually.

FLAC 1.4.3

ktmf01:
The `flac` command had 4 levels of documentation: short help, help, explain and manpage. The explain is removed in favour of referring to the much more comprehensive manpage.

Closes #699 as the `--explain` option is removed.


ktmf01:
@Rollinnn and @H2Swine, any more remarks on this?

H2Swine:
I agree that `--explain`  was most sensible to remove. 

Will the man page source stay in the same .md format? Given time I might give it a shot to try to update it at the end of the summer.

ktmf01:
Yes, the markdown files are considered source files, and are converted to troff (manpage format) and html.

Issue #661, opened 2023-12-18
Samples count is always zero in write callback for Ogg FLAC


njlarsson:
The documentation of `FLAC__StreamEncoderWriteCallback` says “_Unlike when writing to native FLAC, when writing to Ogg FLAC the write callback will be called twice when writing each audio frame; once for the page header, and once for the page body. When writing the page header, the samples argument to the write callback will be 0._” Unfortunately, it appears that the samples argument is **always** 0 with Ogg FLAC.

The code below demonstrates the problem, encoding a buffer of 100 zero samples with native and Ogg FLAC. It prints (where both should be 100):

```
Total samples output without ogg: 100
Total samples output with ogg: 0
```

I tested it with the same results with FLAC 1.4.3 installed via Homebrew in macOS, and with 1.3.3 on Ubuntu (1.3.3-2ubuntu0.2, the version that standard package management on Ubuntu 22.04.3 LTS supplies for package  libflac-dev).

This breaks the use of Ogg FLAC in my application, since I need to determine the number of samples actually output in order not to flood the recipient of the stream. (The number of output samples is sometimes much smaller than the number of processed samples for highly compressible audio.)

```
#include <stdlib.h>
#include <stdio.h>
#include "FLAC/stream_encoder.h"

long outsamples;

static FLAC__StreamEncoderWriteStatus
write_cb(const FLAC__StreamEncoder *encoder,
         const FLAC__byte buffer[],
         size_t bytes,
         uint32_t samples,
         uint32_t current_frame,
         void *handle)
{
    outsamples += samples;
    return FLAC__STREAM_ENCODER_WRITE_STATUS_OK;
}

int main(int argc, char *argv[]) {
    const int samples = 100;
    FLAC__int32 **buffer;
    buffer = malloc(2 * sizeof(FLAC__int32 *));
    if (buffer == NULL) {
        fprintf(stderr, "Failed to allocate buffer");
        exit(1);
    }
    for (int c = 0; c < 2; c++) {
        if ((buffer[c] = malloc(samples * sizeof **buffer)) == NULL) {
            fprintf(stderr, "Failed to allocate buffer");
            exit(1);
        }
        for (int i = 0; i < samples; i++) {
            buffer[c][i] = 0;
        }
    }

    for (int ogg = 0; ogg < 2; ogg++) {
        FLAC__StreamEncoder *encoder;
        if ((encoder = FLAC__stream_encoder_new()) == NULL) {
[...]

PR #731, opened 2024-08-07
Don't just pass 0 for samples in Ogg/FLAC callback

njlarsson:
This is a fix for issue #661 where I motivated it in a comment, but the comments in the code are actually probably enough to explain it.

### DIFF
--- a/include/FLAC/stream_encoder.h
+++ b/include/FLAC/stream_encoder.h
@@ -538,11 +538,16 @@ typedef FLAC__StreamEncoderReadStatus (*FLAC__StreamEncoderReadCallback)(const F
  *  callback is being called to write metadata.
  *
  * \note
- * Unlike when writing to native FLAC, when writing to Ogg FLAC the
- * write callback will be called twice when writing each audio
- * frame; once for the page header, and once for the page body.
- * When writing the page header, the \a samples argument to the
- * write callback will be \c 0.
+ * Unlike when writing to native FLAC, when writing to Ogg FLAC the write
+ * callback will be called at least twice when writing each audio frame; once
+ * for the page header and once for the page body, possibly repeating this
+ * pair of calls several times in a batch with the same value of
+ * \a current_frame.  When writing the page header, as well as in all but the
+ * first page body write of the batch, the \a samples argument to the write
+ * callback will be \c 0. The full number of samples of the batch will be
+ * passed in the first page body write. Furthermore, it is possible that a few
+ * of these samples are retained in an internal Ogg encoding buffer and not
+ * actually encoded until the next batch.
  *
  * \note In general, FLAC__StreamEncoder functions which change the
  * state should not be called on the \a encoder while in the callback.
--- a/src/libFLAC/include/private/ogg_encoder_aspect.h
+++ b/src/libFLAC/include/private/ogg_encoder_aspect.h
@@ -49,6 +49,7 @@ typedef struct FLAC__OggEncoderAspect {
 	FLAC__bool seen_magic; /* true if we've seen the fLaC magic in the write callback yet */
 	FLAC__bool is_first_packet;
 	FLAC__uint64 samples_written;
+	uint32_t samples_in_submit_buffer;
 } FLAC__OggEncoderAspect;
 
 void FLAC__ogg_encoder_aspect_set_serial_number(FLAC__OggEncoderAspect *aspect, long value);
--- a/src/libFLAC/ogg_encoder_aspect.c
+++ b/src/libFLAC/ogg_encoder_aspect.c
@@ -57,6 +57,7 @@ FLAC__bool FLAC__ogg_encoder_aspect_init(FLAC__OggEncoderAspect *aspect)
 	aspect->seen_magic = false;
 	aspect->is_first_packet = true;
 	aspect->samples_written = 0;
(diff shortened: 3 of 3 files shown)


ktmf01:
I see a problem here when very small frames are generated by libFLAC, for example on silence. In that case the write callback isn't called at all, the frame is kept in memory. It seems fixable though.

I suggest adding a variable to the FLAC__OggEncoderAspect struct (defined in ogg_encoder_aspect.h) and then doing the following (I haven't tested this, I assume you have a good way to test this as you actually use this code)

```
			/* For a batch of write_callback calls, pass the number of samples since the last batch in the first non-metadata
			 * page body call, and then set to zero in case there are more iterations of the while loop (so
			 * as not to give the impression of more samples being processed).
			 */
			aspect->samples_in_submit_buffer += samples;
			while(ogg_stream_pageout(&aspect->stream_state, &aspect->page) != 0) {
				if(write_callback(encoder, aspect->page.header, aspect->page.header_len, 0, current_frame, client_data) != FLAC__STREAM_ENCODER_WRITE_STATUS_OK)
					return FLAC__STREAM_ENCODER_WRITE_STATUS_FATAL_ERROR;
				if(write_callback(encoder, aspect->page.body, aspect->page.body_len, pass_samples, current_frame, client_data) != FLAC__STREAM_ENCODER_WRITE_STATUS_OK)
					return FLAC__STREAM_ENCODER_WRITE_STATUS_FATAL_ERROR;
				aspect->samples_in_submit_buffer = 0;
			}
```

If you have a better name for the new variable, feel free to change it. Thanks for looking into this, I couldn't find the time to work on this.

njlarsson:
@ktmf01 Yes, I see how that could be a potential problem, and you suggestion should work (assuming that you meant to replace pass_samples with aspect->samples_in_submit_buffer for the write_callback parameter as well), although handling the case of encoding silence is precisely my reason for looking at the samples parameter in the write callback (I wrote [a blog post](https://klipspringer.avadeaux.net/the-size-of-silence-and-other-quirks-in-streaming-flac-to-browsers/) about it), and it didn’t happen in that case.

I have committed a change based on your suggestion. It works fine in my context. Please tell me if there’s something more I should do – some specific test or something.

ktmf01:
> and it didn’t happen in that case.

Ok. Still, as libogg is an external component we should follow the described behaviour (which is that it is possible a packet is held back) in this case, and not rely on observed behaviour.

> I have committed a change based on your suggestion. It works fine in my context. Please tell me if there’s something more I should do – some specific test or something.

I think it is fine. I will take another look at it next week and merge it if I don't think of anything else that needs improvement.

Once again, thanks for picking this up.

njlarsson:
I noticed another thing when making the change: `FLAC__ogg_encoder_aspect_finish` just clears the stream. Shouldn’t it call `ogg_stream_flush` and possibly the write callback in case there is a packet held back? This is a separate issue, though, which should probably be handled separately.

However, when looking into this I also realized that if I understand correctly, the `is_metadata` case might also involve writing actual samples (again in the case where a packet has been held back), because it calls `ogg_stream_flush`. Therefore, I moved the `aspect->samples_in_submit_buffer` logic to affect both branches of the if statement.

ktmf01:
I just read youe blog entry, are you familiar with the 'limit minimum bitrate' function? See https://www.xiph.org/flac/api/group__flac__stream__encoder.html#gac8c5f361b441d528b7a6791b66bb9d40

I'll review your change next week

njlarsson:
Thanks, yes I had failed to noticed the FLAC__stream_encoder_set_limit_min_bitrate function. I don’t currently seem to need it, but I’ll think about incorporating it.

njlarsson:
In any case, I think the last commit makes the execution more logical and reduces coupling with the calling order. So are you going to merge it?

ktmf01:
Yes, I'd like to merge it, with "squash and merge". However, I just noticed that the email address you used for your commits is not the same as the address you use for github. Does it matter to you under which email address the commit is credited?

njlarsson:
No big deal, but the one I use for commits is my general “repository address” also outside of github, and more likely to remain “forever” than the alias I created for the github account.

njlarsson:
Sorry if I’m confused, but is the change supposed to be merged to master on this repo now? It doesn’t seem to be. Maybe I made a mistake when I created the PR: I was confused about what was left vs right in the Github web interface and may have switched the branches around.

ktmf01:
The problem is, I prefer to have commits in a PR squashed if there is some back-and-forth or when there are fixups. Usually I just press 'squash and merge' to get the whole PR into one commit, but in that case github takes the github email address for the commit, and not the email address in the commits themselves.

So, yesterday, I squashed your commits by hand, and let the checks run again. Not because it is necessary, but because I prefer it that way. I just merged that squashed commit with 'rebase and merge'.

njlarsson:
Ok, great, I like squashing too. It was unclear to me that this was what was happening in the Github interface. (The more I learn to love Git, the less I tend to like Github.)

ktmf01:
Yeah, I agree that github feels a little crude if you are familiar with the finer points of git, like `cherry-pick` or `rebase -i`. Still, I prefer the way of organising discussions under PRs to mailing-list patch submission, and the checks (especially CI-fuzz) is very valuable.

PR #735, opened 2024-08-28
Decode chained ogg streams, attempt #3

ktmf01:
Here's my attempt to implement decoding of chained ogg streams, based on #667. This still needs work squashing bugs, but as far as I know all features are present. This includes additions to the test suite, MD5 verification and seeking.


ktmf01:
@philippe44 @whatamesss @ziplantil @jprjr as you all showed interest in this feature in some form, at some point, could you perhaps take a look at this?

whatamesss:
i just noticed some odd behaviour...while mpd dies after the current song, mpv keeps playing but retains stale metadata

ktmf01:
Decoding of chained streams is turned off by default. This is because users of libFLAC (programs to play music with for example) haven't been equipped to deal with this new behaviour yet. This might explain the odd behaviour you are seeing.

So, this PR equips libFLAC to handle chained streams. However, the programs need to implement this too, and that is why this is turned off by default.

whatamesss:
it would be nice if this was a config time option at least, that way on gentoo it could be set by a USE flag..users needing to patch the source would seriously limit the reach of this feature..

as for the behaviour, i dont think it's my player (mpd) as it works with phlippe44's code

ziplantil:
I do not think it should be a compile-time flag. Applications should have to explicitly enable the feature when they're confident they can support it.

ktmf01:
@whatamesss I introduced a new decoder state and added a few new functions. These are not yet handled and used by mpd. mpv seems to use ffmpeg and not libFLAC.

ktmf01:
@whatamesss I assume this is mpd? As said before, players need to be slightly modified to work with chained streams. libFLAC does the heavy lifting, but this PR needs some 'cooperation' so to say. So, you can't expect to plug it in and work correctly without any changes.

whatamesss:
hello @ktmf01 oddly, gstreamer works fine here...can we get input from @MaxKellermann? i'm not sure how to get this support into mpd as this branch is not merged..

ktmf01:
I just found out that seeking in chained multiplexed streams is terribly broken, and needs to be fixed. Note to self: the current way to determine the location in the bitstream needs to be replaced with something better. Perhaps a combination of whatever the tell callback returns, information about what is left in the buffer, how much of the current page has been used etc.

ktmf01:
Hi all,

I'm planning to merge this very soon, probably tomorrow if nothing comes up. If you have objections, please let me know.

whatamesss:
this is excellent news! congrats @ktmf01

PR #743, opened 2024-09-20
use ogg_page_granulepos to get number of samples written for Ogg FLAC

ziplantil:
The granule position of the Ogg page will tell us how many audio frames are included in the page that we receive from libogg. Therefore, to count the number of frames per page, we simply check the granule position and subtract it from the granule position of the previous page.

A signed 64-bit integer type is used for the granule position, as libogg also uses one as the return type of `ogg_page_granulepos`. An unsigned 32-bit integer is sufficient to contain the number of frames, as the granule position of a page is determined by the last complete packet in that page, the number of frames in each packet is limited by the 32-bit unsigned integer type internally in libflac, and an Ogg page, at most 65,307 bytes, can only contain at most ~20 million frames (tested experimentally with pure silence).

### DIFF
--- a/include/FLAC/stream_encoder.h
+++ b/include/FLAC/stream_encoder.h
@@ -544,10 +544,11 @@ typedef FLAC__StreamEncoderReadStatus (*FLAC__StreamEncoderReadCallback)(const F
  * pair of calls several times in a batch with the same value of
  * \a current_frame.  When writing the page header, as well as in all but the
  * first page body write of the batch, the \a samples argument to the write
- * callback will be \c 0. The full number of samples of the batch will be
- * passed in the first page body write. Furthermore, it is possible that a few
- * of these samples are retained in an internal Ogg encoding buffer and not
- * actually encoded until the next batch.
+ * callback will be \c 0. For the write callback call containing the first
+ * page body, the \a samples argument is the number of samples contained in
+ * all newly added complete packets (not pages). This means that in case a
+ * packet is split over two pages, they are counted in the samples argument
+ * of the page on which the packet is completed.
  *
  * \note In general, FLAC__StreamEncoder functions which change the
  * state should not be called on the \a encoder while in the callback.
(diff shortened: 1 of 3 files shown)


ktmf01:
I don't think this works when granulepos is UINT64_MAX (which is when a page only contains no packet ends). Also, I don't see the advance of doing it this way? Finally, I think the use of the word 'frames' is confusing, it usually means something else in FLAC context.

ziplantil:
> I don't think this works when granulepos is UINT64_MAX (which is when a page only contains no packet ends).

That is a relatively easy thing to fix.

> I think the use of the word 'frames' is confusing, it usually means something else in FLAC context.

Good point, it could just be changed to samples instead.

> Also, I don't see the advance of doing it this way?

It is the more 'correct' approach, because this actually checks how many samples are supposed to be present in each Ogg page.

However, reading the code again, it might indeed not actually be necessary, since we are actually only feeding one packet per each call before we loop over the pages that can be output.

Therefore, this implementation and the previous one will be equivalent when the packets are small enough to fit within a page (i.e. when we are not submitting more than one page for each packet). With large packets, if we fix the -1 case, the only difference will be _when_ the correct number of samples for that packet can actually be provided (for the first page in the old implementation, or for the last page in this one).

ktmf01:
If I understand this correctly, the main difference is in when samples are counted towards a packet: previously samples in an incomplete packet were counted in the first page containing that packet, and with this change, samples in an incomplete packet are counted in the last page contained that packet.

Could you take a look at the commit I added?

ziplantil:
The commit looks good to me.

ktmf01:
@ziplantil I see you're using an email address to sign your commits with, that is hidden in github. Therefore, when I squash your commits, they get signed with a @users.noreply.github.com email address.

Do you have a preference? If you'd rather have the squashed commit signed with your regular email, I will squash these commits manually, no problem.

ziplantil:
I do not have a preference - whichever is easier for you.

PR #759, opened 2024-11-11
Help text: corrected raw options

H2Swine:
As pointed out in #708 (after closed), a raw option came out wrong. Trying to get it right without spending too many lines.
This proposal doesn't say what is mandatory, but "everything is mandatory when indicated". 
Alternative line 1413: 
`	printf("raw format options (no default values provided):\n");
`

### DIFF
--- a/src/flac/main.c
+++ b/src/flac/main.c
@@ -1410,14 +1410,13 @@ void show_help(void)
 	printf("      --force-aiff-c-none-format     Decode to AIFF-C NONE format\n");
 	printf("      --force-aiff-c-sowt-format     Decode to AIFF-C sowt format\n");
 	printf("      --force-raw-format             Treat input or output as raw samples\n");
-	printf("raw format options: (all options mandatory for encoding from raw input,\n");
-	printf("                   --sign and --endian are mandatory for decoding to raw output)\n");
-	printf("      --sign={signed|unsigned}       Sign of samples\n");
-	printf("      --endian={big|little}          Byte order for samples\n");
+	printf("raw format options:\n");
+	printf("      --sign={signed|unsigned}       Sign of samples (input/output) \n");
+	printf("      --endian={big|little}          Byte order for samples (input/output)\n");
 	printf("      --channels=#                   Number of channels in raw input\n");
 	printf("      --bps=#                        Number of bits per sample in raw input\n");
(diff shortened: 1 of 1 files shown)


ktmf01:
Thanks!

ktmf01:
BTW, would you mind 'propagating' the same clarifications to the flac.md document?

H2Swine:
It was actually in reading flac.md and comparing to the help text, I found that this came out wrong. So ... not this week, but hope to do. 

Looking over it for a minute right now, ... should the help text also explain `-- `? Musing aloud, the first or last of these additions:

```
Usage:

 Encoding: flac [<general/encoding/format options>] [INPUTFILE [...]]
 Decoding: flac -d [<general/decoding/format options>] [FLACFILE [...]]
  Testing: flac -t [<general options>] [FLACFILE [...]]
Analyzing: flac -a [<general/analysis options>] [FLACFILE [...]]

Use the shell's quotation marks for filenames with special characters.
Input files starting with dash must appear last and be preceded by --  .

This help text summarizes all available options, for more explanation and
examples please consult the manual. This manual is often distributed
alongside the program as a man page or an HTML file. It can also be found
online at https://xiph.org/flac/documentation_tools_flac.html

General options:
  -- (double dash whitespace)  End of options; only input filename(s) can 
                           follow. Needed when they start with a dash.
  -v, --version                Show the flac version number
```

Against: more spam before you get to the substance.  Which is also the reason why I don't mention special characters in anything but filenames; say powershell needs subdivide_tukey(3) to be quoted, but that's not for beginners.

ktmf01:
Perhaps best to keep the help file short. Usage of `--` is rather common in *nix environments if I'm not mistaken, it is actually not really FLAC-specific behaviour but something provided by the getopt function. As far as I know, it is only really useful if you have a filename that starts with `-` but doesn't otherwise needs to be escaped. So, I'd say leave it out of the help text.

Anyway, it would be much appreciated if you could take a look at it at some point. I'd really like to focus on finishing the Ogg chaining stuff and at some point move to a release. I'm not sure when yet, but at some point I'll have to decide to stop adding API changes and start 'saving them up' for FLAC 1.6.x. On the other hand, I'd like to have a few 1.5.x releases which would mean I cannot do any API changes for a while. So, might take a month, might take a few months. I honestly don't think I'll get 1.5.x out in 2024, but maybe I'll surprise myself.

PR #769, opened 2024-12-03
Update flac.md: Reordering and updating options descriptions & more

H2Swine:
I am not sure about the organization of the initial part of the document, to be honest. For example the way it promises examples before options, but before the examples sections it gives examples (on using stdin and redirects).

Anyway, here are the most significant changes made: 
- Some in the examples. Less confident about what should be there. 
- Options are reorganized to match the flac -h order 
- I used _underscores_ for italics, avoiding some formatting issue that arose from triple asterisks 
- Some " " inserted - you will all too easily need them when using -T - but I dared not use too many of them. 

... and some more text tweaks.

### DIFF
--- a/man/flac.md
+++ b/man/flac.md
@@ -125,33 +125,27 @@ Some common **encoding** tasks using flac:
 :	Like above, except abc.wav is deleted if there were no errors.
 
 `flac --delete-input-file -w abc.wav`
-:	Like above, except abc.wav is deleted if there were no errors or warnings.
+:	Like above, except abc.wav is deleted if there were no errors and no warnings.
 
-`flac --best abc.wav`
-:	Encode abc.wav to abc.flac using the highest compression setting.
+`flac --best abc.wav` or `flac -8 abc.wav`
+:	Encode abc.wav to abc.flac using the highest compression preset.  
 
-`flac --verify abc.wav`
+`flac --verify abc.wav` or `flac -V abc.wav`
 :	Encode abc.wav to abc.flac and internally decode abc.flac to make sure it matches abc.wav.
 
 `flac -o my.flac abc.wav`
 :	Encode abc.wav to my.flac.
 
 `flac -T "TITLE=Bohemian Rhapsody" -T "ARTIST=Queen" abc.wav`
-:	Encode abc.wav and add some tags at the same time to abc.flac.
+:	Encode abc.wav and add some tags at the same time, to abc.flac.
 
-`flac *.wav`
-:	Encode all .wav files in the current directory.
+`flac abc.aiff foo.rf64 bar.w64`
+:	Encode abc.aiff to abc.flac, foo.rf64 to foo.flac and bar.w64 to bar.flac
 
-`flac abc.aiff`
-:	Encode abc.aiff to abc.flac.
+`flac *.wav *.aif`
+:	Encode all .wav files and all .aif files in the current directory.
 
-`flac abc.rf64`
-:	Encode abc.rf64 to abc.flac.
-
-`flac abc.w64`
-:	Encode abc.w64 to abc.flac.
-
-`flac abc.flac --force`
+`flac abc.flac --force` or `flac abc.flac -f`
 :	This one's a little tricky: notice that flac is in encode mode by
 	default (you have to specify -d to decode) so this command actually
 	recompresses abc.flac back to abc.flac. \--force is needed to make
@@ -159,30 +153,41 @@ Some common **encoding** tasks using flac:
 	would you want to do this? It allows you to recompress an existing
 	FLAC file with (usually) higher compression options or a newer
 	version of FLAC and preserve all the metadata like tags too.
+ 	Note: If the FLAC file does not end with .flac - say, it is abc.fla
+  	- then a new abc.flac will be created and the old kept, just like 
+   	for an uncompressed input file.
+
+`flac -8Vfo Track04.flac -T "ARTIST=Queen" -- -4.wav`
+:	Options in their short version - those invoked with single dash - 
+	can be concatenated until one that takes an argument: here -o and 
+ 	-T do, and so one of them requires its own dash. 
+	This command line takes the track -4.wav as input; its filename 
+ 	starts with a dash, so it needs the \-- to keep the encoder from
+  	taking the "4" as an option. All options must go before the --: 
+        It will overwrite any Track04.flac already present, encode with
+	highest compression preset, decode to verify that it matches, and 
+ 	a tag will be added. 
+
 
 Some common **decoding** tasks using flac:
 
-`flac -d abc.flac`
+`flac --decode abc.flac` or `flac -d abc.flac`
 :	Decode abc.flac to abc.wav. abc.flac is not deleted. NOTE: Without 
-	-d it means re-encode abc.flac to abc.flac (see above).
+	-d it would mean to re-encode abc.flac to abc.flac (see above) and 
+ 	would err out because abc.flac already exists.
 
-`flac -d --force-aiff-format abc.flac`  
-
-`flac -d -o abc.aiff abc.flac`
+`flac -d --force-aiff-format abc.flac` or `flac -d -o abc.aiff abc.flac`
 :	Two different ways of decoding abc.flac to abc.aiff (AIFF format).
-	abc.flac is not deleted.
+	abc.flac is not deleted. -d -o could be concatenated to -do.
+ 	The decoder can force other output formats, see the options below.
 
-`flac -d --force-rf64-format abc.flac`  
-
-`flac -d -o abc.rf64 abc.flac`
-:	Two different ways of decoding abc.flac to abc.rf64 (RF64 format).
-	abc.flac is not deleted.
-
-`flac -d --force-wave64-format abc.flac`  
-
-`flac -d -o abc.w64 abc.flac`
-:	Two different ways of decoding abc.flac to abc.w64 (Wave64 format).
-	abc.flac is not deleted.
+`flac -d --keep-foreign-metadata-if-present abc.flac`
+:	FLAC files can store non-audio chunks of input WAVE/AIFF/RF64/W64
+	files. If there are such chunks stored, the decoder will select
+ 	output format accordingly and restore both audio and non-audio.
+  	If there are no such chunks stored, it will decode to abc.wav.
+   	The related option \--keep-foreign-metadata will instead exit 
+    	with an error if no such non-audio chunks are found.
 
 `flac -d -F abc.flac`
 :	Decode abc.flac to abc.wav and don't abort if errors are found
@@ -192,18 +197,16 @@ Some common **decoding** tasks using flac:
 # OPTIONS
 
 A summary of options is included below. For a complete description, see
-the HTML documentation.
+the HTML documentation. Several of the options can be negated, see the 
+**Negative options** section below.
 
 ## GENERAL OPTIONS
 
 **-v, \--version**
-:	Show the flac version number
+:	Show the flac version number, and quit.
 
 **-h, \--help**
-:	Show basic usage and a list of all options
-
-**-H, \--explain**
-:	Show detailed explanation of usage and all options
+:	Show basic usage and a list of all options, and quit.
 
 **-d, \--decode**
 :	Decode (the default behavior is to encode)
@@ -220,45 +223,29 @@ the HTML documentation.
 **-c, \--stdout**
 :	Write output to stdout
 
-**-s, \--silent**
-:	Silent mode (do not write runtime encode/decode statistics to stderr)
-
-**\--totally-silent**
-: Do not print anything of any kind, including warnings or errors. The
-	exit code will be the only way to determine successful completion.
-
-**\--no-utf8-convert**
-:	Do not convert tags from local charset to UTF-8. This is useful for
-	scripts, and setting tags in situations where the locale is wrong.
-	This option must appear before any tag options!
-
-**-w, \--warnings-as-errors**
-:	Treat all warnings as errors (which cause flac to terminate with a
-	non-zero exit code).
-
 **-f, \--force**
 :	Force overwriting of output files. By default, flac warns that the
 	output file already exists and continues to the next file.
-
-**-o** *filename***, \--output-name=***filename*
-:	Force the output file name (usually flac just changes the extension).
-	May only be used when encoding a single file. May not be used in
-	conjunction with \--output-prefix.
-
-**\--output-prefix=***string*
-:	Prefix each output file name with the given string. This can be
-	useful for encoding or decoding files to a different directory. Make
-	sure if your string is a path name that it ends with a trailing \`/'
-	(slash).
 
 **\--delete-input-file**
 :	Automatically delete the input file after a successful encode or
 	decode. If there was an error (including a verify error) the input
 	file is left intact.
 
+**-o** _filename_, **\--output-name=** _filename_
+:	Force the output file name (usually flac just changes the extension).
+	May only be used when encoding a single file. May not be used in
+	conjunction with \--output-prefix.
+
+**\--output-prefix=** _string_
+:	Prefix each output file name with the given string. This can be
+	useful for encoding or decoding files to a different directory. Make
+	sure if your string is a path name that it ends with a trailing \`/'
+	(slash).
+
 **\--preserve-modtime**
-:	Output files have their timestamps/permissions set to match those of
-	their inputs (this is default). Use \--no-preserve-modtime to make
+:	(default!) Output files have their timestamps/permissions set to 
+	match those of their inputs. Use \--no-preserve-modtime to make
 	output files have the current time and default permissions.
 
 **\--keep-foreign-metadata**
@@ -268,20 +255,20 @@ the HTML documentation.
 	transcoded, e.g. WAVE chunks saved in a FLAC file cannot be restored
 	when decoding to AIFF. Input and output must be regular files (not
 	stdin or stdout). With this option, FLAC will pick the right output
-	format on decoding.
+	format on decoding. It will exit with error if no such chunks are found.
 
 **\--keep-foreign-metadata-if-present**
 :	Like \--keep-foreign-metadata, but without throwing an error if
-	foreign metadata cannot be found or restored, instead printing a 
+	foreign metadata cannot be found or restored.  Instead, prints a 
 	warning.
 
-**\--skip={***\#***\|***mm:ss.ss***}**
+**\--skip={**_\#_**\|**_mm:ss.ss_**}**
 :	Skip over the first number of samples of the input. This works for
 	both encoding and decoding, but not testing. The alternative form
 	mm:ss.ss can be used to specify minutes, seconds, and fractions of a
 	second.
 
-**\--until={***\#***\|\[***+***\|***-***\]***mm:ss.ss***}**
+**\--until={**_\#_**\|\[**_+_**\|**_-_**\]**_mm:ss.ss_**}**
 :	Stop at the given sample number for each input file. This works for
 	both encoding and decoding, but not testing. The given sample number
 	is not included in the decoded output. The alternative form mm:ss.ss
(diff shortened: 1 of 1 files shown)


ktmf01:
Thanks! I'll take a look.

man/flac.md
```diff
@@ -125,64 +125,69 @@ Some common **encoding** tasks using flac:
...
 
 Some common **decoding** tasks using flac:
 
-`flac -d abc.flac`
+`flac --decode abc.flac` or `flac -d abc.flac`
 :	Decode abc.flac to abc.wav. abc.flac is not deleted. NOTE: Without 
-	-d it means re-encode abc.flac to abc.flac (see above).
-
-`flac -d --force-aiff-format abc.flac`  
+	-d it would mean to re-encode abc.flac to abc.flac (see above) and
```
ktmf01, line 176: This is a little inconsistent: you've changed `-d`to `--decode` in the command, but not the explanation

man/flac.md
```diff
@@ -125,64 +125,69 @@ Some common **encoding** tasks using flac:
...
-	abc.flac is not deleted.
-
-`flac -d --force-rf64-format abc.flac`  
-
-`flac -d -o abc.rf64 abc.flac`
-:	Two different ways of decoding abc.flac to abc.rf64 (RF64 format).
-	abc.flac is not deleted.
-
-`flac -d --force-wave64-format abc.flac`  
+	abc.flac is not deleted. -d -o could be concatenated to -do.
```
ktmf01, line 181: Maybe use 'abbreviated' instead of 'concatenated'? Or shortened?

man/flac.md
```diff
@@ -220,45 +223,29 @@ the HTML documentation.
...
 
-**\--delete-input-file**
-:	Automatically delete the input file after a successful encode or
-	decode. If there was an error (including a verify error) the input
-	file is left intact.
-
 **\--preserve-modtime**
-:	Output files have their timestamps/permissions set to match those of
-	their inputs (this is default). Use \--no-preserve-modtime to make
+:	(default!) Output files have their timestamps/permissions set to
```
ktmf01, line 247: Was this change intentional or were you planning on improving this later? I think '(default!)' is somewhat ambiguous.

man/flac.md
```diff
@@ -360,14 +408,27 @@ the HTML documentation.
 	option in combination with custom encoding options meant for
 	archival.
 
-**-j** *\#***, \--threads=***\#*  
+**\--limit-min-bitrate**
```
ktmf01, line 411: Maybe we can reuse the text from the removed `-H` options

> Limit minimum bitrate by not allowing frames consisting of only constant  subframes. This ensures a bitrate of at least 1 bit/sample, for example 48kbit/s for 48kHz input. This is mostly beneficial for internet streaming.

ktmf01:
Thanks! Looks good. Please consider my comments.

One more thing:

>     * I used _underscores_ for italics, avoiding some formatting issue that arose from triple asterisks

So, previously, it didn't render correctly in github but did render correctly in pandoc (html and man page). With this change, it renders correctly in github but not in pandoc. I think pandoc is more important, so please revert this change.

H2Swine:
A few brief questions, reverse order, before I give it a go: 
- Underscores etc. Does pandoc honour spacing in between the asterisks? (Or anything escaped?)
- "default": it was intentional to indicate. Is it the too-terse "(default!)" you have issues with, so that one could write it out to e.g. "(This option on by default.)"?
- I have no strong feelings for "concatenated". Unless you think "joined" is better (describing _what_ you do, "shortened" describes _why_) I'm going for "shortened".

ktmf01:
>     * Underscores etc. Does pandoc honour spacing in between the asterisks? (Or anything escaped?) Edit: Maybe more important: Would those cause harm?

Sorry, I don't understand your question.

>     * "default": it was intentional to indicate. Is it the too-terse "(default!)" you have issues with, so that one could write it out to e.g. "(This option on by default.)"?

I think it is too terse, yes. I'd prefer (This option is on by default). I'd say, in a man page there is room enough to not be overly concise.

>     * I have no strong feelings for "concatenated". Unless you think "joined" is better (describing _what_ you do, "shortened" describes _why_) I'm going for "shortened".

Then shortened it is!

H2Swine:
> > Underscores etc. Does pandoc honour spacing in between the asterisks? (Or anything escaped?) Edit: Maybe more important: Would those cause harm?
> 
> Sorry, I don't understand your question.

Like this, where the left underscores are replaced by space+asterisk and right underscores by asterisk+space. Will that solve the problem? If the encoder/decoder objects to spaces here, it is likely a bad idea. 

**\--skip={** *\#* **\|** *mm:ss.ss* **}**

H2Swine:
I did a new attempt locally at https://github.com/H2Swine/flac/commit/7125fa974a7ecc70ea40f2a6558ef041a6815b37 , but before I figure out how to update a PR with a new file (or should one open a new?), you can rather look at it to strike down whatever proposed changes are not that good. That in particular goes for the attempts at fixing the previous reply in this thread. 

There are a couple of more changes proposed. 
- You see several lines changed by replacing double spacing between sentences (dot space space by dot space). Of course, the opposite could be done, but I went by majority vote.
- Then I made an attempt to rewrite the apodization function section, that obviously had had "text added over time" rather than trying to write it as one section. Not sure if that text is good enough, but content-wise: I honestly think you did such a good job on subdivide_tukey that once the reader is past (1) _this is what the option is about_, and (2) _if you want to try, these are your options_, then go to (3) _if you want to try heavier than -8, this is the first suggestion_. Then afterwards explain what more can be fiddled around with. (You agree with my gut feeling that partial_tukey and subdivide_tukey are likely obsoleted?)
 -- I said something about escaping parentheses ... that is not solely for -A. But anyway, since subdivide_tukey uses parentheses, it could be said up there.
 -- You also see that I went all-out "use scientific notation" in that section. Two justifications: It is The Right Thing To Do to avoid locale-dependent interpretation, and: since it is now is at the end, after a recommended "power users' baby step: n", it shouldn't intimidate many who would want to try it.

Then not yet done:
- I saw afterwards that linebreak 471 isn't good, but it isn't the final draft anyway. Also ... hm, linebreaks in .md? I've tried to stick to the seventysomething character count.
- Raw options: yeah, channels interleaved - _but in what order?_ I think that information should be there, but I haven't tried to look it up.
- The overlap parameter ov in partial/punchout tukey: between -1 and 1, is that right? It does accept other values though, but "should be" is not as strong as saying "will only accept". 
- There is no longer the text that partial/punchout tukey from 2 to 6 was a "sane" range. Something like that could of course be added in - for subdivide_tukey, of course - but _maybe_ then one could mention not only the diminishing return, but that at some stage it gets more expensive than -p.

ktmf01:
> So even when sorting out a policy for CAPITALS or not, we might think of one out of two-and-a-half:
> 
>     * Triple asterisks that look bad in github.
> 
>     * Drop the italics, and make choices on how to have it typeset: rather than **--output-prefix=**_string_ , the we would have the choice between **--output-prefix=**string and **--output-prefix=**STRING .
> 
>     * Or maybe just drop some of the italics, where it doesn't do much anyway: **--blocksize=**# vs **--blocksize=**_#_ , does it matter that the latter _#_ is italics? And when there are braces, like --endian={big|little} , it is anyway given that it isn't supposed to be copied literally.

Lets get back to the beginning: this was and is, first and foremost, a manpage. It is a *nix concept that you might be familiar with. I'd like to stick with the convention there. The convention seems to be that everything a user has to change (i.e., that is not to be taken literal) is underlined and ALLCAPS. Now, markdown doesn't support underlining in any way, and manpages use a monospace font which doesn't support italics, so the underlined parts were converted to italics in the markdown/html.

Proposal: Don't apply any markup to symbols. So, `**\--output-name=***filename*` is converted to `**\--output-name**=*filename*`. That seems to work: **\--output-name**=*filename*



> but before I figure out how to update a PR with a new file (or should one open a new?), you can rather look at it to strike down whatever proposed changes are not that good. 

Just add another commit to the branch that belongs to this pull request.

> 
> There are a couple of more changes proposed.
> 
>     * You see several lines changed by replacing double spacing between sentences (dot space space by dot space). Of course, the opposite could be done, but I went by majority vote.

Yeah, I don't know where these came from, I think replacing with single spaces is fine.
 
>     * Then I made an attempt to rewrite the apodization function section, that obviously had had "text added over time" rather than trying to write it as one section. Not sure if that text is good enough, but content-wise: I honestly think you did such a good job on subdivide_tukey that once the reader is past (1) _this is what the option is about_, and (2) _if you want to try, these are your options_, then go to (3) _if you want to try heavier than -8, this is the first suggestion_. Then afterwards explain what more can be fiddled around with. (You agree with my gut feeling that partial_tukey and subdivide_tukey are likely obsoleted?)
>       -- I said something about escaping parentheses ... that is not solely for -A. But anyway, since subdivide_tukey uses parentheses, it could be said up there.
>       -- You also see that I went all-out "use scientific notation" in that section. Two justifications: It is The Right Thing To Do to avoid locale-dependent interpretation, and: since it is now is at the end, after a recommended "power users' baby step: n", it shouldn't intimidate many who would want to try it.
> 

Thanks

> 
> Then not yet done:
> 
>     * I saw afterwards that linebreak 471 isn't good, but it isn't the final draft anyway. Also ... hm, linebreaks in .md? I've tried to stick to the seventysomething character count.
> 
>     * Raw options: yeah, channels interleaved - _but in what order?_ I think that information should be there, but I haven't tried to look it up.
> 

In the order you want them to end up in the FLAC file. First channel first, second channel second etc. I don't think not adding this is a big deal, this is specific enough that people can just try and see what happens a few times if they don't know where to find the format specification. Raw input is messy anyway.

>     * The overlap parameter ov in partial/punchout tukey: between -1 and 1, is that right? It does accept other values though, but "should be" is not as strong as saying "will only accept".

No, between negative infinity and 1. I don't think it is necessary to document however, as I am not sure it is very useful. However, a negative overlap gives very small windows with gaps between them. So, when you specify `partial_tukey(2/0.5)` you get 2 windows that overlap somewhat: first goes from 0 to 66% while the second goes from 33% to 100%. At overlap 1, you get two windows that are the same. At overlap -3, the first window goes from 0% to 20%, while the second goes from 80% to 100%, so there's a gap in the middle. I am unsure whether it actually always works as intended. 

>     * There is no longer the text that partial/punchout tukey from 2 to 6 was a "sane" range. Something like that could of course be added in - for subdivide_tukey, of course - but _maybe_ then one could mention not only the diminishing return, but that at some stage it gets more expensive than -p.

I don't know, experimenting beyond the presets is really that, experimenting. I'm not sure there should be guidelines on that in the manpage. When someone gets to this point in the manual, everything already works anyway, it is not some feature that can be enabled or disabled or something, it is just that maybe, perhaps, a tiny bit smaller file can be generated.

H2Swine:
I'll have a look at the following shortly (object if I have misunderstood):

> The convention seems to be that everything a user has to change (i.e., that is not to be taken literal) is underlined and ALLCAPS. Now, markdown doesn't support underlining in any way, and manpages use a monospace font which doesn't support italics, so the underlined parts were converted to italics in the markdown/html.
> 
> Proposal: Don't apply any markup to symbols. So, `**\--output-name=***filename*` is converted to `**\--output-name**=*filename*`. That seems to work: **--output-name**=_filename_

I am going to presume it shall be in all caps, 
**--output-name**=*FILENAME*

> > * Raw options: yeah, channels interleaved - _but in what order?_ I think that information should be there, but I haven't tried to look it up.
> 
> In the order you want them to end up in the FLAC file. First channel first, second channel second etc. I don't think not adding this is a big deal, this is specific enough that people can just try and see what happens a few times if they don't know where to find the format specification. 

This is the comprehensive help, so I am putting in a reference to the format specification there. I assume it is with the order [as in the format spec section 9.1.3 table 16](https://datatracker.ietf.org/doc/html/draft-ietf-cellar-flac-14#section-9.1.3) - for example, the first one in the stream will go to "what the FLAC format by default (absent channel mask tag) thinks is" FL, the second to FR, and the third to FC except if channel count is 4?

> I don't know, experimenting beyond the presets is really that, experimenting. I'm not sure there should be guidelines on that in the manpage. 

I take that to mean that it is fine to _remove_ the wording about partial/punchout 2 to 6, as it could look like such a recommendation. I agree we should be careful not to present "more functions as any bang for the buck", but it may still be a point that subdivide_tukey is developed to recycle calculations for more windows cheaper. I'll chew on a formulation that you can reject.

H2Swine:
And this sentence ... is there supposed to be anything more comprehensive at all? Delete it?

"A summary of options is included below. For a complete description, see
the HTML documentation."

H2Swine:
I had another shot at it (did I commit it right between this comment and the previous?!). There are a few things I have not touched, not knowing completely what should be done to them all: 

- The items in the previous thirteenth reply: Is this supposed to take over for the HTML documentation? Cf. the text: 
"A summary of options is included below. For a complete description, see
the HTML documentation."

- Under ReplayGain specs, I find the following, which IMHO goes for the full man page. I don't think it has any business being there so far down in the document: either it is needed, and then move it up - or it isn't and can be removed? 
		The \<specification\> is a shorthand notation for describing how to	apply
		ReplayGain. All components are optional but order is important. '\[\]'
		means 'optional'. '\|' means 'or'. '{}' means required. The format is: 

- Brings me to: I have not started on the early part, which I think could need some more thought. Yes likely it should say something about how piping works and why -o etc are more preferred than redirects. And maybe about the previous item, explaining what brackets and braces and bars mean in the man page. And maybe about whether/when the "=" is compulsory (cf. previous comment).  And likely about the relationship between single-dash and double-dash options.

- Did very little about --apply-replaygain-which-is-not-lossless and the Picture spec, I haven't used those much, but: Do I remember right that using URL in PICTURE is discouraged nowadays?

ktmf01:
> I am going to presume it shall be in all caps, **--output-name**=_FILENAME_

Yes
> I assume it is with the order [as in the format spec section 9.1.3 table 16](https://datatracker.ietf.org/doc/html/draft-ietf-cellar-flac-14#section-9.1.3) - for example, the first one in the stream will go to "what the FLAC format by default (absent channel mask tag) thinks is" FL, the second to FR, and the third to FC except if channel count is 4?

I think so yes. As far as I know `flac` doesn't reorder channels, ever.
> I take that to mean that it is fine to _remove_ the wording about partial/punchout 2 to 6

Yes

> Edit: When it comes to characters in the help file: _When is the equals sign mandatory?_ Say, --threads 4 works just as well as --threads=4 .

Okay. This is behaviour of [getopt](https://en.wikipedia.org/wiki/Getopt), I don't know all the nooks and crannies of it. I suppose the equals sign isn't necessary then. Still, probably better not to mention it, or it gets even more confusing.
> And this sentence ... is there supposed to be anything more comprehensive at all? Delete it? "A summary of options is included below. For a complete description, see the HTML documentation."

You can remove that. Can you imagine, `flac` had 5 layers of docs: short help, help, explain, man page and HTML docs. I merged the HTML docs and man page in commit #80d064e but apparently I forgot this bit.

> I had another shot at it (did I commit it right between this comment and the previous?! Edit: Right now I discovered the "Convert to draft" option, which maybe would be the right thing for all that I know - I hope I didn't mess up anything by merely requesting review).

This seems to have happened exactly as it should be. Not using the 'draft PR' option is also the proper way, because on draft PRs, reviewing is impeded/blocked. I only use the draft PR option to signal that nobody should care to look at it yet, which isn't the case here.

> Under ReplayGain specs, I find the following, which IMHO goes for the full man page. I don't think it has any business being there so far down in the document: either it is needed, and then move it up - or it isn't and can be removed?
>       The <specification> is a shorthand notation for describing how to	apply
>       ReplayGain. All components are optional but order is important. '[]'
>       means 'optional'. '|' means 'or'. '{}' means required. The format is:

You could move it up, but people don't read such a document top-to-bottom. At least, I don't. I agree that it applies to the whole man page, but similar specs elsewhere aren't as complicated. Also, the pipe symbol is literal in the picture specification, whereas it means or elsewhere in the document.

> Brings me to: I have not started on the early part, which I think could need some more thought. Yes likely it should say something about how piping works and why -o etc are more preferred than redirects. And maybe about the previous item, explaining what brackets and braces and bars mean in the man page. And maybe about whether/when the "=" is compulsory (cf. previous comment).  And likely about the relationship between single-dash and double-dash options.

I think it is wise to merge this as soon as there are no loose ends, and not wait for further improvements. It is better to have several smaller commits than one big commit, as long as the commit is 'self-contained'.

> Did very little about --apply-replaygain-which-is-not-lossless and the Picture spec, I haven't used those much, but: Do I remember right that using URL in PICTURE is discouraged nowadays?
I don't know how widely supported such URLs are.

> For --skip and until, I did change the minutes:seconds to capitals, and gave example with decimals; I think it was unfortunate to suggest that fractions of a second was mandatory. But then ... I do think MM:SS is unfortunate too in that it suggests it is a compulsory two digits : two digits format. Surely there could be 123:4.567 (presuming dot as decimal point). So maybe it is better to write it out as _MINUTES:SECONDS_ or shortened to _MIN:SEC_ ?

MM:SS is a rather standard way of writing I'd say. I'm not sure that needs any clarification.

man/flac.md
```diff
@@ -125,140 +125,153 @@ Some common **encoding** tasks using flac:
...
 	FLAC file with (usually) higher compression options or a newer
 	version of FLAC and preserve all the metadata like tags too.
+ 	Note: If the FLAC file does not end with .flac - say, it is abc.fla
+	- then a new abc.flac will be created and the old kept, just like 
+	for an uncompressed input file.
+
+`flac --tag-from-file="ALBUM=albumtitle.txt" -T "ARTIST=Queen" *.wav`
+:	Encode every .wav file in the directory and add some tags. Every 
+	file will get the same set of tags.
+	Warning: Will wipe all existing tags, when when the input file is
```
ktmf01, line 164:
```suggestion
	Warning: Will wipe all existing tags, when the input file is 
```

H2Swine:
New one in a minute, hoping I'm about to converge on what I've actually touched upon.  Will describe changes relative to previous one here.


> I think it is wise to merge this as soon as there are no loose ends, and not wait for further improvements. It is better to have several smaller commits than one big commit, as long as the commit is 'self-contained'.

I agree. Although I did a little bit on lines 44 to (new) 56, but any changes between that and the EXAMPLES section can be ... for later. 


EXAMPLES: 

- Subsections for encoding examples and decoding examples. 
- Reworded the recompression example. Better than saying it is tricky, start by saying what it does. 
- nuked the "when when" you found.
- New "last encoding example". I don't think it is a "common" task for an end-user to add tags by hand, so I don't want to keep focus on it. 
The explanation is maybe elaborate on how options can be written, but maybe that part can be removed when one is done with the part above EXAMPLES.
- The -d -F: Here is a suggestion to be careful about it, and also - given what we know about old software throwing wrong data at libflac - to try different decoders and keep the offending files.
Not sure if it is a good wording. 


OPTIONS: 

First this removal I did: 

> You can remove that. Can you imagine, `flac` had 5 layers of docs: short help, help, explain, man page and HTML docs. I merged the HTML docs and man page in commit #80d064e but apparently I forgot this bit.

As an aside, I actually suggest to consider: -H should not give an error, rather a short sentence that says "For long help, use man page / HTML doc / [URL]." 
In case someone uses it, they better be told what to use instead. (In case nobody uses that one out of the five there were: no harm done.)

- For the -0 to -8, I inserted a "Currently" synonymous in the previous version, but for some strange reason I left out doing so for -4 and -5. Did it now.
- The --skip: Now clearer I think, what it must be. 
- The --picture: Italicized some ALLCAPS and minor rewording.
- ReplayGain subsection: Removed the description on brackets and pipe (and also on "="!). I did not reinsert it further up; in addition to what you wrote about Picture spec, I see that description of --cue, --skip etc have everything in brackets, so there is no use in writing outright that it is all optional when there has to be at least one. Better not write anything wrong. 
- Picture spec subsection: Italicized many ALLCAPS. Hope it comes out reasonable.
- Apodization: Minor rewording, reflecting that I am not completely happy with it. But I guess it is good enough.

ktmf01 (approved):

ktmf01:
I'd like to merge this now, if you are OK with that.

H2Swine:
Oh, sorry for one little detail I overlooked: at a glance and scroll, I found one single "n" shining at my eyes, that now is CapItalic-ized. 

Apart from that, I know that the .md file's linebreaks are not at consistent character width, but from your changes today I am assuming you have full control over what is needed, so ... merge if/when you want to. 

Afterwards I'll scratch my head a little for possible changes to the "GENERAL USAGE". (That is not to say I will actually come up with any improvement.)

PR #817, opened 2025-03-02
[CMake] Use CMake's find_package module to detect git

diizzyy:
Instead of enforcing execution of git if available use CMake's find_package module to make it optional.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,9 +48,9 @@ if(WITH_OGG)
     endif()
 endif()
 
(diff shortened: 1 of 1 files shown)


diizzyy:
Counterpart to what's being available using Autotools, https://github.com/xiph/flac/blob/master/configure.ac#L369

dg0yt:
IMO this should be complemented by replacing the use of literal `git` wiht `"${GIT_EXECUTABLE}"`. User may seed this cache variable without even having `git` in `PATH`.

PR #818, opened 2025-03-03
flac.md: fix "-q" misinformation, and more.

H2Swine:
Hope I haven't messed up how pulls work. Wouldn't be the first time. 

This intends to fix xiph#816 , the man page wrongly states -q 16 is permitted.

I have worked on more on the man page, but I don't want to put in too much in one request. 

- Also, reworded -p for consistency with -q; -m and -M for clarity; and --lax for brevity. 
- The following should have been done in the previous round: the --help text has the tagging options right after --picture. Did so here too. 
- But, and this should either also be fixed in the help text or be rejected: --no-utf-8 is a tagging option - aren't those encoding-only? Moved and reworded, but reject as appropriate.

### DIFF
--- a/man/flac.md
+++ b/man/flac.md
@@ -324,11 +324,6 @@ negated, see the **Negative options** section below.
 	sign means that the \--until point is taken relative to the \--skip 
 	point. For other considerations, see \--skip. 
 
-**\--no-utf8-convert**
-:	Do not convert tags from local charset to UTF-8. This is useful for
-	scripts, and setting tags in situations where the locale is wrong.
-	This option must appear before any tag options!
-
 **-s**, **\--silent**
 :	Silent mode (do not write runtime encode/decode statistics to stderr)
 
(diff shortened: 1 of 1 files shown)


man/flac.md
```diff
@@ -426,10 +421,11 @@ Encoding will default to -5, -A "tukey(5e-1)" and one CPU thread.
 	for subset streams with higher samplerates it must be \<= 16384.
 
 **-m**, **\--mid-side**
-:	Try mid-side coding for each frame (stereo only, otherwise ignored).
+:	Try mid-side coding for each frame in addition to left and right, and 
+	select the best compression. (Stereo only, ignored otherwise.)
 
 **-M**, **\--adaptive-mid-side**
-:	Adaptive mid-side coding for all frames (stereo only, otherwise ignored).
+:	Like -m, but adaptive choice (faster, slightly weaker compression).
```
ktmf01, line 428: Maybe use the word `heuristics` here? I think that is more descriptive than adaptive. I know, the word adaptive was there already, but it is rather unclear what it means I think

H2Swine: No objection to that. I'll look over it and - if you don't mind doing it in this thread - submit more textual suggestions. I have briefly tested picture and RG features and tried to make sense out of actual behaviour vs help text, so a question -- indeed two: 
- Could RG still leave PADDING even if --no-padding is used? I didn't get to provoke that. Looks like something similar was addressed in 1.2.1b? 
- General formatting: Should one more generally write out e.g. `-w` with ticks instead of -w without? I assume it would be a good thing to write **flac** double-asterisk-surrounded for the tool, to contrast it from the format (which likely should be FLAC in allcaps?), that is quite common manpage style?

H2Swine: > submit more textual suggestions. 

I tried my own test branch in order not to mess up too much. @ktmf01 , mind having a look at https://github.com/xiph/flac/commit/b359d00fe708bb8f433ba0ecdf68015a91c4695c#r159517147 where I commented the changes? 

... and if you think they are useful, educate me on how to link that one up to this PR. I'm _clumsy._

H2Swine: That draft nearly managed to clear up #840 if only "by collateral damage". To do when we get that far.

PR #822, opened 2025-03-16
Add C11 threads.h implementation in addition to pthread implementation for multithreaded encoding.

manxorist:
The switching between both APIs is done via macros as this seemed simple enough. There are only slight differences between pthread and C11 threads, in particular the way attributes are handled for mutex and condition variable (both not used by FLAC), the signature of the thread function, the return value of trylock, and the error handling, where pthread uses errno and C11 threads use an explicit return value enum. The wrapper macros are prefixed with FLAC__ as is common in FLAC code. All functions are renamed to closely resemble the C11 names instead of the POSIX names, as this is the more forward-looking approach.

The macro to enable C11 threads is HAVE_C11THREADS. When both HAVE_C11THREADS and HAVE_PTHREAD are defined, the C11 implementation is preferred.

This does not add any build system support for C11 threads, so the current default behaviour remains unchanged.


manxorist:
This whole thing is useful because MSVC provides native C11 threads support in recent versions, and has (of course) no pthread support. See <https://devblogs.microsoft.com/cppblog/c11-threads-in-visual-studio-2022-version-17-8-preview-2/>.

Also see <https://stackoverflow.com/questions/24557728/does-any-c-library-implement-c11-threads-for-gnu-linux> for some (outdated) discussion about C11 threads adoption in C standard library implementations.

manxorist:
@sezero As you mentioned the extra pthread dependency in <https://github.com/xiph/flac/pull/634#issuecomment-1732509971>, this might be relevant for you.

I did choose C11 threads instead of native WinAPI threads because the impedance mismatch between C11 threads and pthread is really low implementation-wise. As far as I know, MinGW-w64 does not support C11 threads yet, so this is not of immediate use with that particular toolchain. I would expect MinGW-w64 to eventually support C11 threads (it's part of ISO C after all), so this might be useful in the future.

For now, it adds in particular multithreading support with MSVC, for which the extra pthread dependency was even more of a hassle than it is for MinGW-w64.

sezero:
> @sezero As you mentioned the extra pthread dependency in [#634 (comment)](https://github.com/xiph/flac/pull/634#issuecomment-1732509971), this might be relevant for you.

Sorry, really not interested

ktmf01:
Thanks! Would indeed be great to have MSVC on board.

Just one comment for now (I haven't thoroughly reviewed this), can you move as much as possible into share/compat.h? That seems appropriate, even if it is only used in stream_encoder.c

manxorist:
> Just one comment for now (I haven't thoroughly reviewed this), can you move as much as possible into share/compat.h?

I would rather not use the global `compat.h` that. For MSVC, using `<threads.h>` is documented to induce an additional runtime dependency on `vcruntime140_threads.dll` (which is redistributable and part of the compiler runtime) (see <https://devblogs.microsoft.com/cppblog/c11-threads-in-visual-studio-2022-version-17-8-preview-2/>). As far as I can see, this does (currently) not make use of `#pragma comment(lib, [...])` (which would add the dependency from a sole `#include <threads.h>`), but it looks like to be implemented in the C runtime import library instead (I did not verify that). However, this fact is not documented and it would thus be relying on a current MSVC implementation detail. IFF this behaviour should ever change, everything that includes `share/compat.h` could inherit the dependency on the additional runtime DLL, which would not be desirable.

I can still move the abstraction macros of course, but I would rather use a distinct new `compat_threads.h`. Feels also a bit cleaner to not do totally separate things in a single kitchen sink abstraction header.

manxorist:
> I know close to nothing about CMake, and would prefer if someone else more knowledgeable touches that.

I did have a go at that, however I ended up running into CMake quirks for which I am not sure I found the appropriate work-around. The patch is in a separate branch at <https://github.com/manxorist/flac/tree/c11threads-cmake>.

ktmf01:
Thanks for all your work. I'll need some more time to review this.

For now: The CMake part looks good as far as I can see. Can you add that last commit to this branch, so the CI stuff you added can run?

manxorist:
> Can you add that last commit to this branch, so the CI stuff you added can run?

done

PR #829, opened 2025-04-14
Fix android compilation

abique:
See for reference https://github.com/microsoft/vcpkg/pull/44940

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -122,7 +122,7 @@ else()
 endif()
 
 
-if(ANDROID AND CMAKE_SYSTEM_VERSION VERSION_LESS 24 AND (CMAKE_SYSTEM_PROCESSOR MATCHES "i686" OR CMAKE_SYSTEM_PROCESSOR MATCHES "armv7-a"))
+if(ANDROID AND ANDROID_NATIVE_API_LEVEL VERSION_LESS "24" AND (CMAKE_SYSTEM_PROCESSOR MATCHES "i686" OR CMAKE_SYSTEM_PROCESSOR MATCHES "armv7-a"))
     # fseeko/ftello may link, but it's not usable before Android API 24 on 32-bit Android
     # https://android.googlesource.com/platform/bionic/+/main/docs/32-bit-abi.md
     message(STATUS "Disabling fseeko/ftello for 32-bit Android before API 24")


ktmf01:
@robUx4 Could you please take a look at this? This touches your fix from #691 I'd rather not have this break something else again, and I don't have a build system to test it with, neither does CI cover it

robUx4:
This seems based on bogus assumptions: https://github.com/microsoft/vcpkg/pull/44940#issuecomment-2794442460

> The recommended NDK toolchain (aka "legacy") sets CMAKE_SYSTEM_VERSION to 1, so libflac's approach to determine ftello absence is flawed even with vcpkg passing value 28 in.
>
> https://android.googlesource.com/platform/ndk/+/refs/heads/main/build/cmake/android-legacy.toolchain.cmake#38

This is true if `ANDROID_USE_LEGACY_TOOLCHAIN_FILE` is not set or explicitly [set to OFF in the environment](https://android.googlesource.com/platform/ndk/+/refs/heads/main/build/cmake/android.toolchain.cmake#47). People opting out of the legacy mode will [have the proper value](https://android.googlesource.com/platform/ndk/+/refs/heads/main/build/cmake/android.toolchain.cmake#180). On the other hand, the [documentation](https://developer.android.com/ndk/guides/cmake#the_new_toolchain_file) says people shouldn't use that...

So it's true that the value is not reliable enough. [`ANDROID_NATIVE_API_LEVEL`](https://developer.android.com/ndk/guides/cmake#android_native_api_level) seems appropriate although I think [`ANDROID_PLATFORM`](https://developer.android.com/ndk/guides/cmake#android_platform) is the right one to use here. [2k](path:cmakelists.txt "ANDROID_PLATFORM") uses vs [1.7k](https://github.com/search?q=path%3Acmakelists.txt%20%22ANDROID_NATIVE_API_LEVEL%22&type=code). Also the [NDK CMake file](https://android.googlesource.com/platform/ndk/+/refs/heads/main/build/cmake/android-legacy.toolchain.cmake#675) puts it in the read-only variables: "We'll keep these around for the existing projects that still use them.".

Note that this is only for people using the recommended cmake toolchain file provided by the NDK. In VLC for example we relied on CMake until recently and still do on VLC 3.0 (stable branch). Neither `ANDROID_PLATFORM` nor `ANDROID_NATIVE_API_LEVEL` exist [there](https://cmake.org/cmake/help/latest/manual/cmake-variables.7.html). In this case `CMAKE_SYSTEM_VERSION` seems the only way to tell the NDK target version.

TLDR: I think both values need to be checked. The PR in its current state will break the VLC 3.0 build.

abique:
I'm not expert here, I'm just forwarding what we've done in VCPKG.

robUx4:
This is one of the downside of VCPKG, they make patch locally and don't send them upstream for review. It's good that you submit it here @abique. That patch might be enough for VCPKG that may force the use of the `android.toolchain.cmake` file. 

For the more general case I think the `ANDROID_PLATFORM` and `CMAKE_SYSTEM_VERSION` need to be read first (priority to `ANDROID_PLATFORM`). For VLC the `CMAKE_SYSTEM_VERSION` set to 1 is not an issue because we don't build for anything higher than API 21 (so the check for < 24 is always true).

dg0yt:
I authored the original vcpkg patch, and I came here for upstreaming. I'm surprised about the negative judgements in the comments ("bogus assumptions", "don't send upstream") while confirming at the same time that this judgement is wrong ("documentation", this PR).

I don't claim to know much about flac, but I do have some familarity with vcpkg and most target platforms it supports. The original patch was intentionally minimal, tailored for vcpkg. When authoring the patch, I did research the changes which lead to the build problem, and I did assume that was testing with a different CMake toolchain. So I did expect that "upstreaming" would mean adjustments.

So how to continue from here in a positive way?

dg0yt:
There is one ugly thing about `ANDROID_PLATFORM`: It comes with an `android-` prefix in front of the number. 
Even ANDROID_NATIVE_API_LEVEL could have this prefix.
At least this is what initialization says:
https://android.googlesource.com/platform/ndk/+/refs/heads/main/build/cmake/android-legacy.toolchain.cmake#119

robUx4:
I verified that when using `$ANDROID_NDK/build/cmake/android.toolchain.cmake`, as one should,  and `-DANDROID_PLATFORM=21` the `CMAKE_SYSTEM_VERSION` stays to 1. With the following code:

> message(FATAL_ERROR "CMAKE_SYSTEM_VERSION=${CMAKE_SYSTEM_VERSION} / ANDROID_NATIVE_API_LEVEL=${ANDROID_NATIVE_API_LEVEL}")
> ...
> CMAKE_SYSTEM_VERSION=1 / ANDROID_NATIVE_API_LEVEL=21

Testing without and setting `-DCMAKE_SYSTEM_NAME=Android` (as one does when cross-compiling) and `-DANDROID_PLATFORM=21` I get the following:

> CMAKE_SYSTEM_VERSION=21 / ANDROID_NATIVE_API_LEVEL=

That's with CMake 3.22.1 on Ubuntu.

But there's a plot twist. If I change the `ANDROID_PLATFORM` value with the `android.toolchain.cmake` the `ANDROID_NATIVE_API_LEVEL` reflects the value unless the value is bogus and not handled by the NDK. In that case there's an error. Changing `ANDROID_PLATFORM` value with the  `-DCMAKE_SYSTEM_NAME=Android` does not change CMAKE_SYSTEM_VERSION value, it's stuck to 21, the minimum supported by the NDK. This is actually in line with [CMAKE_SYSTEM_VERSION](https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_VERSION.html) documentation that says:

> When the [CMAKE_SYSTEM_NAME](https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_NAME.html#variable:CMAKE_SYSTEM_NAME) variable is set explicitly to enable [cross compiling](https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html#cross-compiling-toolchain) then the value of CMAKE_SYSTEM_VERSION must also be set explicitly to specify the target system version.

So a user should either use `android.toolchain.cmake` and `-DANDROID_PLATFORM=xxx` or `-DCMAKE_SYSTEM_NAME=Android` and `-DCMAKE_SYSTEM_VERSION=xxx` to target the proper Android version.

> There is one ugly thing about ANDROID_PLATFORM: It comes with an android- prefix in front of the number.
Even ANDROID_NATIVE_API_LEVEL could have this prefix.

That line (119) is used in case the user sets `ANDROID_NATIVE_API_LEVEL` manually. But later in the script (681) the value is forced/reset to [ANDROID_PLATFORM_LEVEL](https://android.googlesource.com/platform/ndk/+/refs/heads/main/build/cmake/android-legacy.toolchain.cmake#681). This seems to be a number.

With all that being, to properly support builds with and without `android.toolchain.cmake` both `CMAKE_SYSTEM_VERSION` and `ANDROID_NATIVE_API_LEVEL` need to be checked. The `ANDROID_NATIVE_API_LEVEL` should only be used if `CMAKE_SYSTEM_VERSION` is 1, as it means `android.toolchain.cmake is used.

You could use something like this instead:
```
if(CMAKE_SYSTEM_VERSION EQUAL 1)
    set(CMAKE_SYSTEM_VERSION ${ANDROID_NATIVE_API_LEVEL})
endif()
```

In both usage it will have the proper Android target value.

robUx4:
Actually it's not entirely clean as it should only be done for Android targets, `CMAKE_SYSTEM_VERSION` may be 1 on other platforms.

Here are 2 ways to do it depending if you want to check ANDROID_NATIVE_API_LEVEL or CMAKE_SYSTEM_VERSION (I have no preference):

```
if(ANDROID AND CMAKE_SYSTEM_VERSION EQUAL 1)
    # android.toolchain.cmake resets CMAKE_SYSTEM_VERSION to 1
    set(CMAKE_SYSTEM_VERSION ${ANDROID_NATIVE_API_LEVEL})
endif()
```

```
if(ANDROID AND NOT ANDROID_NATIVE_API_LEVEL)
    set(ANDROID_NATIVE_API_LEVEL ${CMAKE_SYSTEM_VERSION})
endif()
```

Or simply switch the test to

```
if(ANDROID AND $<IF:CMAKE_SYSTEM_VERSION EQUAL 1,ANDROID_NATIVE_API_LEVEL,CMAKE_SYSTEM_VERSION> VERSION_LESS "24")
```

dg0yt:
I'm creating another PR.

PR #830, opened 2025-04-14
cmake: fix find Threads

### LINKED ISSUE #820: CMake: flac-config.cmake is missing find_dependency(Threads)
The [cmake Package Configuration File](https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#id17) should find all dependencies that the targets need, otherwise configuration fails.

Right now, doing `find_package(FLAC REQUIRED)` in a project fails, unless one uses the work-around of having `find_package(Threads REQUIRED)` beforehand, which is not so nice.

I believe all that needs to be done here is adding the following lines to [flac-config.cmake.in](https://github.com/xiph/flac/blob/master/flac-config.cmake.in) after the `include(CMakeFindDependencyMacro)`:

```
if(@HAVE_PTHREAD@)
    set(CMAKE_THREAD_PREFER_PTHREAD TRUE)
    set(THREADS_PREFER_PTHREAD_FLAG TRUE)
    find_dependency(Threads)
endif()
```

abique:
For reference: https://github.com/microsoft/vcpkg/pull/44940

### DIFF
--- a/flac-config.cmake.in
+++ b/flac-config.cmake.in
@@ -5,6 +5,10 @@ if(NOT TARGET Ogg::ogg)
     find_dependency(Ogg)
 endif()
 
+if (@ENABLE_MULTITHREADING@)
+    find_dependency(Threads)
+endif()
+
 include("${CMAKE_CURRENT_LIST_DIR}/targets.cmake")
 
 if(TARGET FLAC::FLAC)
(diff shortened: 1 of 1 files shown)


Croydon:
Related issue: #820

ktmf01:
Hi all. Sorry for the very much delayed reply. I've changed the PR a bit. I think this better reflects whether multithreading is actually used. As this is probably used locally (not moved between systems), it shouldn't matter much, but still.

Any thoughts?

PR #831, opened 2025-04-17
Fix android >=24 build with NDK toolchain

dg0yt:
Continued from https://github.com/xiph/flac/pull/829.
CC @robUx4


robUx4 (approved):
Looks Good To Me

PR #838, opened 2025-06-18
Allow --skip and --until with a total sample count of 0

Stefan-Olt:
This patch will allow use of `--skip` and `--until` when the total sample count is 0.

`--until` is not possible if relative offset from the end (error is shown). Relative from `--skip` value is possible.
`--skip` with a value after the end of the file will result in a seek error. As there is no information why seeking has failed a more detailed message is not possible.

### DIFF
--- a/man/flac.md
+++ b/man/flac.md
@@ -324,11 +324,6 @@ negated, see the **Negative options** section below.
 	sign means that the \--until point is taken relative to the \--skip 
 	point. For other considerations, see \--skip. 
 
-**\--no-utf8-convert**
-:	Do not convert tags from local charset to UTF-8. This is useful for
-	scripts, and setting tags in situations where the locale is wrong.
-	This option must appear before any tag options!
-
 **-s**, **\--silent**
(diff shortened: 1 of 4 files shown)


ktmf01:
@Stefan-Olt could you synchronize you branch (with the button pictured below, at https://github.com/Stefan-Olt/flac/tree/master), and then check whether it still works as you intended?

afbeelding

ktmf01:
I didn't mean the automated checks. Just to be sure: did you check whether your PR still does what it intends to do? Can you still use skip and until with a total sample count of 0? Because no automated tests for that behaviour have been added yet.

Stefan-Olt:
> I didn't mean the automated checks. Just to be sure: did you check whether your PR still does what it intends to do? Can you still use skip and until with a total sample count of 0? Because no automated tests for that behaviour have been added yet.

Sorry for being unclear: In the first sentence I was referring to manual testing, works just fine like before. Only the automated tests are not completed yet.

PR #842, opened 2025-08-11
encode & decode 32-bit float (IEEE 754 standard) samples.

aliheidary1381:
I've also added support to encode & decode raw, wave riff, & aiff float formats.

The float encoding feature achieves a near 70% compression ratio, which is better than nothing. No oss-fuzz or tests have been added yet (sorry). The replay gain feature should also be expanded in the future to support the new feature. Sorry for the big PR. It's pretty readable tho! I tried to make it as modular & independent as possible. Documentation is good. Should this make it to a release version, an update to the standard RFC could also be considered. The changes are backwards-compatible and are as follows:
1. When storing float samples, the bps bits in the streaminfo metadata block should be 0b00000.
2. When storing float samples, the zero-padded bit, originally reserved, after the bit-depth bits of each frame header should be 1.
3. When storing float samples, the actual data samples stored in each subframe are obtained by doing some bit manipulation before encoding and after decoding. That part is in src/libFLAC/transform_float.c and is necessary to boost the compression ratio (more info in the file comments). Monkey's Audio .ape does something similar, but this one seems to achieve better ratios (~8%). Sorry for the unorthodox type conversions. will fix it if I see it being considered for a merge.


aliheidary1381:
> No, there isn't. 

Oh, OK.
There's also a reserved bit pattern for the frame header's bit depth bits (`0b011`). How about doing something similar to what's done to the streaminfo bps bits?

aliheidary1381:
Well, now that we're modifying the STREAMINFO block, we can also change the 3-byte block size indicator to an elias gamma representation...

aliheidary1381:
Sorry guys... Been busy lately.
What is the final consensus on this? How should we extend the 16MiB block size limitation? A "continued" flag bit or a more fundamental change like replacing the 3-byte block size indicator with an elias gamma representation for streams with STREAMINFO_EXTENSION?

include/FLAC/format.h
```diff
@@ -178,6 +178,9 @@ extern FLAC_API const uint32_t FLAC__STREAM_SYNC_LEN; /* = 32 bits */
 /** The length of the FLAC signature in bytes. */
 #define FLAC__STREAM_SYNC_LENGTH (4u)
 
+typedef enum { NOT_SPECIFIED = -1,
+			   INT = 0,
```
madah81pnz1: Since this is put into the global scope/namespace, I suggest to add a prefix like SAMPLE_TYPE_ for all enum values, or even FLAC__SAMPLE_TYPE_.
Especially INT is a bit too short and too common name as to not be confusing.

include/FLAC/format.h
```diff
@@ -178,6 +178,9 @@ extern FLAC_API const uint32_t FLAC__STREAM_SYNC_LEN; /* = 32 bits */
...
+			   FLOAT = 1 } SampleType; // TODO: merge with endian & unsigned
```
madah81pnz1: The name of the type should be FLAC__SampleType, to follow naming style

src/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 option(ENABLE_64_BIT_WORDS "Set FLAC__BYTES_PER_WORD to 8, for 64-bit machines. For 32-bit machines, turning this off might give a tiny speed improvement" ON)
+option(ENABLE_EXPERIMENTAL_FLOAT_SAMPLE_CODING "Enable experimental feature to encode & decode 32-bit float (IEEE 754 standard) samples" ON)
```
madah81pnz1: Should be OFF by default?

src/flac/decode.c
```diff
@@ -32,6 +32,8 @@
 #include "share/compat.h"
 #include "decode.h"
 
+#define sample_type(is_float) is_float == FLOAT ? "float" : "int"
```
madah81pnz1: Make this a static inline function instead, sample_type is too short name as to not be confused for a local variable somewhere. Especially also since it is used as sample_type(sample_type).

The name could be something more like FLAC__get_sample_type_string().
Don't forget about the undefined enum value (-1), for debugging purposes I'd recommend to also check and print that one instead of "int" for everything else.

src/flac/encode.c
madah81pnz1, line 486: Here you should also check for 3, since that is when WAVEFORMATEXTENSIBLE is used for 32-bit float samples.

src/flac/encode.c
```diff
@@ -657,6 +692,11 @@ static FLAC__bool get_sample_info_aiff(EncoderSession *e, encode_options_t optio
 					e->info.is_big_endian = false;
 				else if(xx == 0x4E4F4E45) /* "NONE" */
 					; /* nothing to do, we already default to big-endian */
+#if ENABLE_EXPERIMENTAL_FLOAT_SAMPLE_CODING
+				else if(xx == 0x666c3332) { /* "fl32" */
```
madah81pnz1: Maybe also check for "FL32" since that seems to also be used:
https://www.mmsp.ece.mcgill.ca/Documents/AudioFormats/AIFF/AIFF.html

src/flac/main.c
```diff
@@ -869,6 +882,17 @@ int parse_option(int short_option, const char *long_option, const char *option_a
 			else
 				return usage_error("ERROR: argument to --endian must be \"big\" or \"little\"\n");
 		}
+#if ENABLE_EXPERIMENTAL_FLOAT_SAMPLE_CODING
+		else if(0 == strcmp(long_option, "sample-type")) {
+			FLAC__ASSERT(0 != option_argument);
+			if(0 == strcmp(option_argument, "float") || 0 == strcmp(option_argument, "IEEE754") || 0 == strcmp(option_argument, "binary32"))
```
madah81pnz1: I suggest to only allow "float" here, having others is only confusing and means it would be impossible to remove these later

src/flac/main.c
```diff
@@ -869,6 +882,17 @@ int parse_option(int short_option, const char *long_option, const char *option_a
...
+				option_values.format_sample_type = FLOAT;
+			else if(0 == strcmp(option_argument, "int") || 0 == strcmp(option_argument, "integer"))
```
madah81pnz1: Same here, only "int".

src/flac/main.c
```diff
@@ -1965,6 +2028,16 @@ int decode_file(const char *infilename)
 		return 1;
 	}
 
+#if ENABLE_EXPERIMENTAL_FLOAT_SAMPLE_CODING
+	if(option_values.format_sample_type == FLOAT) {
+		FLAC__uint32 test = 1;
+		FLAC__bool is_big_endian_host_ = (*((FLAC__byte *)(&test))) ? false : true;
```
madah81pnz1: Check WORDS_BIGENDIAN instead

src/libFLAC/metadata_iterators.c
```diff
@@ -2332,6 +2332,12 @@ FLAC__Metadata_SimpleIteratorStatus read_metadata_block_data_streaminfo_cb_(FLAC
 	block->sample_rate = (unpack_uint32_(b, 2) << 4) | ((uint32_t)(b[2] & 0xf0) >> 4);
 	block->channels = (uint32_t)((b[2] & 0x0e) >> 1) + 1;
 	block->bits_per_sample = ((((uint32_t)(b[2] & 0x01)) << 4) | (((uint32_t)(b[3] & 0xf0)) >> 4)) + 1;
+#if ENABLE_EXPERIMENTAL_FLOAT_SAMPLE_CODING
+	if(block->bits_per_sample == 1) {
```
madah81pnz1: I'm not sure if reusing  bits_per_sample as 1 to mean floats is a good idea here, but this might require further discussions.
For example, supporting float16_t would be possible by allowing bits_per_sample as 16, but then requires some other method to determine if it's int or float

src/metaflac/options.c
```diff
@@ -488,6 +499,24 @@ FLAC__bool parse_option(int option_index, const char *option_argument, CommandLi
 		else
 			undocumented_warning(opt);
 	}
+#if ENABLE_EXPERIMENTAL_FLOAT_SAMPLE_CODING
+	else if(0 == strcmp(opt, "set-sample-type")) {
+		FLAC__ASSERT(0 != option_argument);
+		op = append_shorthand_operation(options, OP__SET_SAMPLE_TYPE);
+		if(strcmp(option_argument, "float") == 0 || strcmp(option_argument, "IEEE754") == 0 || strcmp(option_argument, "binary32") == 0) {
```
madah81pnz1: same here, only allow "float", and "int" below

src/libFLAC/transform_float.c
```diff
@@ -0,0 +1,199 @@
...
+		my tests resulted in an ~80% ratio
+   2. doing some basic bit manipulation (current approach)
+		saves ~3 bits theoretically
+		my tests resulted in a ~70% ratio,
+		even reaching 51% when the stream is a direct conversion from int samples to 32-bit floats.
+   3. splitting the "exponent" (8 bit) and "sign+significand" (24 bit) parts of floats into two channels,
+	subtracting an (automatically recognised) DC offset from the "exponents" channel
+	and	storing the offset in each frame header (i.e. unsigned to signed conversion)
+		could be better (haven't tried it yet. hard for me to implement).
+		my guess would be a *consistent* ~60% ratio, at least.
```
madah81pnz1, line 66: a 10% improvement in compression might be worth investigating a bit further? Would need more discussions on how to split up the signal in "sub-channels", maybe too much complexity to be worth it.

How does the current (choice 2) compare to other coders like WavPack? If it is already as good or very close, then it might not be worth it.

src/flac/decode.c
madah81pnz1, line 1075: You need to handle float here also for waveformatextensible, the first value of the GUID would be 3 for float

src/flac/decode.c
```diff
@@ -1195,6 +1252,30 @@ FLAC__StreamDecoderWriteStatus write_callback(const FLAC__StreamDecoder *decoder
 	if(decoder_session->abort_flag)
 		return FLAC__STREAM_DECODER_WRITE_STATUS_ABORT;
 
+#if ENABLE_EXPERIMENTAL_FLOAT_SAMPLE_CODING
+	/* sanity-check the sample type */
+	if(decoder_session->sample_type != -1) {
```
madah81pnz1: Use the enum value, don't hardcode -1

src/flac/decode.c
```diff
@@ -1582,6 +1663,20 @@ void metadata_callback(const FLAC__StreamDecoder *decoder, const FLAC__StreamMet
 				decoder_session->abort_flag = true;
 				return;
 			}
+#if ENABLE_EXPERIMENTAL_FLOAT_SAMPLE_CODING
+			if(decoder_session->sample_type == FLOAT && metadata->data.stream_info.sample_type != FLOAT) {
+				stats_print_name_and_stream_number(1, decoder_session->inbasefilename, decoder_session->stream_counter);
+				flac__utils_printf(stderr, 1, "ERROR, this link's STREAMINFO is set to float PCM format but was int PCM in previous one\n");
```
madah81pnz1: Use the sample_type string function here, so the name is consistent everywhere

aliheidary1381:
Hey guys. Sorry for taking this long to respond. I resolved the issues noted in the review. Thx for the kind and thorough review! I also removed anything related to changing the frame header bits. As I've mentioned before, float is already outside of the streamable subset ('cause it's 32 bits). There was no need to change anything in frame headers.

jesperpedersen:
@aliheidary1381 Thanks for giving this a go ! CI reports conflicts, but should be easy

src/libFLAC/transform_float.c
aliheidary1381: The current choice (2) that I've implemented performs a little bit better than Monkey's Audio `.ape`, which also implements option 2, but performs worse than WavPack, which implements option 3. It IS worth it. I don't know if I'll be able to do it tho.

aliheidary1381:
Yes, resolving the conflict should be easy. My biggest concern here is our discussion about STREAMINFO_EXTENSION here.

A summary of the discussions about it:

Since we're already deviating from the standard, and older decoders won't be able to read the new float stream types anyway, we might as well introduce a STREAMINFO "extension", future-proofing other potential breaking changes that FLAC could benefit from. Possible future extensions discussed are "1-bit PWM streams", "immersive/3d moving sound objects", and "better channel masks and sample rates".

The STREAMINFO structure must be kept the same, so that the "standard" streams generated by the new encoder versions would still be readable by older decoders. We're using forbidden values in the original STREAMINFO (like sample_rate=0 or bps=1-3) to suggest to the older decoders not to play it, and also to indicate that there's a STREAMINFO_EXTENSION block on the way.

I'll try to write a proposal for the new header in the following week, and will TRY to implement it in the following month or so, if nobody puts an objection to the proposal.

H2Swine:
@ktmf01 (and @aliheidary1381 and anyone): 

> I am surprised you didn't need any new coding methods (subframe types or residual coding methods). I think this is really nice.
> 
> FLAC is also used for compression measurement signals (for scientific experiments and such) where this also might be useful. Not sure whether your bit manipulation works for that as well.

Is it a given that we can do without new residual coding? Especially in the latter situation, we could have several NaN showing up that might not (I don't know this though!) show up in audio test files. Just testing it on audio that has presumably originated as integer-PCM and then been through some float math in the manipulation might not pinpoint the issues that may show up when you get NaNs in. 

_If_ a new residual coding method turns out to be necessary or warranted (like "variable-bit Rice" and possibly with extensions to say, pack at the end of the frame single sample corrections), that could solve something. Old players would just stop at unknown residual methods, so reserve a new method and forbid the two subframes that won't need residuals (VERBATIM and CONSTANT ... or assign two of the twenty-five remaining subframe types)

aliheidary1381:
The STREAMINFO_EXTENTION proposal:

The presence of the STREAMINFO_EXTENTION metadata block MUST be indicated in the STREAMINFO block by setting the bps bits to 0b00000 (meaning 1bps, which was forbidden before). This also suggests to the older decoders to refuse decoding the new stream specifications. If present, the new decoders MUST replace every valid information from STREAMINFO_EXTENTION with anything inside the STREAMINFO block.
If present, the STREAMINFO_EXTENTION block MUST be followed exactly after the STREAMINFO block. The STREAMINFO_EXTENTION block header will use the reserved value 7 as its block type indicator.

The block, 128 bits in total, contains the following bits:
1. 64 bits for sample rate in Hz. The number should be represented in the IEEE 754 standard binary64 format, often called a double, or f64. Special numbers (subnormal numbers, signed zeros, infinities, and NaNs) and negative values are forbidden. The presence of such numbers should result in fallback to the STREAMINFO block info.
2. 8 bits for the (number of channels)-1.
3. 32 bits for a standard WAVEFORMATEXTENSIBLE channel mask.
4. 6 bits for if the coder should ignore the WAVEFORMATEXTENSIBLE bits, and use another custom channel mask. All is reserved for the time being.
5. 4 bits for the sample format. An enum with the following values: 0 for normal PCM, 1 for float PCM (IEEE standards). Other values are reserved for future development.
6. 6 bits for the (bit depth)-1. If the sample format is float PCM, the bps MUST be one of 32 or 64.
7. 8 bits reserved.

When necessary (i.e. when STREAMINFO_EXTENTION is present and it contains valid info), the sample rate in frame headers MUST be 0b0000, the bit depth bits in frame headers MUST be 0b000, and the channel mask bits MUST be 0b1111, to defer parameters to the STREAMINFO_EXTENSION block.

aliheidary1381:
I still insist that extending the frame header, or adding a new subframe type is outside the scope of this PR.

[Current discussions](https://github.com/xiph/flac/discussions/877) about the 16MiB block size limitation doesn't seem to benefit from an extended streaminfo. I think we can put that one outside this PR's scope too.

If PDM sample format support would be added in the near future, we could add: If the sample format is PDM, the bit depth MUST be 0. If the sample format is PCM, the bit depth MUST NOT be 0.

It is possible to extend this proposal for 3D objects using the reserved bits, even after a merge. I intentionally didn't mention it in the proposal. Based on my research, I believe that a new 3D object metadata standard just for FLAC won't be adopted by consumers (especially commercial software like DAWs). DTS:X tried this once with MDA. The most adopted open-source alternative is IAMF, which is a container that supports FLAC already. Nothing we'd need to do here.

madah81pnz1:
> The STREAMINFO_EXTENTION proposal:
> 
> The presence of the STREAMINFO_EXTENTION metadata block MUST be indicated in the STREAMINFO block by setting the bps bits to 0b00000 (meaning 1bps, which was forbidden before). This also suggests to the older decoders to refuse decoding the new stream specifications. If present, the new decoders MUST replace every valid information from STREAMINFO_EXTENTION with anything inside the STREAMINFO block. If present, the STREAMINFO_EXTENTION block MUST be followed exactly after the STREAMINFO block. The STREAMINFO_EXTENTION block header will use the reserved value 7 as its block type indicator.
> 
> The block, 128 bits in total, contains the following bits:
> 
>     1. 64 bits for sample rate in Hz. The number should be represented in the IEEE 754 standard binary64 format, often called a double, or f64. Special numbers (subnormal numbers, signed zeros, infinities, and NaNs) and negative values are forbidden. The presence of such numbers should result in fallback to the STREAMINFO block info.
> 

Yes this makes sense, as the new block enables float types, the decoder should have no problems reading this value. This also enables lossless transformation from CAF, which also uses 64-bit float, and near-lossless from AIFF, which uses 80-bit.

But then we have to decide what to do when decoding to .wav, should it give an error, or we round/truncate to an integer value.

A potential problem is for hardware decoder without float support. But manually decoding the float should be fine, since that is how everyone does it for AIFF's samplerate 80-bit extended float.

>     2. 8 bits for the (number of channels)-1.

256 channels is nice. But how will that work in practice with regards to the channel mask or layout?
As an example, a 64-channel file would still decode fine to wav, as only the first 32 channels can be defined.

> 
>     3. 32 bits for a standard WAVEFORMATEXTENSIBLE channel mask.

Yes this is a good idea since it is now commonly supported. This means we can also get rid of the WAVEFORMATEXTENSIBLE_CHANNEL_MASK vorbis comment.

> 
>     4. 6 bits for if the coder should ignore the WAVEFORMATEXTENSIBLE bits, and use another custom channel mask. All is reserved for the time being.
> 

I don't think 6 bits is enough for anything useful though. It is maybe better to define a whole new block type for advanced channel layouts in that case.

Might be worth checking out CAF Channel Layout, Channel Description and related chunks:
https://developer.apple.com/library/archive/documentation/MusicAudio/Reference/CAFSpec/CAF_spec/CAF_spec.html#//apple_ref/doc/uid/TP40001862-CH210-BCGCIJCF


>     5. 4 bits for the sample format. An enum with the following values: 0 for normal PCM, 1 for float PCM (IEEE standards). Other values are reserved for future development.
> 
>     6. 6 bits for the (bit depth)-1. If the sample format is float PCM, the bps MUST be one of 32 or 64.

Maybe allow for 16 and 128 in a future version, due to the existence of float16_t/float128_t in C/C++ and f16/f128 in Rust.

> 
>     7. 8 bits reserved.
> 
> 
> When necessary (i.e. when STREAMINFO_EXTENTION is present and it contains valid info), the sample rate in frame headers MUST be 0b0000, the bit depth bits in frame headers MUST be 0b000, and the channel mask bits MUST be 0b1111, to defer parameters to the STREAMINFO_EXTENSION block.

What if the STREAMINFO_EXTENTION contains mostly valid data for an older decoder, e.g. only a fractional samplerate of 44100.123 is set, and so the STREAMINFO block contains 44100.
* On one hand, if would be nice if we can maximize backwards compatibility so a file can play and be decodable if only a few new features are used.
* But on the other hand, an old decoder can't losslessly decode or restore back the original file in this case.

Another idea is to not include STREAMINFO at all, but that might mean rejecting the file as a flac file entirely.

madah81pnz1:
> I still insist that extending the frame header, or adding a new subframe type is outside the scope of this PR.
> 

The reason to do this at the same time as STREAMINFO_EXTENTION is this is the one chance for any backwards incompatible format changes. It is better to plan out all potential "flac 2.0" format changes, because I think it will be much harder to do changes to the frame header later.

I agree about separating out the changes, not everything has to be crammed into the same PR. But then these changes should be protected behind some #ifdef until everything is ready. A separate branch is another way to do it.

One idea for a changed format header would be to extend the channel decorrelation. Now it is limited to 2 channels only. For example, a simple left-right/front-back pairings could improve compression for surround audio.

But this feels a bit unexplored though. Maybe it doesn't give so much compression benefits, or there is not enough 8+ channels source material to draw some conclusions from.

> [Current discussions](https://github.com/xiph/flac/discussions/877) about the 16MiB block size limitation doesn't seem to benefit from an extended streaminfo. I think we can put that one outside this PR's scope too.
> 

Yes I agree, I don't know if there's even a need for it, or interest.

aliheidary1381:
Thx for the wonderful feedback.

> This also enables lossless transformation from CAF, which also uses 64-bit float, and near-lossless from AIFF, which uses 80-bit.

What do you think about using f128 (for sample rate) instead? That would make encoding AIFF streams lossless, and will still work on non-x86 architectures (which don't support f80). The downside is that almost no CPU architecture natively supports f128, and the C compiler would fall back to software emulation on most architectures.

> I don't think 6 bits is enough for anything useful though.

You're right. I'll bring it down to just 2 bits to hint to the coder to look for an additional block with explicit channel layouts (like an array of labels/coordinates like CAF).

> Maybe allow for 16 and 128 in a future version, due to the existence of float16_t/float128_t in C/C++ and f16/f128 in Rust.

Agreed. Will bump it to 7 bits instead.

> Might be worth checking out CAF Channel Layout, Channel Description and related chunks

Seems like CAF's layout tags have some advantages over channel bitmaps like WAVEFORMATEXTENSIBLE. The most interesting one for me is differentiating use cases (e.g. standard stereo vs. headphone stereo). I'd keep the remaining 3 bits reserved for this use case.

> The reason to do this at the same time as STREAMINFO_EXTENTION is this is the one chance for any backwards incompatible format changes. It is better to plan out all potential "flac 2.0" format changes, because I think it will be much harder to do changes to the frame header later.

I understand. Convinced me.

I guess that'd mean that we'd have to mandate the "extended" frame header whenever the extended streaminfo is present (like an "if and only if" situation).

Just to summarize: The following ideas have been presented here by now:
An extra byte of feature flags, a channel mask, a decorellation technique descriptor/id.

madah81pnz1:
> What do you think about using f128 (for sample rate) instead? That would make encoding AIFF streams lossless, and will still work on non-x86 architectures (which don't support f80). The downside is that almost no CPU architecture natively supports f128, and the C compiler would fall back to software emulation on most architectures.
> 

f128 would be overkill.
AIFF using 80-bit float is more of a historical artifact, and since CAF is meant as the successor to AIFF; I think we can forget about AIFF compatibility. The exact samplerate can always be restored via foreign metadata.

flac's 1 MHz samplerate range is a bit limited though, since it doesn't even reach up to the lowest DSD(64) samplerate of 2.8 MHz (64*44100).

Interms of decoding complexity, a 32:32 fixed point format could also work. Then there's no worry about NAN, INF, or subnormals.
But in practice, the user application will anyway have to convert this into some float value, or discard the fractional bits. It will only matter for an integer-only player, but on such hardware playing 32-bit float audio would struggle anyway.

aliheidary1381:
Understood.

aliheidary1381:
The frame header is self-contained, and usually doesn't need the STREAMINFO. With the extension, we should also choose wether to let go of this advantage and defer the info to the STREAMINFO extension block (which breaks seeking on raw streams), or to add some frame header overhead (3-16 additional bytes on disk for each frame).
What do you guys think about this?

H2Swine:
> The frame header is self-contained, and usually doesn't need the STREAMINFO. With the extension, we should also choose wether to let go of this advantage and defer the info to the STREAMINFO extension block (which breaks seeking on raw streams), or to add some frame header overhead (3-16 additional bytes on disk for each frame).
> What do you guys think about this?

More important than the streaming _ability to play_ without file headers, I think is to avoid wrong playback. A "current integer-only flac player" must be able/forced to _discard_ a float frame. 

Let's take the frame header in order: 
- Sync word. Could be changed. Current players will not recognize it as FLAC frame. Risky, with fewer guards against false positives.
- Block size. That info needs to be in there anyway, so don't touch.
- Sample rate. Here we can make it mandatory to refer to STREAMINFO. (Got no STREAMINFO? Don't play this.) There are also three different zeroes, which could be different flags. Or the forbidden value could be allowed, to mean that here will immediately follow an 80-bit value (this to be able to compress "any AIFF"). And since 80-bit floats are signed while sample rates are not, the sign bit could be reserved for a flag of "OK to round off to nearest manageable integer when playing back". Side effect of using this forbidden value is that now we have the cat out of the bag, this can flag a "new-format block header": from here on, everything is new until the subframe header. Or ditch the 80 and say 64, where a negative sign says inaccuracy allowed (and if so, let the standard mandate to unpack with foreign metadata so that the lone noninteger-AIFF user can get their file back). I presume that CAF has no special provision for negative sample rates?
- Channels. A few values to spare, but not as much as a bit. _If_ that is to be tweaked, then my two cents on allocating values to e.g. "channel count & allocation in file-level headers only", "channel count & allocation in file-level headers only (FL&FR as MS)",  "channel count & allocation in file-level headers only (any L&R as MS)" and "channel allocation will follow next as more bits". 
- Bit depth bits + the next. Here there is play room. There is a full bit and 9 values left. More than enough for 1-bit, two different 16-bit floats, 32 and 64. I know that ktmf01 has warned against spending the bit, but we need not reserve all of it? The reserved _bit_ can even be saved if we use the reserved _value_ of 0b011 to indicate "other, please read on". 
- ...
- Header CRC. Theoretically you could just add/subtract 1 for "nonstandard header, more header will follow". Current players will dump the frame ... or is there a risk that some don't, and just ignore the CRC and start playing garbage?

aliheidary1381:
The latest STREAMINFO_EXTENTION proposal:

The presence of the STREAMINFO_EXTENSION metadata block MUST be indicated in the STREAMINFO block by setting the bps bits to 0b00000 (meaning 1bps, which was forbidden before). This also causes older decoders to refuse to decode the new stream specifications. If present, new decoders MUST replace the corresponding "legacy" info from the STREAMINFO block with every valid piece of information from STREAMINFO_EXTENSION, and ignore the invalid ones.
If present, the STREAMINFO_EXTENSION block MUST immediately follow the STREAMINFO block. The STREAMINFO_EXTENSION block header will use the reserved value 7 as its block type indicator.

The block, 128 bits in total, contains the following fields:

1. 64 bits for sample rate in Hz. The number should be represented in the IEEE 754 standard binary64 format, often called a double, or f64. Special numbers (subnormal numbers, signed zeros, infinities, and NaNs) and negative values are forbidden. The presence of such numbers should result in fallback to the STREAMINFO block info.

2. 8 bits for (number of channels) - 1.

3. 32 bits for a standard WAVEFORMATEXTENSIBLE channel mask.

4. 3 bits for hinting a special use case for the channel bitmap (e.g., standard stereo vs. headphone stereo). All values are reserved for the time being.

5. 2 bits indicating whether the decoder should ignore the WAVEFORMATEXTENSIBLE bits and use another custom channel mask. Reserved for the time being.

6. 4 bits for the sample format. An enum with the following values: 0 for normal PCM, 1 for float PCM (IEEE standards). Other values are reserved for future development.

7. 7 bits for (bit depth) - 1. If the sample format is float PCM, the bps MUST be either 32 or 64.

8. 8 bits reserved.

Compatibility rules:

1. STREAMINFO_EXTENSION's sample rate may be 0 to indicate that the information is only stored in STREAMINFO. Otherwise, for better backwards compatibility, STREAMINFO's sample rate MUST be a rounded version of the STREAMINFO_EXTENSION sample rate (capped at 1048575 = 2^20-1 for higher rates).

2. STREAMINFO_EXTENSION's channel count bits may be 0 to indicate that the information is only stored in STREAMINFO. Otherwise, if the channel count is higher than STREAMINFO's capacity, STREAMINFO MUST store 0 bits. Otherwise, they MUST both store the same count.

aliheidary1381:
My last commit claimed that it had reached 51% compression. It was a bug :) Though my original 70% compression claim remains true.

STREAMINFO_EXTENTION is now working well. Lots of more changes need to be done that I'm unwilling to do by myself. Though the storage works fine, the playback is probably not (I didn't test it). More specifically, seeking and seek table generation now needs to consider float sample rates, replay gain now needs to consider float samples, and higher bps means expanding most of the codebase (128 can't even fit in a single register).

Still, having the header set up and a working float encoder is still nice. Next milestone is a working frame header extension.

aliheidary1381:
Force push to fix the bug in git history 👆

I know these big changes are making reviews harder and harder. I'm trying to keep it clean.

aliheidary1381:
Frame header extension proposal:

The STREAMINFO extension coincides with the frame header extension. The encoder MUST use the extended frame header if and only if STREAMINFO_EXTENSION is present. The extension bits MUST immediately follow the main frame header bits and directly precede the CRC. The CRC covers the extension bits as well. (This also helps force legacy decoders to drop the frame, because the CRC is no longer in its legacy position, causing shifted bits to fail CRC validation).

Deliberate breaking of backwards compatibility: when the frame header extension is present, the legacy frame header sample rate bits MUST be 0b1111 (previously forbidden), channel bits MUST be 0b1111 (previously reserved), and bit depth bits MUST be 0b011 (previously reserved). This is the primary indicator of an extended frame header.

Uncommon sample rates MUST NOT be used together with the frame header extension. Uncommon sample rates are deprecated; the frame extension should be used instead.

The reserved zero-bit after the bit-depth field in the legacy frame header indicates the extension format:
 - `0`: Compact mode (inherits sample rate, channel count, channel mask, bit depth, and sample format from STREAMINFO_EXTENSION).
 - `1`: Full mode (explicitly defines all stream parameters inline).

In compact mode (8 bits in total), the extension contains the following fields:
1. 3 bits for the inter-channel decorrelation mode descriptor ID (see decorrelation mapping below; non-stereo values are reserved).
2. 5 bits reserved.

In full mode (128 bits in total), the extension contains the following fields:
1. 64 bits for sample rate in Hz. The value MUST be represented in the IEEE 754 standard binary64 format (double / f64). Special values (subnormal numbers, signed zeros, infinities, and NaNs) and negative values are forbidden. The presence of forbidden values MUST result in fallback to the STREAMINFO_EXTENSION metadata block.
2. 8 bits for (number of channels) - 1.
3. 32 bits for a standard WAVEFORMATEXTENSIBLE channel mask.
4. 3 bits for hinting a special use case for the channel bitmap (e.g., standard stereo vs. headphone stereo). All values are reserved for the time being.
5. 2 bits indicating whether the decoder should ignore the WAVEFORMATEXTENSIBLE bits and use a custom channel mask. Reserved for the time being.
6. 3 bits for the inter-channel decorrelation mode descriptor ID (see decorrelation mapping below; non-stereo values are reserved).
7. 7 bits for (bit depth) - 1. If the sample format is float PCM, the bps MUST be either 32 or 64.
8. 5 bits reserved.
9. 4 bits for the sample format enum (0 for integer PCM, 1 for float PCM via IEEE standards; other values reserved for future development).

Inter-channel Decorrelation Mapping
The interpretation of the 3-bit inter-channel decorrelation mode descriptor ID depends on the active channel configuration / WAVEFORMATEXTENSIBLE mask.
For standard 2-channel stereo (SPEAKER_FRONT_LEFT | SPEAKER_FRONT_RIGHT):
 - `0b000`: Independent channels (Channel 0 = Left, Channel 1 = Right).
 - `0b001`: Left-side stereo (Channel 0 = Left, Channel 1 = Side).
 - `0b010`: Side-right stereo (Channel 0 = Side, Channel 1 = Right).
 - `0b011`: Mid-side stereo (Channel 0 = Mid, Channel 1 = Side).
 - `0b100`-`0b111`: Reserved.

For all other channel configurations, all 3-bit values are currently reserved for future multi-channel decorrelation schemes.

aliheidary1381:
> STREAMINFO_EXTENTION is now working well. Lots of more changes need to be done that I'm unwilling to do by myself. Though the storage works fine, the playback is probably not (I didn't test it). More specifically, seeking and seek table generation now needs to consider float sample rates, replay gain now needs to consider float samples, and higher bps means expanding most of the codebase (128 can't even fit in a single register).

These issues are now also resolved.
(except for new bps values like 128 bps or 1 bps, where flac throws an error saying that it's not supported by flac even tho it is standard)

aliheidary1381:
> a 10% improvement in compression might be worth investigating a bit further? Would need more discussions on how to split up the signal in "sub-channels", maybe too much complexity to be worth it.
> 
> How does the current (choice 2) compare to other coders like WavPack? If it is already as good or very close, then it might not be worth it.

This will also be resolved in the following commits to come.

Originally, my float coding trick was to do some bit manipulations on the samples. It reached 70% ratio when the audio stream was a direct conversion from i24 samples. The new method reaches ~45% ratio in such situations, and reaches ~75% ratio otherwise. This is the best ratio accomplished by any of our competitors (lossless codecs) AFAIK. The details are as follows.

There are two "modes" of compressing float samples. One specifically designed to compress streams that are direct conversions from int PCM streams (which I call the 1-subframe mode), and one for everything else (which I call the fallback mode). The float compression mode is frame-wise, meaning that the compression mode for each channel in a single frame is the same. The system automatically detects which mode to operate on, and stores the mode for the decoder in one of the new reserved bits in the frame header extension (4 bits are still reserved). The first (of 5) reserved bit is 1 iff we're in fallback mode. I've also added options to force modes for testing the compression ratio in each mode.

To detect which mode to operate on, we essentially test if we can losslessly convert the float samples to i24. If the test succeeds, we do the conversion and store normal 24-bit ints for each sample. Since we're throwing away the exponent bits (assuming that they're all about the same, because the tests succeeded), we should also store them in the frame header, just as uncommon sample rates are stored by extending the header. One byte for each channel.

If the tests fail, we'll be on fallback mode. Each channel is split into two subframes. So a 2-channel stereo stream would have 4 subframes in each of its frames. One subframe is for storing the exponent bits, and the other is for storing the sign&significand. The sign&significand are processed by converting to two's complement to make it more aligned to what the LPC algorithm expects. The exponent bits are also converted from offset representation to two's complement.

aliheidary1381:
What do you guys think about the changes?

PR #845, opened 2025-08-17
flac.md : Some more fixes: errors, language.

H2Swine:
By request, I have not submitted a full proposal, thus some "Everywhere" changes are not at all (yet) implemented everywhere. 

**Everywhere / general**
 - boldfaced **flac**-the-tool, capitalized FLAC-the-format 
 - italicized *warnings* and *errors*. 
 - What in the text was Warning-but-not-a "*warning*", is now "CAUTION:" (4 instances in this commit). "Note:" is now "NOTE:" Both start on newline.
 - Tried to stick to more consistent phrases about "print" a *warning*, and "exit with" an *error* (when it actually exits), and file already "exists" rather than a mix between "is present" and whatnot.
 - Got rid of a bunch of "specify"/"specifies", when the use was far from a _specification_ as in the FLAC specification or PICTURE specification. I just imagine that someone who text-searches for specification will be more interested in those which are left.

Apart from those, here are some of them explained by referring to line numbers in ***current*** ("old") flac.md; also some possible changes that I have **not** taken on board (yet). Ignoring those listed above, and some I think are self-explanatory, and also skipping justification for some that have gotten a thumb up from @ktmf01 at https://github.com/xiph/flac/commit/b359d00fe708bb8f433ba0ecdf68015a91c4695c
 - 182: **ERROR FIXED** Previous change of mine (my blame!) couldn't stick consistent on "Track04" vs Track07 ...
 - also: rewritten. 
 - 205-206: It does not "exit" with an error if processing multiple files. And the "already exists".
 - 222: Tried to improve language.
 - 231: Rewritten.
 - 245: As discussed earlier: Referring to the **Format options** 
 - 273 deviates from a suggestion from @ktmf01 - "print" a *warning* etc.  
 - 282: I didn't like "force" ... cf -f.
 - 298: Longer explanation. 
 - 312: Elaborates. Written to fit \--until.
 - 321: **ERROR FIX!** The syntax is wrong in current flac.md. Hopefully it is now right. And admittedly, this is now written in more detail *because* I first got it wrong.
 - 328 as mentioned earlier: "mode" is better reserved for encoding mode, decoding mode, ...
 - 335: Statement is not completely correct, but I put a reservation in the `--apply-replaygain-which-is-not-lossless` text.
 - 340: Another ref to "Format options"
 - 342: Elaborates. With CAUTION:.
 - 344: There are exceptions, at least \--apply-replaygain-which-is-not-lossless. 
 - 366 **not yet done** (but I have it drafted): Could need a rewrite.  
 - 376: Encoding starts here. No need to mention the tukey, but something about format choice ... 
 - **Not yet done** (but I have it drafted): Level-3 headlines for compression/audio options and for metadata options. 
 - 378: -V moved downwards, as discussed.
 - 410 to 421: Have been discussed. Mostly, the "specifies" ... and then re-wrote.
 - 438: Reworded. **QUESTION:** Since it _did_ say the wrong comma-separated, should it now for correction outright _say_ semicolon-separated rather than this PR which just removes the wording and leaves it for the end of the document? 
 - Up to 463: I think these have been discussed.
 - 464-ish: Moved `--ignore-chunk-sizes` here, it is audio ... 
 - 484: ReplayGain is tags, moved down. Also: I couldn't provoke the warning about extra padding - omitted it, I don't think it is really necessary.
 - 498 to 511: Moved - and are now longer. **Please review**.
 - 516: Do I understand correctly that -T ... --no-utf-8 -T ... will apply no-utf-8 to the latter and not the first tagging?
 - 529: I don't think it has to say anything about UTF-8 when that is said under -T and this is like -T except ...
 - 537: Rewritten ... 
 - 554: -P is rewritten to get it on the format where it first says what it *does*. 
 - 605: Elaborated on `--force-raw-format` after testing. In particular that you should usually omit it when encoding.
 - 610 ff: Clearer what is input/output, rid of "Specifies", make clear that \--bps must be multiple of 8
 - 650: Also to negate defaults ... if that wasn't obvious.
 - 672 ff: **Not yet done** (but I have it drafted): Extensive rewrite of everything from here to the end.
 - 720 ff: Wording of 0 to 20 now taken from the RFC. I figure that is an uncontroversial change.
 - 789 **ERROR FIX!** The comma/semicolon error. (I have drafted a full rewrite, but not here). 
 - But, **question**: "subdivide_tukey(1/P)" is not implemented ... should be mentioned?
 - 827: Useful place to put the info on where it is maintained?

### DIFF
--- a/man/flac.md
+++ b/man/flac.md
@@ -180,114 +180,115 @@ Some typical encoding and decoding tasks using flac:
 	tags from a source file, use tagging software which supports them.
 
 `flac -Vj2 -m3fo Track07.flac  -- -7.wav`
-:	flac employs the commonplace convention that options in a short 
-	version - invoked with single dash - can be shortened together until 
-	one that takes an argument. Here -j and -o do, and after the "2" a 
-	whitespace is needed to start new options with single/double dash. 
-	The -m option does not, and the following "3" is the -3 compression
-	setting. The options could equally well have been written out as 
-	-V -j 2 -m -3 -f -o Track04.flac , or as -fo Track04.flac -3mVj2. 
-	flac also employs the convention that `-- ` (with whitespace!) 
-	signifies end of options, treating everything to follow as filename.
-	That is needed when an input filenames could otherwise be read as an
-	option, and "-7" is one such.
-	In total, this line takes the input file -7.wav as input; -o will 
-	give output filename as Track07.flac, and the -f will overwrite if 
-	the file Track04.flac is already present. The encoder will select 
-	encoding preset -3 modified with the -m switch, and use two CPU 
-	threads. Afterwards, the -V will make it decode the flac file and 
-	compare the audio to the input, to ensure they are indeed equal. 
-
+:	This longer example displays how shortform "single dash" options can 
+	be written concatenated until one that takes an argument. In this 
+	example, -j and -o do; thus, after the "j2" a whitespace is needed 
+	to start new options with single/double dash. The -m option takes 
+	no argument (the following "3" is the -3 compression preset). 
+	This set of options could equally well have been written out as 
+	-V -j 2 -m -3 -f -o Track07.flac , or as -foTrack07.flac -3mVj2. 
+	The `-- ` (with whitespace!) signifies end of options, treating 
+	everything to follow as filename; that is needed when an input 
+	filename could otherwise be read as an option, like this "-7".  
+	In total, this line takes the input file -7.wav as input; -o sets 
+	output filename to Track07.flac, and the -f will overwrite if the
+	file Track07.flac already exists. The encoder will select encoding 
+	preset -3 modified with the -m switch, and use two CPU threads. 
+	Afterwards, the -V will make it decode the FLAC file and compare 
+	the audio to the input, to ensure they are indeed equal. 
 
 ## Decoding examples
 
 `flac --decode abc.flac` or `flac -d abc.flac`
-:	Decode abc.flac to abc.wav. abc.flac is not deleted. If abc.wav is
-	already present, the process will exit with an error instead of 
+:	Decode abc.flac to abc.wav. abc.flac is not deleted. If abc.wav 
+	already exists, the process will print an *error* instead of 
 	overwriting; use --force / -f to force overwrite.
 	NOTE: A mere flac abc.flac *without --decode or its shortform -d*, 
-	would mean to re-encode abc.flac to abc.flac (see above), and that 
+	would mean to re-encode abc.flac to abc.flac (see above), and that
 	command would err out because abc.flac already exists.
 
 `flac -d --force-aiff-format abc.flac` or `flac -d -o abc.aiff abc.flac`
 :	Two different ways of decoding abc.flac to abc.aiff (AIFF format).
 	abc.flac is not deleted. -d -o could be shortened to -do.
- 	The decoder can force other output formats, or different versions 
+	The decoder can force other output formats, or different versions 
 	of the WAVE/AIFF formats, see the options below.
 
 `flac -d --keep-foreign-metadata-if-present abc.flac`
 :	If the FLAC file has non-audio chunks stored from the original
 	input file, this option will restore both audio and non-audio. 
 	The chunks will reveal the original file type, and the decoder 
-	will select output format and output file extension accordingly 
-	- note that this is not compatible with forcing a particular 
-	output format except if it coincides with the original, as the
-	decoder cannot transcode non-audio between formats.
-	If there are no such chunks stored, it will decode to abc.wav.
-	The related option \--keep-foreign-metadata will instead exit 
-	with an error if no such non-audio chunks are found.
+	will select output format and output file extension accordingly. 
+	If attempting to force a particular output format, including with
+	e.g. -o abc.wav, it must be the correct one according to the 
+	metadata chunk - the decoder cannot convert chunks between formats.
+	(If there are no such chunks stored, it will decode to abc.wav, 
+	with a *warning*, whereas the related \--keep-foreign-metadata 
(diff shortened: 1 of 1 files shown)


man/flac.md
```diff
@@ -180,160 +180,174 @@ Some typical encoding and decoding tasks using flac:
...
-	a corrupted file.
-	Note: Be careful about trying to "repair" files this way. Often it
-	will only conceal an error, and not play any subjectively "better"
-	than the corrupted file. It is a good idea to at least keep it,
-	and possibly try several decoders, including the one that generated 
-	the file, and hear if one has less detrimental audible errors than 
-	another. Make sure output volume is limited, as corrupted audio can
-	generate loud noises.
+:	Decode abc.flac to abc.wav, not aborting upon *error*. Potentially
+	useful for recovering as much as possible from a corrupted file.
```
ktmf01, line 230: Also a double space here.

man/flac.md
```diff
@@ -448,116 +466,121 @@ Encoding will default to -5, -A "tukey(5e-1)" and one CPU thread.
...
-	Specifying only *FILENAME* is just shorthand for "\|\|\|\|FILENAME". 
-	See the section **Picture specification** for *SPECIFICATION* format.
+:	Import a picture and store it in a PICTURE metadata block, one per 
+	\--picture option given (keeping existing ones upon re-encoding). 
+	A *FILENAME* argument is shorthand for a *SPECIFICATION* with default 
+	values applied (type set to front cover, properties to be inferred 
+	from the picture file); see subsection **Picture specification**.
+	Currently the **flac** encoder handles up to 64 \--picture options; 
+	for more, add afterwards (**metaflac** or tagging software).  
+	NOTE: The FLAC format is limited to 16 MiB *total* metadata. Also,
```
ktmf01, line 516: This isn't true. The limit is 16 MiB per block (so, in this case, per picture)

ktmf01: Also, I don't think it is necessary to mention that using a large number of pictures might cause some programs to reject the files. The same is true of other kinds of metadata as well.

man/flac.md
```diff
@@ -448,116 +466,121 @@ Encoding will default to -5, -A "tukey(5e-1)" and one CPU thread.
...
-:	Add a FLAC tag. The comment must adhere to the Vorbis comment spec;
-	i.e. the FIELD must contain only legal characters, terminated by an
-	'equals' sign. Make sure to quote the content if necessary. This
-	option may appear more than once to add several Vorbis comments. 
-	NOTE 1: All tags will be added to all encoded files.
-	NOTE 2: Upon re-encoding a file with one or more -T or \--tag
-	options, all tags in the input file will be ignored, not only those
-	set with -T / \--tag. 
+:	Set a FLAC tag; like Ogg, the FLAC format uses Vorbis comments, see 
+	the format specification section 8.6. Quote content as necessary.
```
ktmf01, line 540: Maybe use RFC 9639 here for consistency

man/flac.md
```diff
@@ -448,116 +466,121 @@ Encoding will default to -5, -A "tukey(5e-1)" and one CPU thread.
...
+	count, bits per sample, and sample rate, which also must be among 
+	the following: 8, 11.025, 12, 16, 18.9, 22.05, 24, 28, 32, 36, 37.8, 
+	44.1, 48, 56, 64, 72, 75.6, 88.2, 96, 112, 128, 144, 151.2, 176.4, 
+	192, 224, 256, 288, 302.4, 352.8, 384, 448, 512, 576, or 604.8 kHz.  
+	NOTE: This option cannot be used when encoding to stdout nor Ogg FLAC.
+
+**-S** {\#\|\#x\|\#s\|X}, **\--seekpoint**={\#\|\#x\|\#s\|X}
+:	Sets seekpoint(s), overriding the default choice of one per ten seconds
+	('-s 10s'). When several -S options are given, the resulting SEEKTABLE 
+	will contain all the seekpoints (duplicates removed), max 32768.
```
ktmf01, line 567: I don't think that number is correct. Where did you get it from?

ktmf01:
Thanks for all the work you did on this one. And thanks for chopping this up for me. It is still quite a large PR (for documentation anyway) but doable.

I'd like to remove the double spaces (i.e. line breaks) because the manpage format cannot handle those. After discussion this and the other comments, I will merge.

man/flac.md
H2Swine: If it is wrong, it doesn't matter where I got it from. 

I don't quite remember but it could have been from (mis-)reading https://xiph.org/flac/changelog.html . It is mentioned for the 1.3.2 release.

man/flac.md
H2Swine: Ah good, ... you just remove the entire "NOTE:" then?

man/flac.md
```diff
@@ -180,160 +180,174 @@ Some typical encoding and decoding tasks using flac:
...
+:	This longer example displays how shortform "single dash" options can 
+	be written concatenated until one that takes an argument. In this 
+	example, -j and -o do; thus, after the "j2" a whitespace is needed 
+	to start new options with single/double dash. The -m option takes 
+	no argument (the following "3" is the -3 compression preset). 
+	This set of options could equally well have been written out as 
+	-V -j 2 -m -3 -f -o Track07.flac , or as -foTrack07.flac -3mVj2. 
+	The `-- ` (with whitespace!) signifies end of options, treating 
+	everything to follow as filename; that is needed when an input 
+	filename could otherwise be read as an option, like this "-7".
```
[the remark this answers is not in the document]
H2Swine: Some were intended to be for line breaks, some might have been mishaps. Feel free to overrule for this or that reason.

man/flac.md
H2Swine: All fine with me.

PR #850, opened 2025-09-09
Handle "twos" in addition to "sowt" encoding

voxsystems:
"twos" has been showing up in encoded files coming from various Apple products.

### DIFF
--- a/src/flac/encode.c
+++ b/src/flac/encode.c
@@ -655,6 +655,8 @@ static FLAC__bool get_sample_info_aiff(EncoderSession *e, encode_options_t optio
 					return false;
 				if(xx == 0x736F7774) /* "sowt" */
 					e->info.is_big_endian = false;
+				else if (xx == 0x74776F73) /* "twos" */
+					; /* default to big-endian */
 				else if(xx == 0x4E4F4E45) /* "NONE" */
 					; /* nothing to do, we already default to big-endian */
 				else {


ktmf01:
Has this been documented (by Apple) anywhere yet?

H2Swine:
Half a year ago I went searching for the most current AIFF spec from Apple (at least one new enough to permit 4 GiB files), and apparently they have removed everything to try to enforce CAF. But I found data types for the QuickTime format, and who knows whether something has done a "codec copy" to it. 

According to https://developer.apple.com/documentation/quicktime-file-format/sound_sample_descriptions , QuickTime permits
- NONE or unspecified. Unwanted, use "raw" or "twos" instead. 
- raw. Do `ffmpeg -i Unsigned8bit.wav -codec copy outfile.aiff` and you will get an AIFF of type 'raw', but nothing else says it should be a thing.
- twos. For 8-bit signed or 16-bit. 
- sowt. For 16 bit only. 
- in24. Without saying anything about endianness.
- in32. Ditto.

H2Swine:
Apologize that I might have caused another PR ... 

> But I'm confused by this -- what is the difference between "sowt" and "twos" that makes it not flac's job? Isn't this a "standard" from Apple, like "sowt" itself? Or are they just recognizing the "standard" that FFmpeg is (uniquely?) writing?

No ... or "maybe". Here is what I found and not. First about AIFF/AIFF-C: 
- Old AIFF and still-old-but-with-AIFF-C specs: nothing about "twos". "sowt" is "twos" spelled backwards, but the AIFF-C ID for big-endian, was "NONE". 
- I tried to find a newer spec that would allow for 4 GiB files and not just 2 GiB. (Essentially use unsigned for the chunk size.) Found no AIFF spec that allowed it, but IIRC a CAF document said use CAF because AIFF is limited at 4 GiB. 
- Try to find current AIFF spec at Apple's website. I found nothing! 
- I see that Apple is even trying to kill off sowt: https://lists.apple.com/archives/coreaudio-api/2009/Mar/msg00358.html , from the dev team manager ... 

But what I listed above is not for AIFF, it is for the QuickTime format. 
- NONE: deprecated, don't use. 
- twos: use that for big-endian! And for signed 8-bit. But there are other types for 24 and 32. 
- sowt: use that for little-endian 16-bit (only!)
- raw: use that for unsigned 8-bit.

I have found no source that "twos" AIFF should even exist - it seems it should be "NONE". But _FFmpeg_ does create them when doing codec copy from .mov. 
On the other hand, I do have a source indicating that the "raw" might be a thing: https://ss64.com/mac/afconvert.html says that AIFF-C can carry unsigned-integer 8-bit. However, that is apparently not supported by Apple (anymore) according to https://developer.apple.com/library/archive/documentation/MusicAudio/Conceptual/CoreAudioOverview/SupportedAudioFormatsMacOSX/SupportedAudioFormatsMacOSX.html .

H2Swine:
A turn of events, with [help from hydrogenaudio](https://hydrogenaudio.org/index.php/topic,127436.msg1068926.html#msg1068926): 

Apple's "afconvert" CLI does indeed convert these .mov files to "twos" and "raw" AIFF-C.

voxsystems:
> Please, no new PR. This leads to unnecessary duplication.

PR closed, apparently no way to delete.

> The last commit in the branch `twos-aiff-c` of your fork is still the same as here in the PR: `a1daf0f` Branch: https://github.com/

Apologies, pushed to wrong branch.  Fixed now.

ktmf01:
Okay, just took some time to review this. Problem is that this patch just make FLAC accept twos encoding, but doesn't add anything to make decoding possible. This will not work with foreign metadata encoding. Just searching the codebase for "sowt" will show how much more code it should touch. 

I currently don't have much time to make this work, but I won't merge it as it currently is.

PR #859, opened 2025-11-09
Skip ID3v2 tags when decode through errors is enabled 

OwenCochell:
Hello,

The `flac` command issues an error when a flac file starts with an ID3v2 tag. However, it still emits an error even when decode through errors is enabled. This patch changes this behavior with this option. Instead of quitting with an error, the command skips guessing the filetype from the first 12 bytes. We instead use the type guess from the file extension. When decoding occurs, it correctly skips this chunk and is able to process the rest of the flac file as expected. I have added a test to make sure this functionality is correct.

Let me know if there are any problems! All in all I am very unfamiliar with the flac file format (outside of a quick skim read of the RFC), so I am sure there is something I am missing. I'd be happy to correct any issues.

Thanks,
Owen Cochell


ktmf01:
I'm currently quite limited in the time I can spend on FLAC, so this'll have to wait.

ktmf01:
Hi @OwenCochell 

I trimmed down the test file to the bare minimum. Please review

OwenCochell:
Hello @ktmf01 

I have looked over the test changes and they look good. I have also confirmed the tests pass on my machine. I am happy with these changes and have nothing else to add. 

Thanks!

PR #862, opened 2025-12-22
bitwriter_grow_() correctly test for out-of-legit-range values

DEATH923:
Current libFLAC fails to write maximum possible padding value of 0xFFFFFF because of incorrect range check in bitwriter_grow_().
This commit addresses the issue.


ktmf01:
I'll need to split this up

PR #868, opened 2026-01-27
Support ARM64 & ARM64EC on MSVC

DEATH923:
Changed CMake files to detect Microsoft ARM64 & ARM64EC correctly.
Enabled Neon for the above targets.
Modified lpc_intrin_neon.c to compile on MSVC - MSVC does not support int32x4_t val = {a,b,c,d} syntax.


aalekhm:
@ktmf01 Any chance of this getting merged soon.
As I have dependency on flac(arm64EC) for SFML!

ktmf01:
It will probably take a while. I currently don't have a lot of free time to spend on maintaining this project.

PR #870, opened 2026-02-02
Fix metaflac --import-tags-from-file, allow missing newline and \r

### LINKED ISSUE #856: The --import-tags-from option of metaflac does not add the last tag if the file does not end with a newline.
This should be reproducible using version 1.5. To reproduce do:
```
echo -ne "TITLE=title1\nARTIST=artist1" > tag.file
metaflac --import-tags-from=tag.file file.flac`
metaflac --list file.flac
```

You'll notice that the `file.flac` has the `TITLE` tag, but not the `ARTIST`. The tag is silently ignored, which may not be caught until a much later date. I'd say it should be considered on par with the severity of a dataloss bug. So at least issue a warning if this won't be fixed in the near future.

madah81pnz1:
Remove one extra feof() call since it would prevent the last line from being added if it was missing a newline at the end.

Check for both \r and \n, since otherwise the tags will be added with an extra \r included in the tag value, which is unexpected.

Fixes #856

### DIFF
--- a/test/test_metaflac.sh
+++ b/test/test_metaflac.sh
@@ -362,7 +362,7 @@ metaflac_test case39 "--import-tags-from=-" "--list"
 
 # Run same test again, but without newline at the end:
 run_metaflac --remove-all-tags --set-tag="f=0123456789abcdefghij" $flacfile
-echo -n "TITLE=Tittle" | run_metaflac --import-tags-from=- $flacfile
+echo $ECHO_N "TITLE=Tittle" | run_metaflac --import-tags-from=- $flacfile
 check_flac
 metaflac_test case39 "--import-tags-from=-" "--list"
 


madah81pnz1:
> Hmm, the test doesn't seem to work on macos for some reason.

That was surprising, I don't have a MacOS system available to test on locally though. I could modify the test to also print out the metadata it got, a hexdump diff.

It could be due to slight differences in shell syntax between dash and zsh, this I could try a bit on my end first.

Otherwise my guess is it has something to do with how \r was the historical line ending on MacOS. Maybe the shell on MacOS auto-translates \r it into \n, thus the \r\n in the test gets converted into \n\n.

If that's the case, I could either disable the \r\n test on MacOS, or break out the \r\n handling into another PR.

Issue #858, opened 2025-11-06
Files that fail --keep-foreign-metadata


H2Swine:
These are WAVE_FORMAT_PCM type files such that the apparent "WAVE_FORMAT_EXTENSIBLE"-type chunks should be ignored. `flac.exe --keep-foreign-metadata` alters them upon an encode-decode roundtrip. 

```
https://www.hifistatement.info/downloads/14-09-01_peterson/Mellow_Tone_16-44.wav
https://www.hifistatement.info/downloads/14-09-01_peterson/Mellow_Tone_24-192.wav
```

Two discussions on whether they are compliant (and one more file): https://github.com/dbry/WavPack/issues/218 and https://hydrogenaudio.org/index.php/topic,128635 

Addendum: some of the test files from that source behave this way, others don't. The [person(s) who run the site](https://www.hifistatement.net/impressum.html) might possibly know what software was used.

ktmf01:
Does the decoder at least return an error that it didn't restore the metadata correctly?

H2Swine:
Yes, sorry for not including that: 
`ERROR verifying foreign metadata restore from Mellow_Tone_16-44.flac to Mellow_Tone_16-44.flac.wav: stored main chunk length differs from written length`

But it gives no warnung upon _encoding_ that you won't get your file back this way.

tobinus:
@H2Swine mentioned this issue in #878 :

> Might I ask whether this has anything to do with the issue reported in #858 , or maybe it is so "close to" what you have fixed, that you might do this too?

I performed a test, and the pull request does not fix that issue. However, you get a new error message, since the size is the same now:

> ERROR verifying foreign metadata restore from Mellow_Tone_24-192.flac to rt2.wav: stored foreign format block differs from written block. Perhaps the file is being restored to a different format than that of the original file

This is because FLAC writes zeroes to the restored `cbSize` field (`0x00 00`), while the original file has some nonsensical bytes there (`0x40 FD`).

I _guess_ encoders are free to store whatever they want in those two bytes, since [the documentation](https://learn.microsoft.com/en-us/previous-versions/windows/desktop/ee419019(v=vs.85)) states:

> For WAVE_FORMAT_PCM formats (and only WAVE_FORMAT_PCM formats), this member is ignored.

[...]

PR #878, opened 2026-02-18
Preserve redundant cbsize = 0 with --keep-foreign-metadata

### LINKED ISSUE #808: Decoding with --keep-foreign-metadata fails fmt chunk size check when extra data size is present and 0
We are encountering an issue with specific WAV files when using `--keep-foreign-metadata`, where the length of the `fmt ` chunk differs before and after a roundtrip through FLAC.

## The problem

The issue seems to be the second to last field of the `fmt ` chunk, which indicates the number of extra format bytes that follows. If it is present, and set to `0x0000`, it will contribute to the original size field of the `fmt ` chunk (`0x12`). However, FLAC does _not_ write the field upon decoding when it is set to 0, opting to leave it out. The size of the decoded `fmt ` chunk is two bytes smaller to match (`0x10`). However, the verification at the end fails:

```
$ flac --keep-foreign-metadata -do /tmp/test.wav /tmp/test.flac
NOTE: --keep-foreign-metadata is a new feature; make sure to test the output file before deleting the original.

flac 1.5.0
Copyright (C) 2000-2009  Josh Coalson, 2011-2025  Xiph.Org Foundation

test.flac: done
ERROR verifying foreign metadata restore from /tmp/test.flac to /tmp/test.wav: stored main chunk length differs from written length
```

Roundtrip seems to work fine for files that include extra format bytes, and files with no extra format bytes when they also leave out the extra format bytes size, so this problem is specific to WAV files that:

1. Have no extra format bytes at the end of the `fmt ` chunk
2. Still include the size field for the extra format bytes (set to `0x00`)

## Expected behaviour

I would expect one of two things:

1. The decoded WAV file includes the second to last field even if it is set to zero, as long as the stored information about the original `fmt ` chunk indicates that its size field was set to `0x12` and therefore included the field. This should make the `fmt ` chunk pass verification.
2. The decoded WAV file leaves out the second to last field (like today), and the verification of the `fmt ` chunk is modified to accept a difference of -2 bytes when the second-to-last field of the original `fmt ` is 0.

## How to reproduce

I will upload an audio file and add it here after the issue is created.

Anyways, I'd like to take the opportunity at the end here to thank you so much for your work on FLAC 😄 It has been, and continues to be, of great help 😊

tobinus:
A year back, I reported a [limitation with the `--keep-foreign-metadata` functionality][1], where decoding fails if the `fmt ` chunk is defined according to the [WAVEFORMATEX structure][2] (with `wFormatTag = WAVE_FORMAT_PCM` while including `cbSize`) despite not having any additional format information (so that `cbSize` is 0). I ran into this issue again this week, so I decided to have a go at solving it.

Thinking aloud, I see three ways of dealing with this limitation:

1. Accept the limitation, but maybe warn about it when encoding
2. Modify decoding so that the decoded `fmt ` chunk equals the original in this case
3. Modify the verification of the decoding, so that it accepts the `fmt ` chunk losing `cbSize`, or provide some way for the user to accept such changes from the original file. Whether `cbSize` is there or not in `WAVE_FORMAT_PCM` files makes no difference for applications consuming WAV anyway

Judging by earlier discussions on the mailing list, I'm guessing that solution 3 is undesirable. Of the two remaining solutions, I prefer solution **2** since it would let us convert the recordings we have at work to and from FLAC with `--keep-foreign-metadata`.

The decoding logic is already complicated, with logic regarding `fmt ` chunk length spread across multiple locations, so I'm a bit worried about adding more to the burden. That is why I have refactored this part in this pull request. Please note that I am not fluent in C, so I may have made some trade-offs concerning performance or memory usage without knowing.

Alternatively, I have also pushed a [branch without refactoring while keeping support for WAVE64][3], and a [branch where cbSize = 0 is not supported for WAVE64][4], simplifying the solution a bit.

What do you think?

[1]: https://github.com/xiph/flac/issues/808
[2]: https://learn.microsoft.com/en-us/previous-versions/dd757713(v=vs.85)
[3]: https://github.com/xiph/flac/compare/master...tobinus:flac:808-handle-redundant-cbsize/all-formats
[4]: https://github.com/xiph/flac/compare/master...tobinus:flac:808-handle-redundant-cbsize/except-wave64

Closes #808

### DIFF
--- a/src/flac/decode.c
+++ b/src/flac/decode.c
@@ -890,10 +890,13 @@ FLAC__bool write_iff_headers(FILE *f, DecoderSession *decoder_session, FLAC__uin
 		if(!write_riff_wave_fmt_chunk_body(f, is_waveformatextensible, preserve_dummy_cbsize, decoder_session->bps, decoder_session->channels, decoder_session->sample_rate, decoder_session->channel_mask))
 			return false;
 
-		if(format == FORMAT_WAVE64 && !is_waveformatextensible && preserve_dummy_cbsize) {
-			/* fmt block size 18 must be padded with 6 bytes in order to align with 8 byte alignment */
-			if(flac__utils_fwrite("\x00\x00\x00\x00\x00\x00", 1, 6, f) != 6)
-				return false;
+		if(format == FORMAT_WAVE64) {
+			unsigned padding = ((8 - ftello(f) % 8) % 8);
+			if(padding > 0) {
+				/* fmt chunk must be padded in order to align with 8-byte alignment */
+				if(flac__utils_fwrite("\x00\x00\x00\x00\x00\x00\x00", 1, padding, f) != padding)
+					return false;
+			}
 		}
 
 		decoder_session->fm_offset2 = ftello(f);


src/flac/decode.c
```diff
@@ -867,21 +875,30 @@ FLAC__bool write_iff_headers(FILE *f, DecoderSession *decoder_session, FLAC__uin
...
+			if(!write_little_endian_uint64(f, calculate_wave64_fmt_chunk_size_field(is_waveformatextensible, preserve_dummy_cbsize)))
 				return false;
 		}
 
-		if(!write_riff_wave_fmt_chunk_body(f, is_waveformatextensible, decoder_session->bps, decoder_session->channels, decoder_session->sample_rate, decoder_session->channel_mask))
+		if(!write_riff_wave_fmt_chunk_body(f, is_waveformatextensible, preserve_dummy_cbsize, decoder_session->bps, decoder_session->channels, decoder_session->sample_rate, decoder_session->channel_mask))
 			return false;
 
+		if(format == FORMAT_WAVE64) {
+			unsigned padding = ((8 - ftello(f) % 8) % 8);
```
tobinus: Is there a better way to get the number of bytes we need to pad with to align at 8 bytes? I had to use a modulo of the final result in order to get `0` instead of `8` when no padding is needed (checking for `padding < 8` below seemed a bit counter-intuitive).

tobinus:
@H2Swine 

> Might I ask whether this has anything to do with the issue reported in #858 , or maybe it is so "close to" what you have fixed, that you might do this too?

The way I see it, this PR fixes a subset of the cases in #858, namely the files that store zero in the `cbSize` field. Files that store anything else will still fail. So this PR can be viewed as a stepping stone towards fixing it, perhaps? [I've followed up with a comment](https://github.com/xiph/flac/issues/858#issuecomment-3943429849) on that issue.

PR #884, opened 2026-02-24
Add support for zeroing least-significant bits in encoder settings

lolrdor1gamer:
Adds CLI/API option to zero N LSBs per sample before encoding.


ktmf01:
https://xiph.org/flac/developers.html

> FLAC should be lossless. This seems obvious but lossy compression seems to creep into every audio codec. This goal also means that flac should stay archival quality and be truly lossless for all input. Testing of releases should be thorough.

PR #891, opened 2026-03-05
Add --escapes, \n-style escapes to allow multiline comments

### LINKED ISSUE #232: metaflac  - error writing multi-line tag from stdin
Seeing as there's no option within metaflac to `--delete-all-tags-except` I devised a workaround that writes tags to be preserved to stdout and pipes same to another call to metaflac which deletes all tags and then writes back input from stdin.

The general form would be:
`metaflac --show-tag tagname1 --show-tag tagname2   --show-tag tagnamex "filename.flac" | metaflac --preserve-modtime --remove-all-tags --import-tags-from=- "filename.flac" `

However, I've come up against an issue, as the following call will demonstrate:
metaflac --show-tag lyrics x.flac | metaflac --import-tags-from=- x.flac

Running the following should result in the corresponding vorbiscomment simply being rewritten, and it works as intended until it in encounters a tag with a CR or LF within the tag, meaning it will not work for the likes of a Lyrics tag.

To illustrate:
`metaflac --list x.flac`
```
METADATA block #0
  type: 0 (STREAMINFO)
  is last: false
  length: 34
  minimum blocksize: 4608 samples
  maximum blocksize: 4608 samples
  minimum framesize: 1009 bytes
  maximum framesize: 14722 bytes
  sample_rate: 44100 Hz
  channels: 2
  bits-per-sample: 16
  total samples: 8627724
  MD5 signature: 74bdec14815e8c901cf1f9a87171e2c5
METADATA block #1
  type: 4 (VORBIS_COMMENT)
  is last: false
  length: 73
  vendor string: reference libFLAC 1.3.1 20141125
  comments: 1
    comment[0]: lyrics=FIRST LINE
SECOND LINE
METADATA block #2
  type: 1 (PADDING)
  is last: true
  length: 22445
```
As soon as there's a CR or LF in a comment, such as in the example above metaflac fails with the following:
```
-: ERROR: malformed vorbis comment field "SECOND LINE",
       field contains no '=' character
```

madah81pnz1:
Same as how it works in vorbiscomment, except \0 is not supported.

\0 is problematic to support in C code, and it is ambiguous what something like \012 should mean, given that octal escapes are not supported by vorbiscomment anyway.

Fixes #232

### DIFF
--- a/src/metaflac/utils.c
+++ b/src/metaflac/utils.c
@@ -32,6 +32,7 @@
 #include "share/safe_str.h"
 #include "share/utf8.h"
 #include "share/compat.h"
+#include "share/grabbag.h"
 
 void die(const char *message)
 {
@@ -233,11 +234,29 @@ FLAC__bool parse_vorbis_comment_field(const char *field_ref, char **field, char 
 	return true;
 }
 
-void write_vc_field(const char *filename, const FLAC__StreamMetadata_VorbisComment_Entry *entry, FLAC__bool raw, FILE *f)
+void write_vc_field(const char *filename, const FLAC__StreamMetadata_VorbisComment_Entry *entry, FLAC__bool raw, const FLAC__bool escapes, FILE *f)
 {
-	if(0 != entry->entry) {
-		if(filename)
+	if(NULL != entry->entry) {
+		const char *entry_str = (const char *)entry->entry;
+		size_t entry_length = entry->length;
+		char *allocated_str = NULL;
+
+		if(filename) {
 			flac_fprintf(f, "%s:", filename);
+		}
+
+		if(escapes) {
+			if(grabbag__escape_string_needed(entry_str, entry_length)) {
+				allocated_str = grabbag__create_escaped_string(entry_str, entry_length);
+				if(allocated_str != NULL) {
+					entry_str = allocated_str;
+					entry_length = strlen(allocated_str);
+				}
+				else {
+					/* error is ignored, the original value will be used */
+				}
+			}
+		}
 
 		if(!raw) {
 			/*
@@ -245,21 +264,26 @@ void write_vc_field(const char *filename, const FLAC__StreamMetadata_VorbisComme
 			 * be truncated by utf_decode().
 			 */
 #ifdef _WIN32
-			flac_fprintf(f, "%s", entry->entry);
+			flac_fprintf(f, "%s", entry_str);
 #else
 			char *converted;
 
-			if(utf8_decode((const char *)entry->entry, &converted) >= 0) {
+			if(utf8_decode(entry_str, &converted) >= 0) {
 				(void) local_fwrite(converted, 1, strlen(converted), f);
 				free(converted);
 			}
 			else {
-				(void) local_fwrite(entry->entry, 1, entry->length, f);
+				(void)local_fwrite(entry_str, 1, entry_length, f);
 			}
 #endif
 		}
 		else {
-			(void) local_fwrite(entry->entry, 1, entry->length, f);
+			(void)local_fwrite(entry_str, 1, entry_length, f);
+		}
+
+		if(allocated_str) {
+			free(allocated_str);
+			allocated_str = NULL;
 		}
 	}
 #ifdef _WIN32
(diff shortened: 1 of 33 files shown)


madah81pnz1:
I tried to be compatible with vorbiscomment, but decided to drop support for \0, since that doesn't really work in vorbiscomment anyway. I don't think it's worth supporting embedded \0, as many tools either have problems or just silently truncates the string.

The case for supporting \0 is that it makes it possible for the user to repair a faulty tag; by exporting the escaped tags, it should be easy to remove "\0" in any text editor, and recover the end part of the tag after it. Resaving such tag in other audio players or tools might truncate the tag at the first \0, which is what vorbiscomment itself claims to do.

\0 is also just a special case of an octal escape sequence. But those are a bit ambiguous, e.g. what should \012 mean; is it \0 + "12", or \n. This means for octal escape encoding, either always use 3 characters, or check if the next unescaped character is a number 0..7, and if so use 3 characters. So \0 + "0" would be encoded as \0000, and \0 + "A" as \0A. Interestingly, Rust does not support octal escape sequences, but does support \0 as a special case.

\x have similar problems, due to how it is defined in C and C++ as "arbitrary number of hexadecimal digits". Consider if you have a long string that only consists of 0-9 and a-f characters, then it's impossible to have a \x escape sequence in the middle of it, as there is no way to tell where it ends, e.g. "\x123456789". Python changed its \x in PEP-223 to be exactly two hex digits, Rust does the same.

I think it's more useful to support \u and \U instead, which would let you export and import without the risk of UTF-8 turning into mojibake. But even the need for this should be rare, and --no-utf8-convert can already be used as a workaround if needed. In most languages, \u is exactly 4 hex digits, and \U is 8. The exception is Bash that allows variable hex digits like for \x. Bash otherwise follows the C standard escaping, so this is some discrepancy.

Another thing to consider is that the exported tags might be processed further by other user tools and scripts, so it makes sense to keep the escaping simple. Replacing newlines with \n should be trivial in any language, but \u requires more parsing logic and a full UTF-8 encoder/decoder.

References:
https://en.cppreference.com/w/c/language/escape.html
https://en.cppreference.com/w/cpp/language/escape.html
https://peps.python.org/pep-0223/
https://docs.python.org/3/reference/lexical_analysis.html#escape-sequences
https://doc.rust-lang.org/reference/tokens.html#ascii-escapes

src/share/grabbag/escapes.c
```diff
@@ -0,0 +1,190 @@
...
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+#include <stdio.h>
+#include <stddef.h>
+#include "share/grabbag.h"
+
+FLAC__bool grabbag__escape_string_needed(const char *src, size_t src_size)
```
madah81pnz1, line 23: Since this is just a performance optimization, and running escape/unescape over all tags shouldn't be a bottleneck, this function could be removed.

src/share/grabbag/escapes.c
```diff
@@ -0,0 +1,190 @@
...
+{
+	size_t d = 0;
+	size_t s = 0;
+	*error = false;
+
+	while((d < dst_size) && (s < src_size)) {
+		switch(src[s]) {
+			case '\\':
+				dst[d++] = '\\';
+				dst[d++] = '\\';
```
madah81pnz1: This might overrun d by one byte, but shouldn't be a problem in this specific usage, as the calling functions has +1 byte available for the null terminator.
Should be fixed,  add a if (d + 1 < dst_size) to also prevent writing a partially escaped character.

src/share/grabbag/escapes.c
```diff
@@ -0,0 +1,190 @@
...
+			dst[size] = '\0';
+		}
+	}
+	return dst;
+}
+
+FLAC__bool grabbag__unescape_string_needed(const char *src, size_t src_size)
+{
+	for(size_t s = 0; s < src_size; ++s) {
+		if(src[s] == '\\') {
```
madah81pnz1, line 114: Could use strchr() instead here, but at the time I wasn't sure if handling non-null terminated strings would be important.

src/share/grabbag/escapes.c
```diff
@@ -0,0 +1,190 @@
...
+			return true;
+		}
+	}
+	return false;
+}
+
+static size_t grabbag__unescape_string_size(const char *src, size_t src_size)
```
madah81pnz1, line 121: The size could also be done by the unescape_string() if provided a NULL dst buffer, which is preferable in case the escaping becomes more complicated in the future.

src/share/grabbag/escapes.c
```diff
@@ -0,0 +1,190 @@
...
+						dst[d++] = '\\';
+						break;
+					case 'r':
+						dst[d++] = '\r';
+						break;
+					case 'n':
+						dst[d++] = '\n';
+						break;
+					default:
+						/* Unsupported escape character */
```
madah81pnz1, line 158: Could add a "ignore errors" mode, which keeps all invalid/unrecognized escape sequences as-is. But for now I don't think it is useful.

src/metaflac/utils.c
```diff
@@ -233,33 +234,56 @@ FLAC__bool parse_vorbis_comment_field(const char *field_ref, char **field, char
...
+
+		if(escapes) {
+			if(grabbag__escape_string_needed(entry_str, entry_length)) {
+				allocated_str = grabbag__create_escaped_string(entry_str, entry_length);
+				if(allocated_str != NULL) {
+					entry_str = allocated_str;
+					entry_length = strlen(allocated_str);
+				}
+				else {
+					/* error is ignored, the original value will be used */
```
madah81pnz1, line 256: There is no way to return an error in this function. But also grabbag__create_escaped_string() can't really fail unless malloc fails.

Ignoring errors is also done for the utf8_decode() call below, which would print the original string in case it fails. Maybe could be improved in the future, since --show-tag might show the tag in the wrong encoding to the user, without giving any indiciation or error, which would be surprising.

Issue #25, opened 2017-01-02
flac-1.3.2: case07-expect.meta test failure


tgurr:
The release tarball available on http://downloads.xiph.org/releases/flac/ doesn't carry the changes from ac39d3719f16dfb6e08d2fbde8ccaf34a266c81d for the case07-expect.meta file, resulting in a test failure:

[...]
test case06: --set-tag=TITLE... OK
test case07: --show-vendor-tag --show-tag=ARTIST... make[1]: Leaving directory '/var/tmp/paludis/build/media-libs-flac-1.3.2/work/flac-1.3.2/test'
ERROR: metadata does not match expected ../test/metaflac-test-files/case07-expect.meta
make[1]: *** [Makefile:707: check] Error 1
make: *** [Makefile:486: check-recursive] Error 1
[...]

erikd:
The mirroring for download.xiph.org (hosted by http://osuosl.org/ ) is borked.

For now https://sourceforge.net/projects/flac/files/flac-src/ may be a better option. The sha256sum should be:
```
91cfc3ed61dc40f47f050a109b08610667d73477af6ef36dcad31c31a4a8d53f  flac-1.3.2.tar.xz
```

tgurr:
Thanks, that one works fine.

erikd:
Cool. I'll close this then. Someone is working on getting the mirrors fixed.

Issue #24, opened 2016-12-31
"Double free or corruption" involving client data in decoder


mrkkrp:
OK, I've spent several hours yesterday and today trying to figure this out, and so I'd like to try to seek help here. I'm writing a Haskell binding to libFLAC. It's quite complete right now and everything works, except for one issue with decoder.

The decoder looks like this:

```haskell
-- | Decode a FLAC file to WAVE.
--
-- 'DecoderException' is thrown when underlying FLAC decoder reports a
-- problem.

decodeFlac :: MonadIO m
  => DecoderSettings   -- ^ Decoder settings
  -> FilePath          -- ^ File to decode
  -> FilePath          -- ^ Where to save the resulting WAVE file
  -> m ()
decodeFlac DecoderSettings {..} ipath' opath' = liftIO . withDecoder $ \d -> do
  ipath <- makeAbsolute ipath'
  opath <- makeAbsolute opath'
  liftInit (decoderSetMd5Checking d decoderMd5Checking)
  (maxFrameSize, wave) <- runFlacMeta def ipath $ do
    let waveFileFormat   = decoderWaveFormat
        waveDataOffset   = 0
        waveDataSize     = 0
        waveOtherChunks  = []
    waveSampleRate <- retrieve SampleRate
    waveSampleFormat <- SampleFormatPcmInt . fromIntegral
      <$> retrieve BitsPerSample
    waveChannelMask <- retrieve ChannelMask
    waveSamplesTotal <- retrieve TotalSamples
    maxFrameSize <- fromIntegral <$> retrieve MaxFrameSize
    return (maxFrameSize, Wave {..})
  let bufferSize = maxFrameSize * fromIntegral (waveBlockAlign wave) + 1
  withTempFile' opath $ \otemp ->
    withBuffer bufferSize $ \buffer -> do
      initStatus <- decoderInitHelper d ipath buffer
      case initStatus of
        DecoderInitStatusOK -> return ()
        status -> throwIO (DecoderInitFailed status)
      liftBool d (decoderProcessUntilEndOfMetadata d)
      processedRef <- newIORef (0 :: Word64)
      writeWaveFile otemp wave $ \h -> fix $ \nextOne -> do
        processed <- readIORef processedRef
        unless (processed == waveSamplesTotal wave) $ do
          liftBool d (decoderProcessSingle d)
          frameSize <- fromIntegral <$> decoderGetFrameSize d
          let toGrab = frameSize * fromIntegral (waveBlockAlign wave)
[...]

Issue #23, opened 2016-12-30
compiling FLAC using Visual Studio 2010 compatibility


bardware:
Hi,

I'm in the process of compiling `sox `audio tool including FLAC using Visual Studio. For compatibility reasons everything has to be compiled using toolset v100 (Visual Studio 2010 compatibility).
FLAC's `compat.h` file has the following lines

```
#if defined _MSC_VER
#  if _MSC_VER >= 1800
#    include <inttypes.h>
```

`inttypes.h` is marked missing by the IDE as the compiler version is right, but the tools - including the used header files - do not match the compiler.

Maybe you can check the test if it's safe to include the header.

erikd:
You don't say what version of FLAC you're using; whether its the 1.3.1 release of current git HEAD. Also, I don't have access to the Microsoft compilers, but @lvqcl should be able to help.

bardware:
I recently downloaded the source from github, the files in the `flac-master.zip` date december 24th

lvqcl:
Is it [IntelliSense](https://en.wikipedia.org/wiki/Intelligent_code_completion#IntelliSense) error or a *real* compile error? If it's the former then it doesn't matter.

See also:
- https://connect.microsoft.com/VisualStudio/feedback/details/761812/msc-ver-does-not-reflect-selected-toolset-in-editor
- http://stackoverflow.com/questions/20728952/does-msc-ver-depend-on-toolkit-version
- http://stackoverflow.com/questions/3040498/determine-when-using-the-vc90-compiler-in-vs2010-instead-of-vs2008

bardware:
As everything is working as needed, I trust it's an IntelliSense "error". Thanks!

Issue #22, opened 2016-12-30
FLAC__stream_decoder_process_single decodes final frame twice?


mrkkrp:
Here is the minimal example that demonstrates the problem:

```c
#include <FLAC/stream_decoder.h>
#include <FLAC/metadata.h>
#include <stdlib.h>
#include <stdio.h>

static FLAC__StreamDecoderWriteStatus write_callback
  ( const FLAC__StreamDecoder *decoder
  , const FLAC__Frame *frame
  , const FLAC__int32 *const ibuffer[]
  , void *client_data)
{
  return FLAC__STREAM_DECODER_WRITE_STATUS_CONTINUE;
}

static void error_callback
  ( const FLAC__StreamDecoder *decoder
  , FLAC__StreamDecoderErrorStatus status
  , void *client_data )
{
  return;
}

int main (int argc, char *argv[])
{
  unsigned block_size, state;
  FLAC__StreamDecoder *decoder = FLAC__stream_decoder_new();

  FLAC__stream_decoder_init_file(decoder, argv[1], write_callback, NULL, error_callback, NULL);
  FLAC__stream_decoder_process_until_end_of_metadata(decoder);

  /* Total samples: 14442750 */

  while ((state = FLAC__stream_decoder_get_state(decoder)) != FLAC__STREAM_DECODER_END_OF_STREAM)
    {
      printf("State so far: %d\n", state);
      FLAC__stream_decoder_process_single(decoder);
      block_size = FLAC__stream_decoder_get_blocksize(decoder);
      printf("Processed frame of size: %d\n", block_size);
    }
  FLAC__stream_decoder_delete(decoder);
}
```

Compile it:

```
$ gcc -lFLAC -o my_prog main.c
```

Next I feed it a FLAC file that I know contains exactly 14442750 samples. With frame size of 4096, the behavior I expect is that the program prints

```
State so far: 2
Processed frame of size: 4096
```

…quite a few times, but eventually it should say that it processed the final frame of length `14442750 % 4096 == 254` samples. After that the status sholud change to `FLAC__STREAM_DECODER_END_OF_STREAM` and the application should finish.

The actual result is:

```
…
Processed frame of size: 4096
State so far: 2
Processed frame of size: 254
State so far: 2
Processed frame of size: 254
```

As you can see it processes 254 final samples twice, which leads to exactly 254 “extra” samples in output (the code that puts the result into a buffer is not shown here for simplicity).

[...]

PR #896, opened 2026-03-22
Fix problem with SetConsoleOutputCP CP_UTF8 on Windows

madah81pnz1:


madah81pnz1:
Note this needs more testing, especially piping binary from stdin.
Also this removes the \r\n and \n translation, which might mean some adaptions are needed in case a user is piping a file that does contains \r\n, since then the \r will appear in the input.

Also might need another test binary to be able to test windows console output properly, e.g. by calling ReadConsoleW() to check what was actually written.

madah81pnz1:
> Note this needs more testing, especially piping binary from stdin.

Reading from stdin turns into mojibake, e.g. --import-tags-from-file=-
I guess it has to use the same method as output, ReadConsoleW/ReadFile, Using fopen with stdin will not work, as the (wide) input will then be translated to the current codepage.

madah81pnz1:
This seems to work now, I will clean up the code and push more commits, but it might take some more time.

The behavior is now changed for both stdout and stdin:
* Console (FILE_TYPE_CHAR) uses ReadConsoleW()/WriteConsoleW() with UTF-16, doesn't use the codepage.
* Redirect to/from pipe (FILE_TYPE_PIPE) uses the codepage from GetConsoleCP()/GetConsoleOutputCP().
* Redirect to/from disk (FILE_TYPE_DISK) uses UTF-8 only, ignores the codepage.
* Command-line arguments uses UTF-16, so also no need for codepage conversion.

All calls to utf8_encode() and utf8_decode() have been removed when #ifdef _WIN32 due to this, since the conversion needs to happen as early as possible, not when handling the tags.

--no-utf8-convert only affects pipe redirects from stdin, not stdout. This is because there is no way to pass down a "raw" flag to vfprintf_utf8 and wprint_console. It could be added with some additional refactoring, but I'm not sure if it is worth it.

fgets() has also been replaced with new functions that handles buffering and any newline combination of \r \n \r\n. The reasoning behind this is that it lets flac import files that were created on other platforms, as otherwise the \r would be part of the tag on non-Windows platforms, and only Windows would handle \r and \r\n as the same.
Newlines are now always a single \n even on Windows. This works fine even in Windows XP console, even when piped via more. The downside is if the user opens a redirected file with an editor that can't handle \n, which is the case with Notepad before Windows 10, see https://devblogs.microsoft.com/commandline/extended-eol-in-notepad/

The pipe redirect is a bit troublesome though. Seems like an unfixable problem, unless you use chcp 65001 (CP_UTF8) by default for the console.
Since printing out any unicode to the console always works due to WriteConsoleW(), it might come as a surprise to the user that characters turn into ? or � when piped via more or find.
But there's no way to make those tools work otherwise, they all have to agree which codepage to use.

But considering also that if you run flac/metaflac with Python, e.g. subprocess.run() and capture_output=True. This is now a redirect and thus everything will be converted to the console output codepage. It is not so straight-forward how you could set the console codepage to 65001/CP_UTF8 from Python, but this is true for any console-mode application on Windows.

One way to handle this could be to print a warning on stderr if the current text can't be losslessly decoded/encoded via the console codepage. This could also be done for corrupted UTF-8 when converted into UTF-16 for the console.

Testing all this is not so straightfoward though. I've used a Python script for the pipe and disk parts, and a C++ program that calls ReadConsoleOutputW() directly for the console output.
However ReadConsoleOutputW() does not handle surrogate pairs at all, and only returns replacement characters (U+FFFD), even if I can see the correct characters in the console, and manual selection copying from the console works.
Testing with a very long tag is hard due to the 32K commandline limit, and also --import-tags-from-file has a 64K line limit. I used the libFLAC API directly to set longer tags, but then there's UTF8_BUFFER_SIZE being defined as 32768, which silently truncates all longer strings in vfprintf_utf8().
This makes it hard to test the new buffer mechanism that converts in several steps between UTF-16, UTF-8 and the console codepage. But also, maybe no-one really cares about such long tags? Even a full album lyrics tag wouldn't come close to that limit.

ktmf01:
Thanks for all the hard work you've put into this! Pity I haven't gotten much time to help out recently. I'm confident you'll/we'll be able to work this out eventually.

madah81pnz1:
Seems I forgot to add an implementation for grabbag__file_input_reader_refill() for non-WIN32. I was too focused on testing this on Windows. Should be fixed now in next commit, all cmake and make tests pass on debian.

madah81pnz1:
Seems fuzzer failed due to timeout? Or it got stuck doing realloc() in FLAC__metadata_object_vorbiscomment_resize_comments(), which does look a bit suspicious.
I will take a closer look at this more later. Is it possible to download the fuzzing input to reproduce this locally?
````
2026-06-30T10:25:09.6055189Z ALARM: working on the last Unit for 27 seconds
2026-06-30T10:25:09.6055559Z        and the timeout value is 25 (use -timeout=N to change)
2026-06-30T10:25:09.6056197Z MS: 5 ChangeByte-InsertRepeatedBytes-ChangeBinInt-CrossOver-CopyPart-; base unit: 79862ff05f1924b25ee8d067c1c12339f14bd976
2026-06-30T10:25:09.6057042Z artifact_prefix='/tmp/tmpfee4241r/'; Test unit written to /tmp/tmpfee4241r/timeout-d582005d608a27ffd2eb4716f5def7e9e523bdb3
2026-06-30T10:25:09.6057609Z ==27741== ERROR: libFuzzer: timeout after 27 seconds
2026-06-30T10:25:09.6058136Z     #0 0x55e7e8e62fa1 in __sanitizer_print_stack_trace /src/llvm-project/compiler-rt/lib/asan/asan_stack.cpp:87:3
2026-06-30T10:25:09.6058915Z     #1 0x55e7e8d54d48 in fuzzer::PrintStackTrace() /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerUtil.cpp:210:5
2026-06-30T10:25:09.6059612Z     #2 0x55e7e8d3781d in fuzzer::Fuzzer::AlarmCallback() /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:304:5
2026-06-30T10:25:09.6060329Z     #3 0x7f9ff2ed541f  (/lib/x86_64-linux-gnu/libpthread.so.0+0x1441f) (BuildId: 9753720502573b97dbac595b61fd72c2df18e078)
2026-06-30T10:25:09.6061020Z     #4 0x7f9ff2d41d34  (/lib/x86_64-linux-gnu/libc.so.6+0xbbd34) (BuildId: 5792732f783158c66fb4f3756458ca24e46e827d)
2026-06-30T10:25:09.6061908Z     #5 0x55e7e8db78fa in __asan::Allocator::Reallocate(void*, unsigned long, __sanitizer::BufferedStackTrace*) /src/llvm-project/compiler-rt/lib/asan/asan_allocator.cpp:782:7
2026-06-30T10:25:09.6062951Z     #6 0x55e7e8db7804 in __asan::asan_realloc(void*, unsigned long, __sanitizer::BufferedStackTrace*) /src/llvm-project/compiler-rt/lib/asan/asan_allocator.cpp:1044:34
2026-06-30T10:25:09.6063759Z     #7 0x55e7e8e57d1f in realloc /src/llvm-project/compiler-rt/lib/asan/asan_malloc_linux.cpp:82:10
2026-06-30T10:25:09.6064554Z     #8 0x55e7e8f09bb0 in FLAC__metadata_object_vorbiscomment_resize_comments /src/flac/src/libFLAC/metadata_object.c:1231:18
2026-06-30T10:25:09.6065310Z     #9 0x55e7e8f0abbc in FLAC__metadata_object_vorbiscomment_insert_comment /src/flac/src/libFLAC/metadata_object.c:1280:7
2026-06-30T10:25:09.6066176Z     #10 0x55e7e8ea6bf2 in set_vc_field /src/flac/oss-fuzz/../src/metaflac/operations_shorthand_vorbiscomment.c:332:7
2026-06-30T10:25:09.6066866Z     #11 0x55e7e8ea6106 in import_vc_from /src/flac/oss-fuzz/../src/metaflac/operations_shorthand_vorbiscomment.c:383:10
2026-06-30T10:25:09.6067639Z     #12 0x55e7e8ea6106 in do_shorthand_operation__vorbis_comment /src/flac/oss-fuzz/../src/metaflac/operations_shorthand_vorbiscomment.c:111:9
2026-06-30T10:25:09.6068391Z     #13 0x55e7e8e9f2d2 in do_shorthand_operations_on_file /src/flac/oss-fuzz/../src/metaflac/operations.c:434:10
2026-06-30T10:25:09.6069152Z     #14 0x55e7e8e9f2d2 in do_shorthand_operations /src/flac/oss-fuzz/../src/metaflac/operations.c:398:9
2026-06-30T10:25:09.6069742Z     #15 0x55e7e8e9f2d2 in do_operations /src/flac/oss-fuzz/../src/metaflac/operations.c:96:8
2026-06-30T10:25:09.6070267Z     #16 0x55e7e8eb4497 in main_to_fuzz /src/flac/oss-fuzz/./../src/metaflac/main.c:68:10
2026-06-30T10:25:09.6070792Z     #17 0x55e7e8eb4497 in LLVMFuzzerTestOneInput /src/flac/oss-fuzz/tool_metaflac.c:119:2
2026-06-30T10:25:09.6071569Z     #18 0x55e7e8d38f7d in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:619:13
2026-06-30T10:25:09.6072591Z     #19 0x55e7e8d385b5 in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool, bool*) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:516:7
2026-06-30T10:25:09.6073508Z     #20 0x55e7e8d3a285 in fuzzer::Fuzzer::MutateAndTestOne() /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:765:19
2026-06-30T10:25:09.6074578Z     #21 0x55e7e8d3aee5 in fuzzer::Fuzzer::Loop(std::__Fuzzer::vector<fuzzer::SizedFile, std::__Fuzzer::allocator<fuzzer::SizedFile>>&) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:910:5
2026-06-30T10:25:09.6075657Z     #22 0x55e7e8d29a85 in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:917:6
2026-06-30T10:25:09.6076428Z     #23 0x55e7e8d556f2 in main /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerMain.cpp:20:10
2026-06-30T10:25:09.6077104Z     #24 0x7f9ff2caa082 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x24082) (BuildId: 5792732f783158c66fb4f3756458ca24e46e827d)
2026-06-30T10:25:09.6077725Z     #25 0x55e7e8d1cddd in _start (build-out/fuzzer_tool_metaflac+0x8eddd)
````

PR #905, opened 2026-05-31
Simplify "extern inlines" in bitreader and bitwriter

### LINKED ISSUE #895: Windows/clang-cl: error linking static FLAC.lib
Several functions in the FLAC library are declared like this:

https://github.com/xiph/flac/blob/master/src/libFLAC/include/private/bitwriter.h#L59

but then defined like this (note `inline`):

https://github.com/xiph/flac/blob/master/src/libFLAC/bitwriter.c#L357

It also has a comment at the bottom of the file:

https://github.com/xiph/flac/blob/master/src/libFLAC/bitwriter.c#L944

They are refering to the C standard that the way these functions are declared and defined is "valid".

When the library gets linked, the linker is unable to find these symbols. They get inlined and the out-of-line function bodies are not emitted.

The full bitwriter.c build rule on my setup (from ninja.build):

```
build src\libFLAC\CMakeFiles\FLAC.dir\bitwriter.c.obj: C_COMPILER__FLAC_unscanned_Release D$:\modding\nlc\engine\reaper-engine\3rd_party\Src\Xiph\flac\flac\src\libFLAC\bitwriter.c || cmake_object_order_depends_target_FLAC
  CONFIG = Release
  DEFINES = -DFLAC__NO_DLL -DHAVE_CONFIG_H -D_CRT_SECURE_NO_WARNINGS -D_DARWIN_C_SOURCE -D_POSIX_PTHREAD_SEMANTICS -D_TANDEM_SOURCE -D_USE_MATH_DEFINES -D__STDC_WANT_IEC_60559_BFP_EXT__ -D__STDC_WANT_IEC_60559_DFP_EXT__ -D__STDC_WANT_IEC_60559_FUNCS_EXT__ -D__STDC_WANT_IEC_60559_TYPES_EXT__ -D__STDC_WANT_LIB_EXT2__ -D__STDC_WANT_MATH_SPEC_FUNCS__
  FLAGS = -Wall -Wextra -Wstrict-prototypes -Wmissing-prototypes -Waggregate-return -Wcast-align -Wnested-externs -Wshadow -Wundef -Wmissing-declarations -Winline /DWIN32 /D_WINDOWS /O2 /Ob2 /Oi /Ot /Oy /DNDEBUG -DNDEBUG /O2 /Ob2 /DNDEBUG /MP /O2 /Ob2 /Oi /Oy /Ot /GL /Gy /Gw /W4 /GF /GS /Zi /GT /Brepro /bigobj /permissive- /volatile:iso /Zc:inline /Zc:preprocessor /Zc:enumTypes /Zc:lambda /Zc:__STDC__ /Zc:__cplusplus /Zc:externConstexpr /Zc:throwingNew /Zc:checkGwOdr /Zc:templateScope /DNOMINMAX -fno-delayed-template-parsing -fstrict-aliasing /D_CRT_STDIO_ISO_WIDE_SPECIFIERS -fno-ms-compatibility -fgnuc-version=0 /utf-8 -DS_IWRITE=_S_IWRITE -Dfileno=_fileno -Dgetenv=getenv -Doff_t=_off_t -Dstrdup=_strdup -Dutimbuf=_utimbuf -march=skylake -mavx2 -mvpclmulqdq /FIintrin.h -flto -fmerge-all-constants -fforce-emit-vtables -fwhole-program-vtables /clang:-fcoro-aligned-allocation /clang:-fopenmp -Wextra -Wmost -Wno-error=unused-command-line-argument -Werror=format -Wformat-nonliteral -Werror=format-pedantic -Werror=format-signedness -Werror=format-type-confusion -Werror=microsoft -Werror=move -Werror=nan-infinity-disabled -Werror=parentheses -Werror=strict-aliasing -Werror=tautological-compare -Werror=typename-missing -Werror=weak-vtables /D_WIN32_WINNT=0x0A00 /DWINVER=0x0A00 /D_MSVC_STL_HARDENING=1 /DNDEBUG /D_ITERATOR_DEBUG_LEVEL=0 -clang:-std=c23 -MD /fp:fast /wd4267 /wd4996 -Wdeclaration-after-statement
  INCLUDES = -ID:\modding\nlc\engine\reaper-engine\3rd_party\Src\Xiph\flac\flac\include -ID:\modding\nlc\engine\reaper-engine\3rd_party\Src\Xiph\flac\_Build -ID:\modding\nlc\engine\reaper-engine\3rd_party\Src\Xiph\flac\flac\src\libFLAC\include -imsvcD:\modding\nlc\engine\reaper-engine\3rd_party\Src\Xiph\ogg\_Build\Release\x64\include
```

I don't see anything illegal here. All other libraries and my whole project compile and work.

I tried adding `-fgnu89-inline` and defining `inline` as `inline __attirubte__((__gnu_inline__))`, but the symbols still don't get emitted.

To me it looks like incorrectly written library code. On the other hand, I don't see any similar issue reports here.
So first of all, I'd like to understand whether it's a compiler problem or a library problem. I've filed the same issue to the LLVM repo to be completely fair: https://github.com/llvm/llvm-project/issues/187759

solbjorn:
Extern inlines are not clearly defined by the C standard and are often interpreted differently depending in the compiler. These ones already involve workarounds for MSVC.
Clang under certain optimization options (LTO etc.) also doesn't intepret this correctly and doesn't emit standalone symbols for those at all, leading to linking errors.

Simplify these by defining plain static inlines to be used within the corresponding files and separate global symbols to be used externally. This doesn't make the codegen worse for bitreader and bitwriter, but guarantees that the global symbols will always be defined regardless of the compiler and compilation options.

Closes #895.

### DIFF
--- a/src/libFLAC/bitreader.c
+++ b/src/libFLAC/bitreader.c
@@ -378,17 +378,17 @@ FLAC__uint16 FLAC__bitreader_get_read_crc16(FLAC__BitReader *br)
 	return br->read_crc16;
 }
 
-static inline FLAC__bool __FLAC__bitreader_is_consumed_byte_aligned(const FLAC__BitReader *br)
+FLAC__bool FLAC__bitreader_is_consumed_byte_aligned(const FLAC__BitReader *br)
 {
 	return ((br->consumed_bits & 7) == 0);
 }
 
-static inline uint32_t __FLAC__bitreader_bits_left_for_byte_alignment(const FLAC__BitReader *br)
+uint32_t FLAC__bitreader_bits_left_for_byte_alignment(const FLAC__BitReader *br)
 {
 	return 8 - (br->consumed_bits & 7);
 }
 
-static inline uint32_t __FLAC__bitreader_get_input_bits_unconsumed(const FLAC__BitReader *br)
+uint32_t FLAC__bitreader_get_input_bits_unconsumed(const FLAC__BitReader *br)
 {
 	return (br->words-br->consumed_words)*FLAC__BITS_PER_WORD + br->bytes*8 - br->consumed_bits;
 }
@@ -552,7 +552,7 @@ FLAC__bool FLAC__bitreader_read_raw_int64(FLAC__BitReader *br, FLAC__int64 *val,
 	return true;
 }
 
-static inline FLAC__bool __FLAC__bitreader_read_uint32_little_endian(FLAC__BitReader *br, FLAC__uint32 *val)
+FLAC__bool FLAC__bitreader_read_uint32_little_endian(FLAC__BitReader *br, FLAC__uint32 *val)
 {
 	FLAC__uint32 x8, x32 = 0;
 
@@ -618,7 +618,7 @@ FLAC__bool FLAC__bitreader_skip_byte_block_aligned_no_crc(FLAC__BitReader *br, u
 
 	FLAC__ASSERT(0 != br);
 	FLAC__ASSERT(0 != br->buffer);
-	FLAC__ASSERT(__FLAC__bitreader_is_consumed_byte_aligned(br));
+	FLAC__ASSERT(FLAC__bitreader_is_consumed_byte_aligned(br));
 
 	if(br->read_limit_set && br->read_limit < (uint32_t)-1){
 		if(br->read_limit < nvals*8){
@@ -663,7 +663,7 @@ FLAC__bool FLAC__bitreader_read_byte_block_aligned_no_crc(FLAC__BitReader *br, F
 
 	FLAC__ASSERT(0 != br);
 	FLAC__ASSERT(0 != br->buffer);
-	FLAC__ASSERT(__FLAC__bitreader_is_consumed_byte_aligned(br));
+	FLAC__ASSERT(FLAC__bitreader_is_consumed_byte_aligned(br));
 
 	if(br->read_limit_set && br->read_limit < (uint32_t)-1){
 		if(br->read_limit < nvals*8){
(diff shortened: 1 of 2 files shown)


ktmf01:
I wonder whether the inline keyword should be dropped entirely. I just did for the bitreader, as those functions were used only for debugging. I'll need to check whether it makes sense for the bitwriter too.

ktmf01:
I just did some performance testing, and it didn't seem to have any effect really. I'd rather fix it this way. Any thoughts?

ktmf01:
I see I missed this warning

```
bitwriter.c:335:49: warning: 'bitwriter_grow_' is static but used in inline function 'FLAC__bitwriter_write_raw_uint32_nocheck' which is not static
  335 |         if(bw->capacity <= bw->words + bits && !bitwriter_grow_(bw, bits))
      |
```

Need to check that out as well

solbjorn:
> I just did some performance testing, and it didn't seem to have any effect really. I'd rather fix it this way. Any thoughts?

Sure, go ahead!

My primary goal was just to remove "extern inlines" as their definition in the C spec is very ambigous and can lead to such problems. So I only manually made sure that inlines are still inlines, but their global copies are always emitted.

I realize that FLAC has a CI matrix and never had such issues as I had, but out of 70 external libs that I have (mixed C and C++), only FLAC didn't want to link. The compilation flags are a bit custom, but consistent across the whole project.

I also wanted to compare object code size before/after, but as my home project is for Windows, I'm unable to do that some easy way (only 3rd-party apps non-trivial to build). It's much easier on Linux for me (`scripts/bloat-o-meter.pl` from the kernel repo, which just uses `nm`).

PR #908, opened 2026-06-11
ci: add read-only token permissions and pin all action references to SHAs

Alb3e3:
## Summary

All five CI workflows lacked a top-level `permissions:` block, leaving
each run to inherit the repository default (potentially write-all).
Each workflow now has `permissions: read-all` at the top level.

Every action reference using a mutable version tag was pinned to its
full commit SHA (17 pins total), including `actions/checkout`,
`actions/upload-artifact`, `KyleMayes/install-llvm-action`,
`msys2/setup-msys2`, and the OSS-Fuzz cifuzz actions.

## Verification

```
uvx zizmor --min-severity high .github/workflows/
```
Result: **no findings** after this patch (was 17 high-severity before).

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -4,6 +4,8 @@ on:
   push:
   pull_request:
 
+permissions: read-all
+
 jobs:
   build:
     strategy:
@@ -67,9 +69,9 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         if: startsWith(matrix.build-system,'cmake')
         with:
           repository: xiph/ogg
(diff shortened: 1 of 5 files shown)


ktmf01:
> Every action reference using a mutable version tag was pinned to its full commit SHA (17 pins total), including `actions/checkout`, `actions/upload-artifact`, `KyleMayes/install-llvm-action`, `msys2/setup-msys2`, and the OSS-Fuzz cifuzz actions.

Why?

Alb3e3:
The reason is supply-chain control: a version tag is mutable, so a workflow can start running different action code without any change in this repository. Pinning to a commit SHA makes action updates explicit and reviewable, and the explicit read-only permissions limit the repository token available if an action is ever compromised.

That said, I agree there is a maintenance tradeoff, especially for provider-owned actions like `actions/checkout` / `actions/upload-artifact` and for the Google OSS-Fuzz actions. The higher-value part here is the explicit read-only token permissions, and if you want SHA pins at all, the narrowest target is the non-provider third-party action (`msys2/setup-msys2`).

If that better matches the project's policy, I can re-scope this PR to keep the permission blocks and only pin `msys2/setup-msys2`, leaving GitHub-owned and Google-maintained actions on their normal version refs.

ktmf01:
I'd like to not pin anything (msys2/setup-msys2 too) but perhaps it is time to update stuff. I'll take the liberty of doing that directly in this PR, and squash it to one co-authored commit when everything works.

PR #911, opened 2026-06-17
Add CI support for Windows ARM64

tsomanna:
**PR Description**

This pull request adds a native Windows ARM64 job to the existing CI workflow.

The change introduces a configuration that runs on the `windows-11-arm` runner and builds
flac using ARM64 as the target platform. The changes are added to:

- action.yml
- msys2.yml

No existing CI jobs or platforms are modified.
This is a minimal, additive change intended to improve CI coverage for Windows on ARM and enable native Windows ARM64 builds.

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -64,6 +64,26 @@ jobs:
             build-system: cmake
             configure-opts: '-DCMAKE_C_STANDARD=11 -DBUILD_SHARED_LIBS=ON'
 
+          - name: windows-11-arm-cmake
+            os: windows-11-arm
+            build-system: cmake
+            configure-opts: ''
+
+          - name: windows-11-arm-cmake-shared
+            os: windows-11-arm
+            build-system: cmake
+            configure-opts: '-DBUILD_SHARED_LIBS=ON'
+
+          - name: windows-11-arm-cmake-c11
+            os: windows-11-arm
+            build-system: cmake
+            configure-opts: '-DCMAKE_C_STANDARD=11'
+
+          - name: windows-11-arm-cmake-c11-shared
+            os: windows-11-arm
+            build-system: cmake
+            configure-opts: '-DCMAKE_C_STANDARD=11 -DBUILD_SHARED_LIBS=ON'
+
     runs-on: ${{ matrix.os }}
 
     steps:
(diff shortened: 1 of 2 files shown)


tsomanna:
Hi, any updates on this?

khmyznikov:
@ktmf01 @Alb3e3 sorry for nudging, do you guys still maintain this repo? Maybe we can do something to make it easier?

Alb3e3:
No worries on the ping, but just to set expectations: I'm not a maintainer here. My only involvement was a small CI-hardening patch (#908), and I have no merge rights, so please don't wait on me for this one. @ktmf01 is the maintainer who reviews and merges, so this PR needs their eyes.

Alb3e3:
Not a maintainer, but since I touched these workflows recently (#908) I took a proper look. Reviewing as a contributor in case it helps move this along.

The change itself looks sound:

- `action.yml`: the four `windows-11-arm-*` cmake entries mirror the existing `windows-latest-*` cmake variants one-for-one (plain / shared / c11 / c11-shared), so the matrix stays consistent and the footprint is proportionate rather than excessive.
- `msys2.yml`: converting the single job into an x64 + arm64 matrix is the right shape, and `clangarm64` is the correct MSYS2 environment for native ARM64.

A few things worth resolving before it can go green:

1. The actual blocker is that CI has not run on the current head. This is a PR from a fork, so the ARM64 jobs need a maintainer to approve the workflow run before anyone can see whether they build. Until that happens there is no signal on whether flac actually compiles and tests clean on `windows-11-arm`, which is the whole point of the PR. That approval is the single thing that would move this forward.

2. `choco install pandoc` is the most likely place the arm64 `msys2` job falls over. The Chocolatey pandoc package is an x64 build, so on `windows-11-arm` it runs through x64 emulation, which is at best slow and at worst not present on the runner. Since pandoc is only needed for the man pages (which are architecture independent and already built on the x64 job), the cleanest fix is probably to skip the pandoc step on the arm64 leg, or gate the docs build on `matrix.arch == 'x64'`.

3. The branch is three commits behind master and was cut before the recent workflow updates landed (`permissions: read-all`, and `actions/checkout` / `actions/upload-artifact` bumped to v7). A normal merge preserves those since this PR does not touch those lines, but a rebase on master is still worth doing so the new ARM64 jobs run against the current workflow rather than the older one.

4. Minor: `msys2.yml` loses its trailing newline in this diff.

Happy to help test the pandoc-on-arm64 behaviour if that is useful.

tsomanna:
Hi @Alb3e3, thank you for the suggestions. That was very helpful!
@ktmf01, do let us know if the changes look fine to you and how we can proceed from here.

PR #913, opened 2026-06-19
Add missing overflow check to safe_calloc_

hyder365:
Unlike the other safe_* allocation wrappers in this file,
safe_calloc_ did not check whether nmemb * size would overflow
before calling calloc().  All existing call sites pass tightly
bounded arguments, so this is defense-in-depth against future
misuse.


ktmf01:
According to the calloc manpage, calloc has built-in overflow checks, so I don't see how this is necessary?

hyder365:
You're right. I didn't realize that calloc already handles it internally. Sorry for the noise.

PR #920, opened 2026-07-18
Support decoding directly to .aifc

madah81pnz1:
Fixes so that flac doesn't default to WAVE for .aifc extension

### DIFF
--- a/src/flac/main.c
+++ b/src/flac/main.c
@@ -1500,6 +1500,8 @@ int encode_file(const char *infilename, FLAC__bool is_first_file, FLAC__bool is_
 			input_format = FORMAT_AIFF;
 		else if(infilename_length >= 5 && 0 == FLAC__STRCASECMP(infilename+(infilename_length-5), ".aiff"))
 			input_format = FORMAT_AIFF;
+		else if(infilename_length >= 5 && 0 == FLAC__STRCASECMP(infilename + (infilename_length - 5), ".aifc"))
+			input_format = FORMAT_AIFF_C;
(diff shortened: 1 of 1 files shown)


src/flac/main.c
```diff
@@ -1500,6 +1500,8 @@ int encode_file(const char *infilename, FLAC__bool is_first_file, FLAC__bool is_
 			input_format = FORMAT_AIFF;
 		else if(infilename_length >= 5 && 0 == FLAC__STRCASECMP(infilename+(infilename_length-5), ".aiff"))
 			input_format = FORMAT_AIFF;
+		else if(infilename_length >= 5 && 0 == FLAC__STRCASECMP(infilename + (infilename_length - 5), ".aifc"))
```
madah81pnz1, line 1503: I wanted to add a test for this in test_flac.sh, and it turns out we already have this for aiff:

````
convert_to_aiff ()
{
	run_flac "$2" $1.raw || die "ERROR converting $1.raw to AIFF"
	run_flac "$3" $1.flac -o $1.aiff || die "ERROR converting $1.raw to AIFF"
}
````

However, adding .aifc here in the same way would still pass even without this commit, since there is nothing that checks that the actual output file is AIFF or AIFF-C; the test would still pass but we get a WAV file with a .aifc file extension.

Maybe there is a portable way to use `file` to check what format the output file actually is, I can try to add it in next commit. File would return one of `IFF data, AIFF audio`, `IFF data, AIFF-C compressed audio`, `RIFF (little-endian) data, WAVE audio, Microsoft PCM, 16 bit, stereo 44100 Hz`, so grepping for "AIFF audio", "AIFF-C" or "WAVE audio" should do the trick.

